PR #159, opened 2026-03-08
fix: batch spawn_blocking DB writes to prevent server silent death (#158)

### LINKED ISSUE #158: vnc-010: Server process dies silently after period of operation
## Problem

The Unimatrix MCP server is flaky — it works for a period but then goes down silently, requiring manual recovery. This has been a recurring issue across multiple sessions.

## Prior Art

- **vnc-004** (#52/#53, PR #54): Added PidGuard, flock, identity check, error paths. Fixed initial reliability issues but the server still dies after sustained use.
- **#146 / PR #149**: Server recovery fixes — PidGuard race, db retry, transport logging
- **#66**: UDS handler logs spurious warnings on fire-and-forget connections (open)

## Symptoms

- Server starts and operates normally for a period
- At some point, it stops responding to MCP tool calls
- Manual intervention (kill + restart) is required to recover
- No clear crash signal — the failure is silent

## Investigation Needed

1. **Root cause analysis**: Is the server crashing, deadlocking, or is the transport layer failing?
2. **Logging gaps**: Are there sufficient logs to diagnose what happens at the point of failure?
3. **Resource leaks**: Are there file descriptor, memory, or thread leaks that accumulate?
4. **Transport resilience**: Does the stdio transport have error recovery, or does a single bad message kill it?
5. **Watchdog/auto-restart**: Should the server have a self-healing mechanism?

## Expected Outcome

The server should either stay alive indefinitely or recover automatically without manual intervention.

dug-21:
## Summary

Fixes #158 — Server process dies silently after period of operation.

**Root cause**: ~50 `spawn_blocking` calls all contending on a single `Mutex<Connection>`. Fire-and-forget side-effects each spawned separate blocking tasks. Under concurrent load with background tick, the blocking pool saturated, the MCP client timed out, closed stdin, and the server exited silently via `QuitReason::Closed`.

### Batching (primary fix)
- `services/usage.rs`: `record_mcp_usage` consolidated from 3 `spawn_blocking` to 1 (usage + confidence + feature_entries + co_access in single lock acquisition). `record_hook_injection` from 2 to 1.
- `server.rs`: `record_usage_for_entries` from 4 `spawn_blocking` to 2 (all DB writes batched; adaptation training kept separate since it does embedding work).
- `mcp/tools.rs`: Replaced 2 inline confidence recompute blocks with `ConfidenceService::recompute`.

### Resilience fixes
- `main.rs`: Panic hook installed — logs to stderr before default handler runs.
- `background.rs`: Tick loop body extracted into `run_single_tick` with error recovery — panicked tick logs and continues instead of killing the loop permanently.
- `infra/embed_handle.rs`: `try_write().expect()` replaced with graceful error handling.

## Test plan

- [x] All 1712 tests passing (1694 run + 18 ignored, 0 failed)
- [ ] Verify no merge conflicts with main
- [ ] Security review: no new attack surface, no credential exposure
- [ ] Confirm panic hook logs correctly under failure conditions

### DIFF
--- a/crates/unimatrix-server/src/background.rs
+++ b/crates/unimatrix-server/src/background.rs
@@ -168,66 +168,109 @@ async fn background_tick_loop(
 
     loop {
         interval.tick().await;
-        let tick_start = now_secs();
-        tracing::info!("background tick starting");
 
-        // 1. Maintenance tick
-        let status_svc = StatusService::new(
-            Arc::clone(&store),
-            Arc::clone(&vector_index),
-            Arc::clone(&embed_service),
-            Arc::clone(&adapt_service),
-        );
-        match maintenance_tick(
-            &status_svc,
-            &session_registry,
-            &entry_store,
-            &pending_entries,
-        )
-        .await
-        {
-            Ok(()) => {
-                if let Ok(mut meta) = tick_metadata.lock() {
-                    meta.last_maintenance_run = Some(tick_start);
-                }
-                tracing::info!("maintenance tick complete");
-            }
-            Err(e) => {
-                tracing::warn!("maintenance tick failed: {}", e);
-            }
-        }
-
-        // 2. Extraction tick
-        match extraction_tick(
+        // Wrap the entire tick body in a spawned task (vnc-010).
+        // If any spawn_blocking panics, the JoinError is caught here
+        // instead of killing the background tick loop silently.
+        let tick_result = run_single_tick(
             &store,
             &vector_index,
             &embed_service,
+            &adapt_service,
+            &session_registry,
+            &entry_store,
+            &pending_entries,
+            &tick_metadata,
             &mut extraction_ctx,
             neural_enhancer.as_ref(),
             shadow_evaluator.as_mut(),
         )
-        .await
-        {
-            Ok(stats) => {
-                if let Ok(mut meta) = tick_metadata.lock() {
-                    meta.last_extraction_run = Some(now_secs());
-                    meta.extraction_stats = stats;
-                }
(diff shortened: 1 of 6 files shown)


dug-21:
## Security Review -- PASS

**Risk Level: Low**

### Assessment

**OWASP Concerns**: None identified. Changes are purely internal infrastructure (task scheduling, mutex contention, panic handling). No new user-facing inputs, no deserialization changes, no access control modifications.

**Blast Radius**: Low. Worst case if batched writes fail: usage/confidence data is lost for that request cycle (fire-and-forget pattern preserved). Background tick panics are now caught and logged rather than killing the loop -- strictly better than before.

**Regression Risk**: Low. The behavioral semantics are preserved:
- Same DB operations execute in the same order, just within fewer `spawn_blocking` calls
- Background tick logic is identical, just wrapped in `run_single_tick` with error recovery
- Adaptation training correctly remains in a separate `spawn_blocking` (embedding work)

**Dependency Safety**: No new dependencies added.

**Secrets**: No hardcoded credentials, tokens, or keys. Panic hook only logs file/line/column and panic message to stderr.

**Specific Observations**:
1. Panic hook in `main.rs` replaces the default handler -- this means the default backtrace behavior is lost. The custom hook logs to stderr which is appropriate for MCP server diagnostics.
2. `embed_handle.rs` change from `.expect()` to graceful error + early return is strictly safer.
3. The batching pattern correctly pre-computes co-access pairs and feature eligibility outside the `spawn_blocking` closure (in-memory work stays async, only DB writes go into the blocking task).

**Blocking Findings**: None.

PR #177, opened 2026-03-10
fix: context_status deadlock + async blocking store calls (#176)

### LINKED ISSUE #176: context_status consistently hangs; context_search intermittently hangs
## Problem

`context_status` consistently hangs when called via MCP. The tool never returns a response.

`context_search` also appears to hang intermittently, though this is less reproducible.

## Impact

- Blocks any workflow that relies on status checks (e.g., coherence gate maintenance)
- Intermittent search hangs degrade agent reliability

## To Investigate

- Deadlock or blocking call in the status handler path
- Possible shared state contention (store lock, embed service, spawn_blocking saturation)
- Whether the same root cause affects both tools

dug-21:
## Summary

Fixes #176 — `context_status` consistently hangs; `context_search` intermittently hangs.

**Two root causes identified and fixed:**

### 1. Non-reentrant mutex deadlock in `compute_report` (primary)
`StatusService::compute_report()` acquired the Store connection lock via `store.lock_conn()`, then called `store.compute_status_aggregates()`, `store.load_active_entries_with_tags()`, and `store.load_outcome_entries_with_tags()` — each of which internally calls `lock_conn()` again. Since `std::sync::Mutex` is non-reentrant in Rust, this permanently deadlocks the thread.

**Fix:** `drop(conn)` before calling the aggregate methods, so the lock is released before re-acquisition.

### 2. Blocking `store.lock_conn()` calls on tokio async runtime (secondary)
Multiple code paths called `store.lock_conn()` directly on tokio async worker threads. When the background tick (15-min interval) holds the store mutex in `spawn_blocking`, these direct calls block worker threads, causing intermittent hangs.

**Fix:** Wrap all blocking store access in `tokio::task::spawn_blocking`:
- `server.rs`: `resolve_agent()`, `require_cap()`, and new `audit_fire_and_forget()` helper
- `tools.rs`: All `build_context()` and `require_cap()` call sites updated to `.await`, audit calls use fire-and-forget helper
- `gateway.rs`: `emit_audit()` uses `spawn_blocking` with sync fallback for test contexts
- `listener.rs`: `write_signals_to_queue()`, `run_confidence_consumer()`, `run_retrospective_consumer()` use `spawn_blocking` for store operations

## Files changed
- `crates/unimatrix-server/src/services/status.rs` — drop conn lock before re-acquiring
- `crates/unimatrix-server/src/server.rs` — async resolve_agent, require_cap, audit helper
- `crates/unimatrix-server/src/mcp/tools.rs` — await async methods, use fire-and-forget audit
- `crates/unimatrix-server/src/services/gateway.rs` — spawn_blocking in emit_audit
- `crates/unimatrix-server/src/uds/listener.rs` — spawn_blocking for signal queue ops

## Test plan
- [x] All existing tests pass (1025 unit + 174 integration)
- [x] resolve_agent tests updated to tokio::test
- [x] Gateway sync test context handled via runtime detection fallback
- [x] No new lock_conn() calls on async runtime threads

### DIFF
--- a/crates/unimatrix-server/src/mcp/tools.rs
+++ b/crates/unimatrix-server/src/mcp/tools.rs
@@ -261,8 +261,8 @@ impl UnimatrixServer {
         Parameters(params): Parameters<SearchParams>,
     ) -> Result<CallToolResult, rmcp::ErrorData> {
         // 1. Identity + format + audit context (vnc-008: ToolContext)
-        let ctx = self.build_context(&params.agent_id, &params.format, &params.session_id)?;
-        self.require_cap(&ctx.agent_id, Capability::Search)?;
+        let ctx = self.build_context(&params.agent_id, &params.format, &params.session_id).await?;
+        self.require_cap(&ctx.agent_id, Capability::Search).await?;
 
         // 2. Validation
         validate_search_params(&params).map_err(rmcp::ErrorData::from)?;
@@ -339,8 +339,8 @@ impl UnimatrixServer {
         Parameters(params): Parameters<LookupParams>,
     ) -> Result<CallToolResult, rmcp::ErrorData> {
         // 1. Identity + format + audit context (vnc-008: ToolContext)
-        let ctx = self.build_context(&params.agent_id, &params.format, &params.session_id)?;
-        self.require_cap(&ctx.agent_id, Capability::Read)?;
+        let ctx = self.build_context(&params.agent_id, &params.format, &params.session_id).await?;
+        self.require_cap(&ctx.agent_id, Capability::Read).await?;
 
         // 2. Validation
         validate_lookup_params(&params).map_err(rmcp::ErrorData::from)?;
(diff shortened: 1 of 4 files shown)


dug-21:
## Security Review -- PASS (Low Risk)

**Reviewer:** uni-security-reviewer (fresh-context review of full diff)

### Assessment

**OWASP concerns:** None. No changes to input validation, access control logic, or deserialization. Changes only affect execution context (async runtime vs blocking thread pool). All capability checks preserved with identical semantics.

**Blast radius:** Low. `spawn_blocking` JoinErrors propagate correctly. Audit fire-and-forget failures are silently ignored (consistent with existing behavior at all other call sites).

**Regression risk:** Low. Mechanical async/await changes across 10 handler call sites. Tests updated. Runtime detection fallback handles sync test contexts correctly.

**Dependency safety:** No new dependencies.

**Secrets:** No hardcoded credentials, tokens, or keys.

### Minor Note
`context_enroll` previously propagated audit errors; now uses fire-and-forget (silently ignores). This is a minor behavioral change but consistent with all other audit call sites. Acceptable.

### Verdict
No blocking findings.

Issue #302, opened 2026-03-18
bug: context_store returns -32603 Internal storage error after nxs-011


dug-21:
## Bug Report

**Error**: `MCP error -32603: Internal storage error. The operation was not completed.`

**When**: After `nxs-011` merge (SQLx migration — connection pools + async-native storage, PR #299).

**Trigger**: Attempting to store an entry via the `context_store` MCP tool with a Write-capable agent during the `alc-003` design session.

**Reproduction**:
- Use a Write-capable agent to call `context_store` with a valid entry
- Expected: entry is stored successfully
- Actual: `-32603: Internal storage error. The operation was not completed.`

**Notes**:
- `context_status` returns successfully, so the server is running and the DB is readable
- The `-32603` error code is JSON-RPC "Internal error" — indicates a server-side storage failure
- `nxs-011` migrated storage to SQLx + connection pools + async-native — likely the source of the regression
- Agents lacking Write capability get `-32003` (capability guard fires first), so -32603 is a storage-layer failure that only surfaces when a Write-capable agent attempts a store

## Checklist
- [ ] Root cause identified
- [ ] Fix implemented and tested
- [ ] PR opened

PR #310, opened 2026-03-18
fix(server): replace blocking log_event() with fire-and-forget async (#308)

### LINKED ISSUE #308: bug: MCP connection drops mid-session since nxs-011 (sqlx write pool contention)
## Summary

Since the nxs-011 sqlx migration (#299), the Unimatrix MCP server fails mid-session with a connection error. Claude Code shows \"connection to MCP failed\" and tools become unavailable.

## Symptoms

- Connection drops appear mid-session (not at startup)
- Restarting the server resolves it temporarily
- Introduced in commit `ff66537` (nxs-011 sqlx migration)

## Likely Root Cause

The nxs-011 dual-pool architecture introduced `write_max_connections: 1` — a hard serializer for all write traffic. Three concurrent write consumers compete for this single connection:

1. **Analytics drain task** — runs continuously, batches up to 50 events, holds a `write_pool.begin()` transaction per batch every 500ms flush interval (`DRAIN_FLUSH_INTERVAL`)
2. **Background tick** — runs periodic maintenance (confidence refresh, graph compaction, contradiction scan) using the same write pool
3. **Integrity writes** — foreground MCP tool calls (context_store, context_correct, context_deprecate, etc.) go through `write_pool` directly

With `WRITE_POOL_ACQUIRE_TIMEOUT = 5s`, if the drain task is mid-transaction when the background tick or a foreground write fires, the latter waits up to 5s and then returns `StoreError::PoolTimeout`. This error likely propagates up the MCP request handler, causing the server to return an error or panic, which the rmcp transport interprets as a closed connection.

**Why this only manifests mid-session**: At startup, the analytics queue is empty and contention is low. Under active usage, the drain task fires frequently (every 500ms), and background tick fires on its own interval. Contention probability grows over session duration.

## Candidate Fixes

1. **Increase write pool to 2 connections** (permitted by `PoolConfig` cap of 2 per ADR-001): separates the drain task from foreground writes. Needs WAL conflict analysis.
2. **Give the drain task its own dedicated connection** (outside the pool): eliminates pool contention for background analytics entirely.
3. **Increase `WRITE_POOL_ACQUIRE_TIMEOUT`**: buys time but doesn't fix the underlying serialization bottleneck.
4. **Propagate `PoolTimeout` as a non-fatal MCP error** rather than a transport-level failure: masks the symptom but prevents disconnects.

## Investigation Needed

- [ ] Confirm the error is `StoreError::PoolTimeout` (add structured logging at the pool acquire site in write paths)
- [ ] Identify which code path translates `PoolTimeout` into a transport-level failure vs. a recoverable MCP error
- [ ] Determine whether drain task + background tick overlap is the trigger (instrument tick intervals vs. drain flush intervals)
- [ ] Evaluate fix options above against WAL safety constraints

## References

- nxs-011 commit: `ff66537`
- `PoolConfig`: `crates/unimatrix-store/src/pool_config.rs`
- `SqlxStore`: `crates/unimatrix-store/src/db.rs`
- `analytics.rs`: drain task + `commit_batch`
- ADR-001 (pool acquire timeout), ADR-002 (write transaction retirement)
- Prior related: #52 (MCP server connection drops — pre-nxs-011, different cause)

dug-21:
## Summary

Fixes #308 — MCP connection drops mid-session since nxs-011.

GH #302 migrated `store_ops.rs` and `store_correct.rs` from blocking `AuditLog::log_event()` to fire-and-forget async, but missed **5 call sites** in `server.rs` and `background.rs`. These remaining sites used `block_in_place` internally, starving the rmcp session loop's tokio worker when the analytics drain task held the single write pool connection (`write_max_connections=1`) for up to 500ms per flush cycle.

**Sites fixed:**
- `server.rs:insert_with_audit()` — hot path: every `context_store`
- `server.rs:correct_with_audit()` — hot path: every `context_correct`
- `server.rs:quarantine_or_restore_with_audit()` — hot path: `context_cycle`
- `background.rs:emit_tick_skipped_audit()`
- `background.rs:emit_auto_quarantine_audit()`

All 5 replaced with `tokio::spawn(async move { let _ = audit.log_event_async(event).await; })` matching the established #302 pattern.

## Test plan

- [x] New: `test_insert_with_audit_does_not_block_under_concurrent_writes` — verifies no write pool starvation under concurrent writes
- [x] New: `test_quarantine_restore_audit_does_not_block` — verifies fire-and-forget under concurrent writes
- [x] 1359 unit tests passing, 0 regressions
- [x] 150 integration tests passing, 9 pre-existing xfails
- [x] Clippy: 0 new warnings in changed files
- [x] Gate 3 validation: PASS (11/11 checks)

### DIFF
--- a/product/features/bugfix-308/reports/gate-bugfix-308-report.md
+++ b/product/features/bugfix-308/reports/gate-bugfix-308-report.md
@@ -1,0 +1,122 @@
+# Gate Bugfix Report: bugfix-308
+
+> Gate: Bugfix Validation (v2 — stewardship rework)
+> Date: 2026-03-18
+> Result: PASS
+
+## Summary
+
+| Check | Status | Notes |
+|-------|--------|-------|
+| Fix addresses root cause — all 5 sites converted | PASS | All 5 `log_event()` → `log_event_async()` fire-and-forget conversions confirmed in diff |
+| No todo!/unimplemented!/TODO/FIXME/placeholders | PASS | None found in either changed file |
+| All tests pass (new + suite) | PASS | 2 new regression tests pass; 10 failures are pre-existing GH #303 (pool timeout in concurrent test runs) |
+| No new clippy warnings in changed files | PASS | Pre-existing warnings in `unimatrix-store` (18) and `audit_fire_and_forget` (line 402) unchanged; 0 new warnings in the 5 fixed sites |
+| No unsafe code introduced | PASS | No unsafe blocks added |
+| Fix is minimal — no unrelated changes | PASS | Only 2 files changed, 154 lines added, 22 removed; all changes are the 5 audit site conversions + test updates |
+| New tests would have caught the original bug | PASS | `test_insert_with_audit_does_not_block_under_concurrent_writes` fires 10 concurrent inserts with 10s timeout; would have failed under the blocking `log_event()` |
+| Integration smoke tests pass | PASS | 53 integration tests pass across export, pipeline_e2e, migration, and infra suites; import_integration failures are pre-existing GH #303 |
+| xfail markers have corresponding GH Issues | PASS | No new xfail markers added in this PR |
+| Knowledge stewardship — Phase 1 investigator report | PASS | Addendum comment `IC_kwDORTRSjM7zfnx9` contains `## Knowledge Stewardship` with `Queried:` and `Stored:` entries |
+| Knowledge stewardship — Phase 2 rust-dev report | PASS | Addendum comment `IC_kwDORTRSjM7zfn4T` contains `## Knowledge Stewardship` with `Queried:` and `Stored:`/`Declined:` entries |
+
+## Detailed Findings
+
+### Fix Addresses Root Cause — All 5 Sites Converted
+
+**Status**: PASS
+
+**Evidence**: `git show c7b83a2 -- crates/unimatrix-server/src/server.rs` and `background.rs` show all 5 sites converted:
+
+- `server.rs:insert_with_audit()` — `self.audit.log_event(...)?` → `tokio::spawn(async move { let _ = audit.log_event_async(...).await; })`
+- `server.rs:correct_with_audit()` — same conversion
+- `server.rs:quarantine_or_restore_with_audit()` — same conversion
+- `background.rs:emit_tick_skipped_audit()` — same conversion
+- `background.rs:emit_auto_quarantine_audit()` — same conversion
+
+Residual `log_event()` calls at server.rs lines 403 and 406 are inside the pre-existing `audit_fire_and_forget()` helper which uses `spawn_blocking` (not `block_in_place`). This helper was not changed by this PR and is not part of the reported 5 sites.
+
+### No Placeholder Functions
+
+**Status**: PASS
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — bugfix-308-security-reviewer

**Risk Level: low. No blocking findings.**

---

### Finding 1: Silent audit loss on three server.rs sites (informational)

The three `server.rs` fire-and-forget spawns silently discard errors:

```rust
let _ = audit.log_event_async(audit_event_with_target).await;
```

The two `background.rs` sites correctly emit a `tracing::warn!` on failure. The server.rs sites do not. Under the previous (pre-PR) code, audit failures were propagated to the caller via `?`. With fire-and-forget, there is no caller to propagate to, so the pattern should include a warn log to preserve observability.

**Recommendation** (non-blocking): mirror the `background.rs` pattern:
```rust
if let Err(e) = audit.log_event_async(event).await {
    tracing::warn!(error = %e, "...: failed to write audit event");
}
```

This is not a security gate blocker — the previous code also had no guaranteed delivery once the write pool was contended. It is a marginal observability regression worth tracking.

---

### Finding 2: Two co-existing audit dispatch strategies (informational)

After this fix, two fire-and-forget patterns co-exist:

1. `audit_fire_and_forget()` (`server.rs:399`) — 7 callers in `mcp/tools.rs`. Uses `spawn_blocking` + `log_event()` (which calls `block_in_place` internally). While `spawn_blocking` moves the work off the async thread, `log_event()` + `block_in_place` inside a blocking thread can still contend on the single write connection during the analytics drain window.
2. `tokio::spawn(async move { log_event_async() })` — the new pattern from this PR, used at the 5 fixed sites.

The 7 `mcp/tools.rs` callers are **outside the stated scope of this PR** and this is pre-existing behaviour. However, they represent a residual starvation risk if the analytics drain task is active. Recommend a follow-up issue to convert `audit_fire_and_forget` callers to the async pattern.

---

### Finding 3: Test sleep(50ms) timing dependency (informational)

Four existing tests now use `tokio::time::sleep(Duration::from_millis(50))` to allow spawned audit tasks to commit before asserting log contents. This is a time-based synchronisation pattern. On a severely loaded CI runner 50ms could be insufficient. The two new regression tests correctly use `yield_now()` instead. Acceptable as-is; track as a test-reliability improvement.

---

### OWASP Summary

| Category | Result |
|----------|--------|
| Injection | No new SQL construction; parameterisation unchanged |
| Access control | No change to trust boundaries or capability checks |
| New dependencies | None (`Cargo.toml`/`Cargo.lock` not modified) |
| Deserialization | No new deserialization of external input |
| Input validation | No new external inputs; AuditEvent fields are in-process only |
| Secrets | No hardcoded credentials or tokens |

**Blast radius**: audit gap without data corruption or service disruption. Data commits succeed before the audit task is spawned; a panic in the audit task cannot corrupt data or escalate privilege.

**Verdict**: approve. The fix is minimal, correct, and consistent with the established GH #302 pattern. The three informational findings are pre-existing weaknesses made slightly more visible by this PR, not new vulnerabilities.

PR #343, opened 2026-03-22
fix: outcome/phase length check uses char count not byte count (#340)

### LINKED ISSUE #340: crt-025 follow-up: outcome length check should use char count not byte count
## Context

During the crt-025 security review, `validation.rs` was found to use `s.len()` (byte count) for the 512-character `outcome` limit instead of `s.chars().count()`.

For ASCII input this is equivalent. For multibyte input (emoji, CJK) the effective limit is lower than documented. No security risk — parameterized binds prevent injection and the behavior errs on rejecting more input, not less.

## Fix

```rust
// validation.rs — replace byte-count with char-count
if s.chars().count() > MAX_OUTCOME_LEN {
    return Err(format!("outcome exceeds {} characters", MAX_OUTCOME_LEN));
}
```

Apply the same pattern to `phase` and `next_phase` if they also use `s.len()`.

## Severity

Low / informational. No blocking behavior.

Depends on: crt-025 merged first.

dug-21:
## Summary

- Replace `s.len()` (byte count) with `s.chars().count()` at 3 sites in `infra/validation.rs`: `check_length` (all general string fields), `validate_cycle_params` (outcome), and `validate_phase_field` (phase/next_phase)
- For ASCII input behaviour is identical; multibyte input (emoji, CJK) now correctly respects the documented character limit
- `is_valid_feature_id` (line ~369) left unchanged — ASCII-only context, `.len()` is correct there

## Test plan

- [ ] 8 new multibyte boundary tests added (at-max and over-max for check_length, outcome, phase, next_phase)
- [ ] Full workspace unit suite: ~2929 passed, 0 failed
- [ ] Clippy: 0 new errors in changed crate
- [ ] Integration smoke: 20/20 passed
- [ ] Integration cycle/validation/unicode suites: 35/35 passed
- [ ] Gate: Bug Fix Validation — PASS (10/10 checks)

Closes #340

### DIFF
--- a/product/features/bugfix-340/agents/bugfix-340-security-reviewer-v2-report.md
+++ b/product/features/bugfix-340/agents/bugfix-340-security-reviewer-v2-report.md
@@ -1,0 +1,146 @@
+# Security Review: bugfix-340-security-reviewer-v2
+
+## Risk Level: low
+
+## Summary
+
+The expanded diff (v2) adds two security-adjacent gap fixes on top of the core `chars().count()` fix:
+control-character rejection for the outcome field and an ASCII-only allowlist for phase/next_phase
+fields. Both additions tighten security posture. No validation was loosened, no new dependencies
+were introduced, and no hardcoded secrets are present. One informational finding on the U+007F (DEL)
+boundary is documented below; it is consistent with the existing `check_control_chars` function and
+is not a regression. No blocking findings.
+
+---
+
+## Findings
+
+### Finding 1 — U+007F (DEL) not covered by outcome control-char check
+
+- **Severity**: informational
+- **Location**: `validation.rs:434` (outcome inline check)
+- **Description**: The new inline check `(c as u32) <= 0x1F` covers the 32 C0 control characters
+  (U+0000–U+001F) but does not cover U+007F (DEL, decimal 127), which is also defined as a control
+  character in ISO 6429 / ASCII. DEL is above the 0x1F boundary and would pass this check. The
+  pre-existing `check_control_chars` function (line 56) uses the same boundary (`code <= 0x1F`), so
+  the gap is consistent across the codebase — this is not a regression introduced by the PR. DEL in
+  a stored outcome value has no known injection or privilege-escalation vector in the current
+  architecture (SQLite parameterized inserts, internal read-back only).
+- **Recommendation**: No immediate action. If a future audit tightens control-character hygiene
+  codebase-wide, both the inline outcome check and `check_control_chars` should be extended to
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review v2 — bugfix-340 (expanded diff)

**Risk level: low. No blocking findings.**

This is a fresh-context re-review of the expanded diff, which now includes the two gap fixes (outcome control-char rejection, phase ASCII allowlist) in addition to the original `chars().count()` substitutions.

---

### What was reviewed

1. `check_length` (line 40): `value.chars().count() > max` — correct Unicode scalar count.
2. `validate_cycle_params` outcome (lines 431–436): `chars().count()` length check + inline control-char rejection for `<= U+001F`.
3. `validate_phase_field` (lines 462–472): `chars().count()` length check + ASCII allowlist `[a-z0-9\-_]`.
4. 14 new/updated tests covering multibyte boundaries, control-char rejection, ASCII allowlist.

---

### OWASP findings (informational only)

**U+007F (DEL) not covered by the outcome control-char check**

The inline check `(c as u32) <= 0x1F` covers C0 controls (U+0000–U+001F) but not U+007F (DEL). The pre-existing `check_control_chars` helper at line 56 uses the same boundary — consistent across the codebase, not a regression. DEL in a stored outcome has no injection vector (parameterized SQLite inserts, internal read-back only). If a future hardening pass extends control-char coverage, both the inline check and `check_control_chars` should be updated together (`code == 0x7F`).

**Phase ASCII allowlist confirmed correct**

`c.is_ascii_alphanumeric() || c == '-' || c == '_'` post-`.to_lowercase()` is a strict positive allowlist. Correctly rejects emoji, non-ASCII Latin, spaces, punctuation, and U+007F. Slug patterns used by the protocol ("discovery", "phase-end", "delivery") pass. No issues found.

**Outcome check is inline rather than using the shared helper**

Expected: `validate_cycle_params` returns `Result<_, String>` (not `ServerError`), so the `check_control_chars` helper cannot be called directly. The inline reimplementation is behaviorally identical. Not a security concern.

---

### Blast radius

Worst case for a subtle gap-fix bug: a control character or non-ASCII character reaches OUTCOME_INDEX columns. No SQL injection vector (parameterized queries), no rendering path. Bounded to data quality in an internal store.

---

### Regression risk

Low. All changes are additive (new rejection rules). No validation removed. All protocol-defined phase slugs ("discovery", "phase-end", "delivery") are ASCII-safe and pass the new allowlist. Gate-v2 confirms 144 integration tests pass.

---

Full report: `product/features/bugfix-340/agents/bugfix-340-security-reviewer-v2-report.md`

Issue #351, opened 2026-03-23
fix(observe): extraction pipeline creates noise instead of deprecating dead knowledge


dug-21:
## Problem

The background extraction pipeline has two rules that generate low-value \`lesson-learned\` entries at scale, polluting the knowledge base with noise instead of actionable lessons.

### 1. \`DeadKnowledgeRule\` — self-replicating noise (~1,925 entries, 87% of all lesson-learned)

**File:** \`crates/unimatrix-observe/src/extraction/dead_knowledge.rs\`

Every 15-minute background tick, for every active entry with \`access_count > 0\` that wasn't seen in the last 5 sessions, the rule creates a new \`lesson-learned\` entry:

\`\`\`
Title: "Possible dead knowledge: {entry title}"
Content: "Entry '...' has N accesses but was not used in the last 5 sessions. Consider deprecating."
Category: lesson-learned
Topic: knowledge-management
\`\`\`

**The problem:** This is a maintenance action signal, not a lesson. The rule tells you to deprecate an entry but never actually deprecates it. With 2,802 active entries, 96 of these are created per tick. Dead-knowledge lesson entries themselves accumulate, become dead knowledge, and trigger more dead-knowledge lessons — a self-replicating noise loop.

**The fix:** Remove \`DeadKnowledgeRule\` from \`default_extraction_rules()\` and add deprecation logic to the background tick instead. When the rule detects dead knowledge, it should call \`store.deprecate(id)\` on the flagged entries — not insert new lesson entries. The effectiveness analysis in \`context_status\` already surfaces unmatched/settled entries, making the lesson-entry approach redundant.

### 2. \`RecurringFrictionRule\` — thin, session-UUID-filled entries (~114 entries)

**File:** \`crates/unimatrix-observe/src/extraction/recurring_friction.rs\`

When a detection rule fires in 3+ sessions, the rule creates:

\`\`\`
Title: "Recurring friction: {rule_name}"
Content: "Detection rule '{rule_name}' fired in N sessions: [uuid, uuid, uuid, ...]"
Category: lesson-learned
Topic: process-improvement
\`\`\`

**The problem:** The content is a raw list of session UUIDs — not actionable. There is no description of what the friction IS, no root cause, no remediation. The same friction type (e.g., \`permission_retries\`) generates a new entry every time the session count increases, with no duplicate check against existing entries.

**The fix:** Two options (implement both):
[...]

PR #368, opened 2026-03-24
fix(observe): widen dead-knowledge session window to prevent over-deprecation (#367)

### LINKED ISSUE #367: fix(observe): dead_knowledge_deprecation_pass window too narrow — over-deprecating valid entries
## Problem

The \`dead_knowledge_deprecation_pass()\` introduced in #351 is over-aggressively deprecating valid knowledge base entries.

**Root cause**: \`detect_dead_knowledge_candidates\` is called with \`window=5\` — any entry not accessed in the **last 5 sessions** is flagged as dead knowledge and deprecated. At 50 deprecations/tick every 15 minutes, this sweeps up valid ADRs, conventions, and patterns that are simply not accessed every session.

**Impact**:
- Deprecated count before #351 merge: **216**
- Deprecated count now: **911**
- Incorrectly deprecated by tick pass: **~495 entries**

The mismatch: \`DEAD_KNOWLEDGE_SESSION_THRESHOLD = 20\` (sessions fetched) vs \`window = 5\` (sessions that define "recently used"). An entry not touched in the last 5 sessions is deprecated even if it's a valid ADR accessed 6 sessions ago.

## Immediate Fix

Set \`DEAD_KNOWLEDGE_SESSION_THRESHOLD\` (and the \`window\` argument passed to \`detect_dead_knowledge_candidates\`) to **1000** to halt further incorrect deprecations while the restoration pass runs.

\`\`\`rust
// crates/unimatrix-server/src/background.rs
const DEAD_KNOWLEDGE_SESSION_THRESHOLD: usize = 1000;  // was 20

// call site — must match:
detect_dead_knowledge_candidates(&observations, &store_for_detection, 1000)  // was 5
\`\`\`

## Follow-on

After deploying the threshold change, run a restoration pass to reactivate all entries incorrectly deprecated by the tick pass since the #351 merge. Entries with topic=\`knowledge-management\` and tag=\`dead-knowledge\` (the old noise entries, deprecated by the one-shot migration) should remain deprecated. All others deprecated after the #351 merge should be restored to Active.

## Out of scope

- Re-evaluating the correct long-term threshold (separate issue once the knowledge base is restored)
- Changes to the one-shot migration or RecurringFrictionRule fixes (those are correct)

dug-21:
## Summary

- Sets `DEAD_KNOWLEDGE_SESSION_THRESHOLD` from `20` → `1000` in `background.rs` to halt over-deprecation of valid entries
- Passes `DEAD_KNOWLEDGE_SESSION_THRESHOLD` at the `detect_dead_knowledge_candidates` call site instead of the hardcoded `5`, eliminating the two-constant mismatch class

## Root Cause

`dead_knowledge_deprecation_pass()` fetched 20 sessions but used a hardcoded `window=5` to define "recently used" — any entry accessed in sessions 6–20 was flagged as dead and deprecated. At 50 deprecations/tick × 96 ticks/day, this silently swept valid ADRs and patterns. Impact: ~495 entries incorrectly deprecated since #351 merged.

## Deployment Note

This PR halts further incorrect deprecations. A separate restoration pass to recover the ~495 affected entries will be tracked in a follow-up issue.

## Test Plan

- [x] Full workspace unit tests: 1908 passed, 0 failed
- [x] No new clippy warnings (19 pre-existing in unimatrix-store, unaffected)
- [x] Existing `test_dead_knowledge_deprecation_pass_unit` updated to use `DEAD_KNOWLEDGE_SESSION_THRESHOLD + 1`
- [x] Gate 3 validation: PASS

Closes #367

### DIFF
--- a/product/features/bugfix-367/agents/bugfix-367-security-reviewer-report.md
+++ b/product/features/bugfix-367/agents/bugfix-367-security-reviewer-report.md
@@ -1,0 +1,76 @@
+# Security Review: bugfix-367-security-reviewer
+
+## Risk Level: low
+
+## Summary
+
+The diff consists of two constant changes and one test update in `background.rs`. No new logic, no new external inputs, no new SQL interpolation, no new dependencies. The primary security-relevant question is whether raising `DEAD_KNOWLEDGE_SESSION_THRESHOLD` from 20 to 1000 breaks the SQLite bind-parameter limit for the dynamic IN-clause in `fetch_recent_observations_for_dead_knowledge`. It does not: the bundled `libsqlite3-sys` compiles with `SQLITE_MAX_VARIABLE_NUMBER = 32766`, making a 1000-parameter IN-clause safe. All other OWASP concerns reviewed below are unchanged from the pre-fix state (not introduced by this diff).
+
+## Findings
+
+### Finding 1: Dynamic SQL IN-clause at 1000 bind parameters — within SQLite bundled limit
+- **Severity**: low (informational, not a defect)
+- **Location**: `crates/unimatrix-server/src/background.rs` lines 931–946 (`fetch_recent_observations_for_dead_knowledge`)
+- **Description**: `DEAD_KNOWLEDGE_SESSION_THRESHOLD` now controls both the `LIMIT ?1` in Step A (capped at 1000 session IDs) and the number of `?N` bind parameters in the Step B IN-clause. The session IDs returned by Step A are database-internal values (fetched from the `observations` table `session_id` column) and are never user-supplied strings — they are bound as parameterised values, not interpolated into the SQL text. This is not an injection vector.
+
+  The concern evaluated here is the SQLite bind-variable limit. The prior security review for bugfix-351 (finding 4) noted the IN-clause was bounded at 20 parameters and was safe against the common 999-parameter assumption. With this fix the bound rises to 1000. Verified: the bundled `libsqlite3-sys-0.30.1` compiles `sqlite3/sqlite3.c` with `#define SQLITE_MAX_VARIABLE_NUMBER 32766`. A 1000-parameter IN-clause is well within this limit.
+
+  No SQL injection surface exists. No limit violation.
+- **Recommendation**: None required. The bound is correct and safe. Optionally document the 32766 limit vs the 1000 constant in a comment near `DEAD_KNOWLEDGE_SESSION_THRESHOLD` for future maintainers.
+- **Blocking**: no
+
+### Finding 2: `extract_entry_ids` parses `response_snippet` from database rows — no injection surface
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review: bugfix-367

**Risk Level: low — no blocking findings**

### Scope

This review covers the three-line diff in `background.rs`: raising `DEAD_KNOWLEDGE_SESSION_THRESHOLD` from 20 to 1000, passing that constant to `detect_dead_knowledge_candidates` instead of the previous hardcoded `5`, and the corresponding test update.

### OWASP Assessment

| Concern | Status | Notes |
|---------|--------|-------|
| SQL injection | clear | Session IDs in the dynamic IN-clause are database-internal values fetched via parameterised query (Step A), then bound as `?N` parameters in Step B. No user-controlled input is interpolated into SQL text. |
| Bind-parameter limit | clear | The bundled `libsqlite3-sys-0.30.1` compiles sqlite3 with `SQLITE_MAX_VARIABLE_NUMBER = 32766`. A 1000-parameter IN-clause is well within this limit. (The prior review for bugfix-351 noted the 999-parameter assumption from older/system SQLite; that assumption does not apply to the bundled build.) |
| Access control | clear | No trust boundary changes. The deprecation pass runs server-side under the maintenance tick; no external inputs gate entry into this path. |
| Deserialization | clear | `extract_entry_ids` parses `response_snippet` strings via text splitting and `parse::<u64>()`. Extracted IDs flow only into a `HashSet<u64>` used as a filter set — never back into SQL. |
| Hardcoded secrets | clear | No credentials, tokens, or keys in the diff. |
| New dependencies | clear | No `Cargo.toml` changes. |
| Unsafe / unwrap | clear | Gate 3b confirms zero new `unsafe` blocks and zero new `.unwrap()` in production paths. |

### Blast Radius

Worst case: over-deprecation of up to 50 active entries per maintenance tick (the existing `DEAD_KNOWLEDGE_DEPRECATION_CAP`). Status changes are reversible via `context_correct`. No data deletion occurs. If the SQLite query fails (e.g. at extreme session counts approaching 32766), the pass falls back to `return vec![]` and skips that tick — no crash, no corruption.

### One Informational Note

At 1000 sessions the observation fetch in Step B could return a large row set on a high-activity instance (1000 sessions × N observations per session). The blocking thread pool absorbs the cost, but this is a latent scaling concern to revisit if observation volume grows significantly. Not a blocker.

### Conclusion

No blocking findings. The fix is correct, minimal in scope, and safe. The SQLite bind-parameter limit is verified against the actual bundled library. Recommend merge.

---
Reviewed by: `bugfix-367-security-reviewer` (uni-security-reviewer)
Report: `product/features/bugfix-367/agents/bugfix-367-security-reviewer-report.md`

PR #462, opened 2026-03-31
fix: orphaned-edge compaction does not clean GRAPH_EDGES for quarantined entries (#458)

### LINKED ISSUE #458: bugfix: orphaned-edge compaction does not clean GRAPH_EDGES for quarantined entries
## Problem

The orphaned-edge compaction step in `run_single_tick` (`background.rs`) deletes `GRAPH_EDGES` rows only when the referenced `source_id` or `target_id` no longer exists in the `entries` table (fully deleted). Quarantined entries (status=3) remain in the `entries` table, so their edges are never cleaned up.

On every tick, `build_typed_relation_graph` builds an active-entries-only snapshot (status=0). When it encounters `GRAPH_EDGES` rows whose endpoints are quarantined, it logs:

```
WARN build_typed_relation_graph: source_id not in entries snapshot, skipping edge source_id=X target_id=Y relation_type=CoAccess
```

These warns fire every 15 minutes indefinitely for every quarantined entry that has graph edges. Currently 2,491 quarantined entries are generating this noise on every tick.

## Root Cause

The orphaned-edge compaction SQL filters on `NOT IN (SELECT id FROM entries)` — entries that don't exist at all. Quarantined entries DO exist in the table (status=3) so the condition is never true for them.

The prune pass (Step 0a of `run_maintenance`) already removes quarantined entries from VECTOR_MAP/HNSW. The orphaned-edge compaction should follow the same logic for GRAPH_EDGES.

## Fix

Extend the orphaned-edge compaction query to also delete edges where the source or target entry has status = Quarantined (3):

```sql
DELETE FROM graph_edges
WHERE source_id NOT IN (SELECT id FROM entries WHERE status != 3)
   OR target_id NOT IN (SELECT id FROM entries WHERE status != 3)
```

Or equivalently — delete edges where either endpoint is absent OR quarantined.

This is consistent with the existing prune pass behavior: quarantined entries are removed from all index structures (VECTOR_MAP, HNSW). GRAPH_EDGES should follow.

## Impact

- Warn flood suppressed: ~2,491 quarantined entries currently generating warns every tick
- GRAPH_EDGES table cleaned of dead weight
- No behavior change to active or deprecated entries

## Notes

- Deprecated entries (status=1) currently also generate warns if they have graph edges. Whether deprecated entries should retain their edges (they can be un-deprecated) is a design decision outside this bugfix. The immediate fix targets quarantined entries only, matching the prune pass behavior.
- Related: the `build_typed_relation_graph` warn itself is correct behavior — it is successfully skipping the invalid edges. This fix eliminates the source rather than suppressing the warning.

dug-21:
## Summary

- Extends orphaned-edge compaction SQL in `run_single_tick` to also delete `GRAPH_EDGES` rows where `source_id` or `target_id` references a quarantined entry (status=3)
- Eliminates WARN flood from `build_typed_relation_graph` (~2,491 quarantined entries generating noise every 15-minute tick)
- Consistent with the existing prune pass behaviour: quarantined entries are removed from VECTOR_MAP/HNSW; GRAPH_EDGES now follows the same rule

## Root Cause

SQL used `NOT IN (SELECT id FROM entries)` — matching only fully-deleted entries. Quarantined entries exist in the table so their edges survived indefinitely.

**Fix** — bound parameter form (matching prune pass convention):
```sql
DELETE FROM graph_edges
WHERE source_id NOT IN (SELECT id FROM entries WHERE status != ?1)
   OR target_id NOT IN (SELECT id FROM entries WHERE status != ?1)
```

## Test plan

- [ ] `test_background_tick_compaction_removes_quarantined_source_edges` — CoAccess + Supports edges with quarantined source deleted
- [ ] `test_background_tick_compaction_removes_quarantined_target_edges` — CoAccess + Supports edges with quarantined target deleted
- [ ] Full workspace test suite: 0 failures
- [ ] Integration smoke: 22 passed, 0 failed
- [ ] Integration lifecycle: 41 passed, 2 xfailed (pre-existing)
- [ ] Clippy: no new warnings from this change

Closes #458

### DIFF
--- a/product/features/bugfix-458/agents/bugfix-458-security-reviewer-report.md
+++ b/product/features/bugfix-458/agents/bugfix-458-security-reviewer-report.md
@@ -1,0 +1,71 @@
+# Security Review: bugfix-458-security-reviewer
+
+## Risk Level: low
+
+## Summary
+
+The diff contains a single-file SQL maintenance fix to `background.rs`. The change adds a `WHERE status != ?1` predicate to a parameterized DELETE and binds `Status::Quarantined as u8 as i64` — no user-controlled input reaches either the query or the parameter. No new dependencies, no unsafe code, no secrets. All OWASP concerns checked; none apply.
+
+## Findings
+
+### F-1: SQL Query Binding — Correct Pattern Confirmed
+- **Severity**: informational (no issue)
+- **Location**: `crates/unimatrix-server/src/background.rs:519–526`
+- **Description**: The new predicate `WHERE status != ?1` is a parameterized query. The bound value `Status::Quarantined as u8 as i64` is a compile-time constant derived from a `#[repr(u8)]` enum. No runtime user input participates in either the query string or the bound parameter. This is the established codebase pattern (consistent with `services/status.rs:1023`). Previous security review of bugfix-444 (Unimatrix #3766) established that bare integer literals in SQL status filters would be blocking; this fix uses the typed bind pattern correctly.
+- **Recommendation**: No action required.
+- **Blocking**: no
+
+### F-2: Test-Only `.expect()` in Test Helper
+- **Severity**: low (test code only)
+- **Location**: `crates/unimatrix-server/src/background.rs:2945`
+- **Description**: `run_graph_edges_compaction` uses `.expect("compaction DELETE must succeed")`, which panics on failure. This is existing behavior in the test helper, not introduced by this diff — the test helper existed before this PR. Test panics are acceptable and expected in unit tests; the production path at line 519–548 correctly uses `match` with a logged non-fatal error path.
+- **Recommendation**: No action required. Production error handling is correct.
+- **Blocking**: no
+
+### F-3: Test INSERT Uses Raw SQL With Full Column List
+- **Severity**: informational (test code only)
+- **Location**: `crates/unimatrix-server/src/background.rs:3107–3145, 3168–3205`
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — bugfix-458

**Risk level: low. No blocking findings. Clear to merge.**

### Scope

Single file changed: `crates/unimatrix-server/src/background.rs`. Changes are: 2 production SQL lines, 4 test helper SQL lines + sync comment, 114 lines of new tests.

### OWASP Checks

| Concern | Verdict |
|---------|---------|
| SQL injection (A03) | No risk — parameterized query (`?1`); bound value is a compile-time enum constant, no external input |
| Broken access control (A01) | No risk — internal background maintenance tick, no trust boundary |
| Input validation (A03) | No risk — `Status::Quarantined as u8 as i64` is compile-time verified via `#[repr(u8)]` enum |
| Vulnerable components (A06) | No risk — no new dependencies |
| Deserialization (A08) | No risk — no deserialization in changed paths |
| Secrets | None found |

### SQL Binding Pattern

The fix uses `.bind(Status::Quarantined as u8 as i64)` — the established codebase standard confirmed in `services/status.rs:1023` and required per the convention established in bugfix-444 review. Previous security review (Unimatrix #3766) noted that bare integer literals in SQL status filters would be blocking; this fix uses the typed pattern correctly.

### Blast Radius

Worst case for a subtle regression: the DELETE removes too many edges, producing a thinner graph for one tick cycle. Not a panic, not data corruption — the tick error path is non-fatal (lines 539–547) and the typed graph rebuilds from live `entries` every tick.

### Regression Risk

Low. The predicate `status != 3` includes Active (0), Deprecated (1), and Proposed (2) — standard orphaned-edge deletion is preserved. Pre-existing test `test_background_tick_compacts_orphaned_graph_edges` continues to pass.

### Minor Observations (non-blocking)

- **F-3**: The two new test INSERTs use raw SQL with an explicit 23-column list. This is consistent with existing test patterns in the file and not a security issue. Future agents adding columns to `entries` should grep for raw INSERT helpers in tests (per Unimatrix lesson #3543).

---

Full report: `product/features/bugfix-458/agents/bugfix-458-security-reviewer-report.md`

Issue #477, opened 2026-04-01
hygiene: add quarantine guard at co_access write time to prevent stale-pair accumulation


dug-21:
## Context

Identified during bugfix #476 audit (co_access promotion quarantine oscillation).

## Problem

`execute_analytics_write` in `analytics.rs` (the `AnalyticsWrite::CoAccess` match arm) has no quarantine status filter at write time. When an entry is quarantined, it stops being returned from search so new co_access pairs involving it naturally stop accumulating — but any co_access rows that exist at quarantine time continue aging in place until `cleanup_stale_co_access` prunes them by `last_updated`.

The correctness fix (PR for #476) moves the quarantine filter to the promotion SELECT, which is the authoritative gate. The write-time gap is a hygiene issue only: stale pairs accumulate longer than necessary and consume batch slots during the cleanup lag.

## Fix (when scoped)

Add a quarantine pre-check at the `AnalyticsWrite::CoAccess` write path. The safe form is a read against an in-memory quarantine set or status cache — **not** a live SQL JOIN in the analytics drain (hot-path risk, see crt-014 lesson).

A bloom filter or a cached `HashSet<i64>` of quarantined IDs rebuilt by the background tick would be appropriate.

## Priority

Low. The tick-side JOIN (bugfix #476) is the correctness fix. This is cleanup-lag reduction only.

## References

- Bugfix #476 (root cause fix)
- Lesson #3979 (co_access write-time path has no quarantine guard; asymmetry must be documented at write site)
- crt-014 lesson (Store reads on hot path cause pool saturation)

dug-21:
## Resolution: WONTFIX

After design review during bugfix-471, this issue is resolved as WONTFIX.

**Rationale:** Entry #3979 in Unimatrix records the explicit architectural decision to leave the co_access write path unguarded, with the tick-side JOIN in `co_access_promotion_tick.rs` as the authoritative correctness gate. Threading an `Arc<RwLock<HashSet<i64>>>` from the background tick into the drain task would require changing the `spawn_drain_task` signature in `unimatrix-store` and crossing the `unimatrix-server` → `unimatrix-store` crate boundary — a real architectural cost for hygiene-only benefit.

The stale-pair accumulation window (quarantine event → `cleanup_stale_co_access` prune) is an accepted trade-off. The tick-side JOIN ensures no quarantined-endpoint pair ever promotes to `GRAPH_EDGES`.

[...]

PR #527, opened 2026-04-06
fix(compaction): use allowlist WHERE status = Active so deprecated-endpoint edges are deleted (#471)

### LINKED ISSUE #471: fix: orphaned-edge compaction should also remove edges for deprecated entries
## Bug

The orphaned-edge compaction tick removes GRAPH_EDGES rows where either endpoint is quarantined or non-existent, but not where either endpoint is **deprecated**. Deprecated entries are excluded from the active entries snapshot used by `build_typed_relation_graph`, so their edges produce a WARN every tick indefinitely:

```
WARN unimatrix_engine::graph: build_typed_relation_graph: target_id not in entries snapshot, skipping edge source_id=141 target_id=1045 relation_type=CoAccess
```

The same 32 rows trigger this WARN on every background tick with no cleanup mechanism.

## Root Cause

The compaction SQL in `background.rs` targets only quarantined/missing entries:

```sql
DELETE FROM graph_edges
WHERE source_id NOT IN (SELECT id FROM entries WHERE status != ?)
   OR target_id NOT IN (SELECT id FROM entries WHERE status != ?)
-- bind: Status::Quarantined
```

`SELECT id FROM entries WHERE status != Quarantined` returns Active + Deprecated entries, so deprecated entry edges are never deleted.

## Fix

Extend the compaction to also remove edges where either endpoint is deprecated. Change the bound status to include both Quarantined and Deprecated, or rewrite as an allowlist (Active only):

```sql
DELETE FROM graph_edges
WHERE source_id NOT IN (SELECT id FROM entries WHERE status = ?)
   OR target_id NOT IN (SELECT id FROM entries WHERE status = ?)
-- bind: Status::Active
```

This removes edges for any entry that is not active — quarantined, deprecated, or non-existent.

## Trade-off

If a deprecated entry is ever un-deprecated (restored to Active), its GRAPH_EDGES rows will have been removed and will not be automatically rebuilt. The co_access promotion tick will eventually re-promote CoAccess edges for the restored entry as it accumulates co-access signal. Other edge types (Supports, Contradicts, Informs) would need to wait for the next NLI detection tick to re-infer them.

This is acceptable — deprecated entries rarely return to Active, and the co_access and NLI ticks are designed to rebuild graph state from the ground up.

## Acceptance

- The 32 WARN rows no longer appear after the next compaction tick
- Deprecated entry edges are deleted during the same compaction pass
- A newly-deprecated entry's edges are removed on the following tick
- Tests confirm the compaction removes edges for deprecated (not just quarantined/missing) endpoints

dug-21:
## Summary

- Compaction DELETE in `background.rs` used denylist form (`WHERE status != Quarantined`), which included `Deprecated` entries in the protected set — deprecated-endpoint edges were never cleaned up, causing a WARN flood in `build_typed_relation_graph` on every tick
- Changed to allowlist form (`WHERE status = Active`) in both the production SQL and the identical test-helper mirror — future status values are excluded by default, no SQL update needed when new statuses are added
- Added 4 new tests covering all combinations of deprecated-source/target × CoAccess/Supports
- Updated comment at `analytics.rs:392–398` to remove stale `GH #477` forward reference; cites Unimatrix entry #3979 as the rationale for the intentionally unguarded write path
- GH #477 closed as WONTFIX (write path unguarded by deliberate ADR decision, tick-side JOIN is the correctness gate)

## Test plan

- [x] `test_compaction_removes_deprecated_source_co_access` — pass
- [x] `test_compaction_removes_deprecated_target_co_access` — pass
- [x] `test_compaction_removes_deprecated_source_supports` — pass
- [x] `test_compaction_removes_deprecated_target_supports` — pass
- [x] All existing compaction tests (quarantine cases) continue to pass
- [x] Full workspace: 4539/4539 unit tests pass
- [x] Integration smoke gate: 22/22 pass
- [x] Clippy clean in changed crates (`unimatrix-server`, `unimatrix-store`)
- [x] Gate 3c: PASS — report at `product/features/bugfix-471/reports/gate-3c-report.md`

Closes #471

### DIFF
--- a/product/features/bugfix-471/agents/bugfix-471-security-reviewer-report.md
+++ b/product/features/bugfix-471/agents/bugfix-471-security-reviewer-report.md
@@ -1,0 +1,70 @@
+# Security Review: bugfix-471-security-reviewer
+
+## Risk Level: medium
+
+## Summary
+
+The core compaction fix is correct: switching `status != Quarantined` to `status = Active` in the GRAPH_EDGES DELETE properly closes the deprecated-endpoint gap. However, a medium-severity finding exists: the co_access promotion tick (`co_access_promotion_tick.rs`) still uses the old denylist form `status != Quarantined` and is not covered by deprecated-endpoint tests. This creates an asymmetric predicate between compaction (allowlist) and promotion (denylist) — exactly the oscillation pattern documented in Unimatrix entry #3978. Deprecated-endpoint co_access pairs will be filtered by compaction each tick then re-promoted by the promotion tick on the next tick. The diff does not introduce security vulnerabilities, injection risks, or secrets.
+
+## Findings
+
+### Finding 1: Promotion tick uses denylist form — deprecated endpoints will re-promote after compaction deletes them
+- **Severity**: medium
+- **Location**: `crates/unimatrix-server/src/services/co_access_promotion_tick.rs:232-237,244`
+- **Description**: The compaction DELETE (just fixed by this PR) now uses the allowlist `status = Active`. The promotion tick SELECT uses `status != ?3` bound to `Status::Quarantined`. This denylist form passes Deprecated entries (status=1) through: a co_access pair where one endpoint is Deprecated will be excluded from GRAPH_EDGES by compaction, then re-inserted by the promotion tick on the same or next tick. This is not a new regression introduced by this PR — it is a pre-existing asymmetry. However, this PR explicitly fixes compaction to use the allowlist form and its gate-3c report characterises the fix as resolving deprecated-endpoint edge accumulation. With the asymmetric promotion tick still live, deprecated-endpoint edges will continue to accumulate in GRAPH_EDGES via the promotion path, making the fix incomplete at runtime. Unimatrix entry #3978 documents this exact oscillation pattern from bugfix-476: "compaction correctly deletes...promotion immediately re-inserts them." The promotion tick comment (line 216-220) states it "excludes quarantined-endpoint pairs" but makes no mention of deprecated endpoints.
+- **Recommendation**: The promotion tick SELECT should be updated to use `status = Active` (allowlist) on both endpoints and in the max_count subquery, replacing `status != ?3` / `Status::Quarantined`. This would be consistent with the compaction fix and with entry #4156's stated rule: "Every compaction or prune DELETE that references `entries.id` must use `WHERE status = Active` (allowlist)." The promotion tick SELECT is the mirror-image operation. Alternatively, if the intent is to allow deprecated-endpoint promotion (e.g., for supersedes-chain traversal reasons — see typed_graph.rs:101 comment), this must be explicitly documented and the compaction fix must be re-evaluated for consistency.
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — bugfix-471

**Risk level: medium** | Blocking findings: No

---

### Summary

The compaction SQL fix is correct and minimal. Switching `status != Quarantined` to `status = Active` in the GRAPH_EDGES DELETE is the right call per the allowlist rule in Unimatrix entry #4156. OWASP checks clean: parameterized queries throughout, no new input surfaces, no secrets, no new dependencies.

One medium finding requires a follow-up issue; two low findings are notes only.

---

### Finding 1 (medium — non-blocking, follow-up required): Promotion tick still uses denylist form

**Location**: `crates/unimatrix-server/src/services/co_access_promotion_tick.rs:232-237,244`

The compaction DELETE now uses `status = Active` (allowlist). The co_access promotion tick SELECT uses `status != ?3` bound to `Status::Quarantined` (denylist). Deprecated entries are `status = 1`, which passes the denylist filter — so the promotion tick will re-promote deprecated-endpoint co_access pairs into GRAPH_EDGES on the same or next tick after compaction deletes them.

This is the exact oscillation pattern in Unimatrix entry #3978 (bugfix-476): "compaction correctly deletes...promotion immediately re-inserts them." The comment at line 216-220 states it "excludes quarantined-endpoint pairs" with no mention of deprecated endpoints.

This is a pre-existing asymmetry, not introduced by this PR. However, this PR's stated goal is to fix deprecated-endpoint edge accumulation — at runtime the fix will have no net effect on deprecated edges until the promotion tick is also corrected.

**Recommended follow-up**: Update the promotion tick SELECT (and max_count subquery) to use `status = Active` (allowlist) on both endpoint JOINs, mirroring the compaction fix. A tracking issue should be filed.

---

### Finding 2 (low — note only): Promotion tick has no deprecated-endpoint test

**Location**: `crates/unimatrix-server/src/services/co_access_promotion_tick_tests.rs`

Entry #4156 requires tests for (a) fully absent, (b) quarantined, (c) deprecated, and (d) active endpoints for every compaction DELETE. The 4 new tests in `background.rs` satisfy this for compaction. The promotion tick tests have no case (c). A test asserting deprecated-endpoint pairs are NOT promoted would either pass after Finding 1 is fixed, or serve as a regression sentinel until then.

---

### Finding 3 (low — note only): typed_graph rebuild interaction undocumented

**Location**: `crates/unimatrix-server/src/services/typed_graph.rs:99-102`

`TypedGraphState::rebuild` retains Deprecated entries (for supersedes-chain traversal) but compaction now removes their GRAPH_EDGES. The interaction is correct but undocumented in the rebuild code. A one-line comment would prevent future confusion.

---

### Blast radius

If the `status = Active` bind produced the wrong integer, compaction would delete all GRAPH_EDGES each tick (immediately visible: `edges_inserted` explodes in logs) or fail silently (original bug persists, caught by new tests). Neither case is a security event — failure mode is graph scoring quality only.

---

### Verdict

No blocking findings. The fix is correct, minimal, and well-tested for the compaction path. Finding 1 should be tracked as a follow-up to complete the fix at the promotion tick.

PR #530, opened 2026-04-06
fix: co_access promotion tick allowlist — exclude all non-Active statuses (#528)

### LINKED ISSUE #528: fix: co_access promotion tick denylist (status != Quarantined) includes deprecated endpoints — follow-up to #471
## Context

Follow-up identified during security review of bugfix-471 (PR #527).

## Problem

`co_access_promotion_tick.rs:232-237` uses the denylist form:
```sql
JOIN entries ea ON ca.id_a = ea.id AND ea.status != ?3
JOIN entries eb ON ca.id_b = eb.id AND eb.status != ?3
-- ?3 = Status::Quarantined
```

This is the same denylist-vs-allowlist problem fixed by bugfix-471 in the compaction DELETE. Deprecated entries (status=1) pass the `!= Quarantined` filter, so co_access pairs whose endpoints are deprecated can be promoted to `GRAPH_EDGES` on the next tick.

After bugfix-471 ships: the compaction DELETE will correctly clean up deprecated-endpoint edges using `status = Active`. However, if the co_access table still has pairs for deprecated entries, the promotion tick will re-insert those edges before compaction runs. This creates a slow oscillation (insert on promotion tick → delete on compaction tick → insert again).

## Fix

Convert the promotion SELECT JOINs to allowlist form:
```sql
JOIN entries ea ON ca.id_a = ea.id AND ea.status = ?3
JOIN entries eb ON ca.id_b = eb.id AND eb.status = ?3
-- ?3 = Status::Active
```

Also add a test for deprecated-endpoint pair (case (c) from entry #4156): assert that a deprecated-endpoint co_access pair does not get promoted to `GRAPH_EDGES`.

## References

- bugfix-471 / PR #527 — compaction DELETE allowlist fix
- Unimatrix entry #4156 — allowlist rule and four-case test matrix
- Unimatrix entry #3978 — oscillation pattern lesson (bugfix-476)

dug-21:
## Summary

Fixes the denylist-vs-allowlist problem in the co_access promotion tick, the direct counterpart to the compaction DELETE fix in bugfix-471 (#527).

- **SQL fix**: Converts all four JOIN conditions in the batch SELECT from denylist form (`ea.status != ?3` bound to `Status::Quarantined`) to allowlist form (`ea.status = ?3` bound to `Status::Active`). This is the same pattern applied to the compaction DELETE in bugfix-471.
- **Scope of the fix**: The allowlist correctly excludes *all* non-Active statuses — `Deprecated=1`, `Proposed=2`, and `Quarantined=3` — not just Deprecated. The old denylist was silently passing Proposed entries through as well. Whether Proposed entries exist in practice today is unclear, but the allowlist is the correct defensive default.
- **Comment update**: Updates the associated comments at lines 215–224 to accurately describe the allowlist form, including the NULL failure mode (operator is now `=`, so a missing bind produces `status = NULL`, which — like `status != NULL` — is always NULL and silently promotes nothing).
- **New test**: `test_deprecated_endpoint_pair_not_promoted` (Group K) — seeds a deprecated-endpoint pair with a *higher* co_access count than the active pair. Primary assertion: active edge weight = 1.0 (not 0.5), confirming the subquery-side filter is also fixed. Without the subquery fix, max_count would be inflated by the deprecated pair and deflate the active edge weight.
- **Documentation**: Expands the comment in `typed_graph.rs` documenting why deprecated nodes are intentionally retained in the `TypedGraphState` snapshot (SR-01 Supersedes chain traversal), and that their having no edges post-compaction is expected and correct.

## Test plan

- [x] `test_deprecated_endpoint_pair_not_promoted` — weight assertion (primary) + edge count assertion (secondary)
- [x] All 35 promotion tick unit tests pass
- [x] Full workspace test suite: pass
- [x] Clippy: zero new warnings
- [x] Integration smoke: 22/22 pass
- [x] Integration lifecycle: 44 pass, 5 pre-existing xfail (unrelated tick-timing)
- [x] Gate 3b: PASS (15/15 checks)

Closes #528.

### DIFF
--- a/product/features/bugfix-528/agents/bugfix-528-security-reviewer-report.md
+++ b/product/features/bugfix-528/agents/bugfix-528-security-reviewer-report.md
@@ -1,0 +1,90 @@
+# Security Review: bugfix-528-security-reviewer
+
+## Risk Level: low
+
+## Summary
+
+The change is a surgical SQL filter flip in the co_access promotion tick: four JOIN conditions change from denylist (`status != Quarantined`) to allowlist (`status = Active`), and the bound parameter changes from `Status::Quarantined` to `Status::Active`. All OWASP-relevant categories were evaluated and no security concerns were identified. The change is logically correct, minimal, and consistent with the existing compaction DELETE which already uses the same allowlist pattern.
+
+## Findings
+
+### Finding 1 — SQL Parameterization (Injection)
+- **Severity**: low (informational — no issue found)
+- **Location**: `crates/unimatrix-server/src/services/co_access_promotion_tick.rs:232-251`
+- **Description**: All SQL parameters are bound via sqlx positional placeholders (`?1`, `?2`, `?3`). No string interpolation or format! macros are used. The status value is bound as `Status::Active as u8 as i64` — a compile-time constant, not derived from any external input. SQL injection risk is zero.
+- **Recommendation**: None — existing parameterization is correct.
+- **Blocking**: no
+
+### Finding 2 — Status Numeric Encoding Correctness
+- **Severity**: low (informational — no issue found)
+- **Location**: `crates/unimatrix-store/src/schema.rs:10-15` vs `co_access_promotion_tick.rs:251`
+- **Description**: Verified the `Status` enum directly: `Active = 0`, `Deprecated = 1`, `Proposed = 2`, `Quarantined = 3`. The bind `Status::Active as u8 as i64` resolves to `0i64` at compile time. The compaction DELETE in `background.rs:518` uses the identical binding pattern `Status::Active as u8 as i64`. Both sides of the compaction/promotion symmetry now use the same encoding, eliminating the oscillation root cause.
+- **Recommendation**: None — encoding is correct and consistent.
+- **Blocking**: no
+
+### Finding 3 — Blast Radius of Allowlist Flip
+- **Severity**: low
+- **Location**: `co_access_promotion_tick.rs:239-244`
+- **Description**: Worst case if the fix has a subtle bug: the tick silently promotes nothing (if `Status::Active` had the wrong value, `status = 0` would match no rows, `qualifying_count == 0`, tick returns early after logging). The failure mode is data starvation (no new graph edges), not data corruption. The existing SR-05 early-tick warn! at `current_tick < 5` would surface this signal loss within five ticks. Existing tests (35 co_access_promotion tests) cover the zero-qualifying-pairs early-return path.
+- **Recommendation**: None — failure mode is safe and observable.
+- **Blocking**: no
+
+### Finding 4 — typed_graph.rs Comment-Only Change
+- **Severity**: low (informational — no issue found)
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — bugfix-528 (uni-security-reviewer)

**Risk Level: low**

Reviewed cold from the diff. No blocking findings. Full report at `product/features/bugfix-528/agents/bugfix-528-security-reviewer-report.md`.

---

### What Was Checked

| Check | Result |
|-------|--------|
| SQL injection — parameterization | Clean. All four JOIN conditions use sqlx positional placeholders (`?3`); no string interpolation. `Status::Active as u8 as i64` is a compile-time constant, not externally supplied. |
| Status numeric encoding | Verified against schema.rs: `Active=0`, `Deprecated=1`, `Proposed=2`, `Quarantined=3`. Bind resolves to `0i64`. Consistent with compaction DELETE in background.rs:518 which uses the identical pattern. |
| Allowlist symmetry | Compaction DELETE (bugfix-471, background.rs:514-516) already uses `WHERE status = ?1` allowlist. This fix symmetrizes the promotion INSERT side. Both sides now use `Status::Active as u8 as i64`. Oscillation root cause is eliminated. |
| Blast radius | Worst case: wrong bind value → tick returns early (zero qualifying rows), SR-05 warn! fires within 5 ticks, graph edges stagnate but are not corrupted. No data deleted, no access widened. |
| Regression risk | Low. All 35 co_access_promotion tests pass, 2769 server unit tests pass, smoke + lifecycle integration suites pass. The allowlist is stricter than before, which is the intended behavior. |
| Access control / trust boundaries | Tick is server-internal. No external input crosses a trust boundary in this path. Filter change does not alter caller permissions or data supply. |
| New dependencies | None. No Cargo.toml changes. |
| Secrets / credentials | None in diff. |
| typed_graph.rs | Comment-only change. Filter line unchanged (`status != Status::Quarantined`). No behavioral change, no security impact. |

### Subquery Side (Critical Verification)

The critical correctness element — that BOTH the outer JOIN and the inner scalar subquery (aliases ea2/eb2) use the allowlist — is confirmed in source at lines 239-244 of co_access_promotion_tick.rs:

```sql
JOIN entries ea2 ON ea2.id = ca2.entry_id_a AND ea2.status = ?3   -- subquery
JOIN entries eb2 ON eb2.id = ca2.entry_id_b AND eb2.status = ?3   -- subquery
...
JOIN entries ea ON ea.id = ca.entry_id_a AND ea.status = ?3       -- outer
JOIN entries eb ON eb.id = ca.entry_id_b AND eb.status = ?3       -- outer
```

All four conditions are allowlist form. The new test `test_deprecated_endpoint_pair_not_promoted` catches a missed subquery filter: it seeds the deprecated pair with `count=10` and the active pair with `count=5`, then asserts the promoted edge weight is `1.0` (not `0.5`). A wrong subquery filter would inflate `max_count` to 10 and fail the assertion.

**No blocking findings. Approved from security perspective.**

PR #543, opened 2026-04-07
[crt-049] Knowledge Reuse: explicit read signal + total_served redefinition

### LINKED ISSUE #539: redesign Knowledge Reuse metric: log context_get calls and report explicit agent reads
## Problem

The current Knowledge Reuse block in `context_cycle_review` sources from `query_log.result_entry_ids` — entries returned by `context_search`. This is a weak proxy. A search result appearing in an agent's context doesn't mean the agent read or acted on it. The strongest signal of intentional knowledge use is `context_get`: an agent explicitly fetching a specific entry by ID. `context_get` is currently not logged anywhere.

## Goals

1. Log `context_get` calls to a new `get_log` table: `entry_id`, `session_id`, `ts`, `phase`, `found` (bool). Schema version bump.
2. Redesign the Knowledge Reuse block in `context_cycle_review` to report distinct entries explicitly retrieved via `context_get` during the cycle, attributed via `session_id → feature_cycle`.
3. Retain the existing `query_log`-sourced data as a secondary signal if warranted (separate design decision), but the headline metric becomes explicit reads.

## Why a new table, not query_log

`query_log` is search-shaped — query text, JSON array of results, similarity scores. A `context_get` has none of those. A dedicated `get_log` with one row per call follows the established single-row-per-event pattern (injection_log does the same thing). Fitting `context_get` into `query_log` via a special `source` value would require nullable columns and awkward JSON-array-of-one encoding.

## Retention

`get_log` retention follows the same `session_id → feature_cycle` chain as `query_log` and `injection_log` — no new retention logic needed.

## Open Questions (resolve in design)

1. Log failed gets (entry not found)? Probably yes — an agent looking for a missing entry is also a signal.
2. Combine `injection_log` entries with `get_log` in the metric output, or report separately? Both represent "entries the agent saw" but with different intent: injected vs. explicitly fetched.

## Non-Goals

- Changing `context_search` or `context_briefing` logging paths
- Backfilling `get_log` for historical cycles (cold-start handled by existing pattern)
- Changing `context_get` behavior or response — logging only

## Dependencies

Builds on #536 (bugfix: Knowledge Reuse label cleanup — once merged).

dug-21:
## Summary

- Adds `explicit_read_count` and `explicit_read_by_category` to `FeatureKnowledgeReuse`, derived from `context_get` / single-ID `context_lookup` `PreToolUse` observations already loaded in `context_cycle_review` — no new DB pass
- Renames `delivery_count` → `search_exposure_count` with stacked serde aliases (`"delivery_count"`, `"tier1_reuse_count"`) for full backward compat with stored `cycle_review_index` rows
- Redefines `total_served` as `|explicit_read_ids ∪ injection_ids|` — search exposures excluded (they are not "served")
- Corrects render guard to `total_served == 0 && search_exposure_count == 0` — injection-only cycles now render correctly
- Bumps `SUMMARY_SCHEMA_VERSION` 2 → 3 with advisory message naming the `total_served` semantic change
- Provides the cycle-level `explicit_read_by_category` field that ASS-040 Group 10 depends on

**Key implementation detail**: `ObservationRecord.input` arrives from the hook listener as `Value::String(raw_json)`, not `Value::Object`. Extraction uses a two-branch parse (confirmed by `extract_topic_signal` at `listener.rs:1911`). Both integer-form `{"id": 42}` and string-form `{"id": "42"}` IDs are handled.

## Test plan

- [x] 4,336 unit tests pass (0 failures)
- [x] 188 integration tests pass — smoke (23/23), lifecycle (48+5xfail), tools (117+2xfail)
- [x] All 7 gate ACs verified: AC-02 (triple-alias serde), AC-06 (normalize_tool_name), AC-13 (explicit_read_by_category contract), AC-14/AC-15 (total_served semantics), AC-16 (string-form IDs), AC-17 (injection-only render guard)
- [x] Gates 3a, 3b, 3c: all PASS

Closes #539

### DIFF
--- a/crates/unimatrix-server/src/mcp/tools.rs
+++ b/crates/unimatrix-server/src/mcp/tools.rs
@@ -2720,11 +2720,22 @@ fn check_stored_review(
     current_version: u32,
 ) -> Result<(unimatrix_observe::RetrospectiveReport, Option<String>), serde_json::Error> {
     let advisory = if record.schema_version != current_version {
+        let context = match record.schema_version {
+            2 => " (schema_version 2 predates the explicit read signal and total_served \
+                  redefinition — search exposures no longer contribute to total_served)"
+                .to_string(),
+            v if v < 2 => format!(
+                " (schema_version {} predates curation health metrics and the explicit read signal)",
+                v
+            ),
+            v => format!(
+                " (schema_version {} is newer than current version {}; downgrade not supported)",
+                v, current_version
+            ),
+        };
         Some(format!(
-            "Stored review has schema_version {} (current: {}). \
-             schema_version 2 predates the explicit read signal and total_served redefinition \
-             (search exposures no longer contribute to total_served); use force=true to recompute.",
-            record.schema_version, current_version
+            "Stored review has schema_version {} (current: {}).{} use force=true to recompute.",
+            record.schema_version, current_version, context
         ))
     } else {
         None
--- a/product/features/crt-049/agents/crt-049-agent-4-tester-report.md
+++ b/product/features/crt-049/agents/crt-049-agent-4-tester-report.md
@@ -1,0 +1,46 @@
+# Agent Report: crt-049-agent-4-tester
+
+**Phase**: Stage 3c — Test Execution
+**Feature**: crt-049 — Knowledge Reuse Metric: Explicit Read Signal
+
+## Summary
+
+All tests pass. All 7 gate ACs verified. No GH Issues filed.
+
+## Unit Test Results
+
+- Workspace total: 4,336 passed / 0 failed
+- crt-049 new tests: ~43 (types.rs:8, knowledge_reuse.rs:~24, retrospective.rs:6, tools.rs:4, cycle_review_index.rs:1 updated)
+
+## Integration Test Results
+
+| Suite | Passed | Failed | XFail | XPass |
+|-------|--------|--------|-------|-------|
+| smoke | 23 | 0 | 0 | 0 |
+| lifecycle | 48 | 0 | 5 (pre-existing) | 2 (pre-existing) |
+| tools | 117 | 0 | 2 (pre-existing) | 0 |
+
+## Gate AC Status
+
+| Gate AC | Result |
+|---------|--------|
+| AC-02 (triple-alias serde chain) | PASS |
+| AC-06 (normalize_tool_name prefix) | PASS |
+| AC-13 (explicit_read_by_category contract) | PASS |
+| AC-14 (total_served excludes search exposures) | PASS |
+| AC-15 (total_served deduplication) | PASS |
+| AC-16 (string-form ID handling) | PASS |
+| AC-17 (injection-only render guard) | PASS |
+
+## Risk Coverage
+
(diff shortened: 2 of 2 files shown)


dug-21:
## Security Review: crt-049-security-reviewer

**Risk Level: LOW**

All security-relevant properties of this change are sound. One non-blocking documentation accuracy issue identified.

---

### Finding 1: Advisory message hardcodes "schema_version 2" for all version mismatches

**Severity**: Low | **Location**: `crates/unimatrix-server/src/mcp/tools.rs` `check_stored_review` ~line 2724 | **Blocking**: No

The advisory message always says "schema_version 2 predates the explicit read signal..." regardless of the stored record's actual version. A version=1 record (pre-crt-047) will receive this text, which is factually wrong — version 1 predates curation health columns, not the explicit read signal.

This is a display-accuracy issue only. The record is still returned correctly and `record.schema_version` appears in the output so callers can see the actual value. Not blocking.

---

### OWASP Assessment

| Check | Result |
|-------|--------|
| SQL injection | CLEAR — `build_batch_meta_query` uses `?` placeholders; IDs are `u64` bound values, no string interpolation |
| Deserialization | CLEAR — `extract_explicit_read_ids` parses input via `serde_json::from_str(...).ok()` with no panic path; malformed JSON silently skipped |
| Input validation | CLEAR — `id` extracted only via `as_u64()` or `str.parse::<u64>()`, both reject negatives, floats, and non-numeric strings |
| Path traversal | N/A — no file path operations in changed code |
| Access control | CLEAR — no new trust boundaries; `extract_explicit_read_ids` processes data already loaded by the existing step 12 attribution load |
| Hardcoded secrets | CLEAR — no credentials, tokens, or API keys in diff |
| Unsafe code | CLEAR — no `unsafe` blocks in any changed file |
| New dependencies | CLEAR — no new crates added (Cargo.toml unchanged) |

---

### Blast Radius

Worst case if `extract_explicit_read_ids` has a subtle bug: `explicit_read_count` returns 0 for all cycles (silent undercount). No data corruption, no crash, no write path affected. The only consequence is a metric being wrong in the cycle review report.

Overcount is structurally prevented: `obj["id"]` on an absent key returns `Value::Null`, which fails `as_u64()`, so filter-path `context_lookup` calls (no `id` field) cannot be incorrectly included.

The cardinality cap at 500 IDs is applied only to the metadata lookup input, not to `explicit_read_ids.len()`. `explicit_read_count` is always the true distinct count. Correct per ADR-004.

---

### Regression Risk

The `total_served` semantic change (search exposures removed) is the highest regression risk. Mitigations are all present:

1. `SUMMARY_SCHEMA_VERSION` bumped 2 → 3 — stale stored records trigger the advisory path, not silently return with new semantics.
2. `#[serde(default)]` on new fields — pre-crt-049 stored JSON deserializes without error.
3. `#[serde(alias = "delivery_count")]` on `search_exposure_count` — old stored rows with the old key name deserialize correctly.
4. AC-02 GATE tests verify all three alias paths and canonical serialization key.

The two sequential `batch_entry_meta_lookup` awaits use separate `await` points with no connection held across both — no deadlock risk (integration risk I-02 from RISK-TEST-STRATEGY).

---

### Summary

- 1 finding (low severity, non-blocking): advisory message text accuracy for version mismatch < 2
- 0 blocking findings
- No secrets, no injection surfaces, no unsafe code, no new dependencies

PR #571, opened 2026-04-21
fix: guard rework candidate normalization with always-on assert_ne! (#565)

### LINKED ISSUE #565: bug: guard against post_tool_use_rework_candidate leaking to observations.hook
## Summary

`post_tool_use_rework_candidate` is an internal event type that `extract_observation_fields()` normalizes to `"PostToolUse"` before writing to `observations.hook`. However, the raw string appears in test code, and there is no guard at the ingest boundary to prevent it from reaching the DB unhidden.

If a code bug bypassed the normalization step, rows with `hook = 'post_tool_use_rework_candidate'` would be written to the DB and silently invisible to all downstream queries — no error, no log, just missing data.

## Fix

Add a guard assertion (or explicit error return) in `extract_observation_fields()` after normalization to ensure the value written to `observations.hook` is never `"post_tool_use_rework_candidate"`. For example:

```rust
debug_assert_ne!(hook, "post_tool_use_rework_candidate",
    "internal event type must be normalized before DB write");
```

Or an explicit `Err` path if normalization was supposed to have occurred and did not.

## Source

Discovered during ASS-051 blast radius audit (`product/research/ass-051/FINDINGS.md`, Out-of-Scope Discovery #2).

dug-21:
## Summary

- Removes the `debug_assert` at `extract_observation_fields()` line 2695 that fired on the valid `RecordEvents` batch dispatch path, silently gating col-019 normalization tests under `#[cfg(not(debug_assertions))]` and making them invisible in CI
- Replaces it with an always-on `assert_ne!` post-condition after the normalization rewrite — fires in both debug and release; message includes the actual hook value for immediate diagnostic context on panic
- Ungates 4 col-019 tests (they now run in all build modes)
- Converts the now-contradictory `#[should_panic]` `#[cfg(debug_assertions)]` test to a positive assertion
- Adds `test_record_events_rework_candidate_written_as_post_tool_use` — end-to-end batch-path test through the DB write

## Test plan

- [ ] `test_rework_candidate_normalizes_to_post_tool_use` — positive assertion, ungated
- [ ] `test_record_events_rework_candidate_written_as_post_tool_use` — batch path → DB write → `hook = 'PostToolUse'`
- [ ] All 4 previously cfg-gated col-019 tests now pass in debug builds
- [ ] Full workspace: `cargo test --workspace` — all pass
- [ ] Clippy: `cargo clippy --workspace -- -D warnings` — no new warnings in changed file
- [ ] Integration smoke: 23/23 pass

Closes #565

### DIFF
--- a/crates/unimatrix-server/src/uds/listener.rs
+++ b/crates/unimatrix-server/src/uds/listener.rs
@@ -8169,17 +8169,22 @@ mod tests {
             "expected HookResponse::Ack, got {response:?}"
         );
 
-        // Allow the fire-and-forget spawn_blocking write to complete.
-        for _ in 0..20 {
-            tokio::task::yield_now().await;
-        }
-        std::thread::sleep(std::time::Duration::from_millis(50));
-
-        let hook: String =
-            sqlx::query_scalar("SELECT hook FROM observations WHERE session_id = 'sess-gh565'")
-                .fetch_one(store.read_pool_test())
-                .await
-                .expect("observation row must exist after RecordEvents dispatch");
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(2);
+        let hook = loop {
+            if let Ok(h) = sqlx::query_scalar::<_, String>(
+                "SELECT hook FROM observations WHERE session_id = 'sess-gh565'",
+            )
+            .fetch_one(store.read_pool_test())
+            .await
+            {
+                break h;
+            }
+            assert!(
+                tokio::time::Instant::now() < deadline,
+                "timed out waiting for spawn_blocking DB write to complete"
+            );
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+        };
 
         assert_eq!(
             hook, "PostToolUse",
--- a/crates/unimatrix-server/src/uds/listener.rs
+++ b/crates/unimatrix-server/src/uds/listener.rs
@@ -2686,18 +2686,6 @@ fn extract_observation_fields(event: &unimatrix_engine::wire::ImplantEvent) -> O
     let ts_millis = (event.timestamp as i64).saturating_mul(1000);
     let hook = event.event_type.clone();
 
-    // Normalization contract enforcement (AC-16, ADR-005, FR-08.1).
-    // "post_tool_use_rework_candidate" is an internal routing label that must be
-    // converted to "PostToolUse" in build_request() before reaching this function.
-    // If this assert fires in debug builds, normalization failed or was bypassed.
-    // Scoped to rework candidate only — PostToolUseFailure is intentionally preserved
-    // (ADR-003 col-027). Compiled out in release builds.
-    debug_assert!(
-        hook != "post_tool_use_rework_candidate",
-        "rework candidate string escaped normalization boundary and reached extract_observation_fields: {}",
-        hook
-    );
-
     let (tool, input, response_size, response_snippet) = match hook.as_str() {
         "PreToolUse" => {
             let tool = event
@@ -2770,6 +2758,11 @@ fn extract_observation_fields(event: &unimatrix_engine::wire::ImplantEvent) -> O
     } else {
         hook
     };
+    assert_ne!(
+        hook, "post_tool_use_rework_candidate",
+        "BUG: normalization failed, hook={}",
+        hook
+    );
 
     ObservationRow {
         session_id,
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — uni-security-reviewer (agent 565-security-reviewer)

**Risk level: LOW**
**Blocking findings: NO**

---

### Change Summary (read cold from diff)

Removes a `debug_assert!` in `extract_observation_fields()` that was compiled out in release builds, replaces it with an always-on `assert_ne!` post-condition placed after the normalization rewrite, and ungates four col-019 tests plus converts one `#[should_panic]` test to a positive assertion. Adds one new end-to-end batch-path integration test.

---

### OWASP / Security Findings

**Input Validation — PASS**

The function receives an `ImplantEvent` whose `session_id` is validated by `sanitize_session_id()` at all three call sites (lines 669, 745, 877) before `extract_observation_fields()` is ever reached. The changed code touches only the `hook` (event_type) field — an internal string value, not used in SQL interpolation. No new external inputs are introduced. No validation is removed.

**Injection — PASS**

The `hook` string written to the DB is the output of the normalization expression, not raw user input. The match arm and the post-normalization if-expression both act as a whitelist; the assert enforces the rewrite happened. No SQL, shell, or path operations are introduced.

**Deserialization — PASS**

No new deserialization paths. Payload parsing (serde_json) is unchanged. The `payload` fields read in the match arm were already present before this diff.

**Access Control — PASS**

`uds_has_capability(Capability::SessionWrite)` gates all three dispatch arms that call `extract_observation_fields()`. The change does not add or remove capability checks.

**Secrets — PASS**

No hardcoded credentials, tokens, keys, or environment variable references in the diff.

**Dependencies — PASS**

No new crate dependencies.

---

### Assert Placement Analysis

The assert_ne! fires **after** the normalization rewrite (lines 2756–2760 normalize, lines 2761–2765 assert). The logic is:

```
hook = if hook == post_tool_use_rework_candidate { PostToolUse } else { hook };
assert_ne!(hook, post_tool_use_rework_candidate, ...);
```

This is a tautological post-condition. After the if-expression, `hook` is by construction never equal to `post_tool_use_rework_candidate` — the assert cannot fire on any code path that reaches it through the existing match arms. It acts as documentation of the invariant enforced by the compiler, not a runtime guard against untrusted input.

The assert_ne! **cannot fire on a valid hook value** because:
1. The normalization rewrite is unconditional and executes on every call path.
2. The assert checks the post-rewrite value, which is either `PostToolUse` or unchanged (any other string).
3. No other string equals `post_tool_use_rework_candidate` after that rewrite.

There is no scenario in which the assert_ne! fires in production that wouldn't also have been a logic bug requiring investigation — which is the correct behavior for a broken invariant guard.

---

### Blast Radius

If a future code change breaks the normalization (e.g., moves the assert above the rewrite, or the rewrite is removed in a refactor), the assert_ne! will panic in both debug and release. The panic occurs inside `extract_observation_fields()`, which is called synchronously before spawn_blocking. The panic would propagate into the dispatch_request call stack and cause the UDS connection handler to unwind for that event. Dependent sessions would receive no Ack for that event; the MCP connection would drop. No data corruption or privilege escalation is possible — the failure mode is a clean unwinding panic with a diagnostic message including the offending hook value.

Worst case: a rogue hook name matching `post_tool_use_rework_candidate` from a future code path (e.g., a new LLM provider normalization table mapping an event to this string) would trigger a panic. The assert should remain as-is; any new normalization that produces this string should be treated as a bug by definition.

---

### Regression Risk

**CI test ungating — LOW RISK**

The four previously `#[cfg(not(debug_assertions))]`-gated tests now run in all build modes. In debug mode these tests call `extract_observation_fields()` with `event_type = "post_tool_use_rework_candidate"`. The new assert_ne! post-condition will pass because the normalization rewrite runs first. No test breakage is expected; this is the explicit intent of the fix.

The converted test (`test_rework_candidate_normalizes_to_post_tool_use`) was previously `#[should_panic]` — it will now panic if normalization fails, which is the correct failure mode.

**New batch-path test** — uses a 50ms `thread::sleep` to let spawn_blocking complete before querying the DB. This is a timing dependency; under high load it could be flaky. Not a security concern but worth noting for CI stability.

---

### Minimal Change Verification

The diff is confined to a single file (`crates/unimatrix-server/src/uds/listener.rs`). All changed lines are either the removed debug_assert, the replacement assert_ne!, test cfg-gate removal, test conversion, or the new end-to-end test. No unrelated changes detected. The change is appropriately minimal.

PR #534, opened 2026-04-06
[crt-047] Curation health metrics — per-cycle correction rate, orphan deprecation ratio, σ-based anomaly detection

dug-21:
## Summary

- Adds `CurationSnapshot` computation to `context_cycle_review`: per-cycle corrections (agent vs. human source), orphan deprecation count, σ comparison against rolling 10-cycle baseline with history-length annotation
- Adds Phase 7c curation health aggregate to `context_status`: correction rate mean/stddev, agent%/human% breakdown, orphan ratio, trend direction (Increasing/Decreasing/Stable at ≥6 cycles)
- Schema v23→v24: 7 new columns on `cycle_review_index` (`corrections_total`, `corrections_agent`, `corrections_human`, `corrections_system`, `deprecations_total`, `orphan_deprecations`, `first_computed_at`)
- New `services/curation_health.rs` module with all types and pure compute functions
- Cold-start handled: σ at ≥3 cycles, trend at ≥6 cycles; raw counts only below threshold

## Key design decisions

- **ENTRIES-only orphan attribution** — no AUDIT_LOG join needed; only `context_deprecate` produces `superseded_by IS NULL` and always sets `updated_at` (ADR-003)
- **`first_computed_at` ordering key** — stable across `force=true` reruns; plain `INSERT OR REPLACE` replaced with two-step upsert to preserve this value (ADR-001)
- **`corrections_total` excludes system writes** — only agent + human corrections signal intentional curation behavior (ADR-002)

## Test plan

- [x] 4621 unit tests pass (0 failures)
- [x] Integration smoke gate: 23/23 PASS
- [x] Lifecycle suite: 47 passed, 5 xfailed (pre-existing)
- [x] Edge cases suite: 23 passed, 1 xfailed (pre-existing GH#111)
- [x] 4 new integration tests added
- [x] All 27 acceptance criteria verified (AC-01–AC-18, AC-R01–AC-R05)
- [x] Gate 3a PASS, Gate 3b PASS, Gate 3c PASS

Closes #529

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #529, opened 2026-04-06
feat: curation health metrics — per-cycle correction rate, orphan deprecation ratio, σ-based anomaly detection


dug-21:
## Summary

Add a curation health metrics layer to `context_cycle_review` and `context_status`. Where Lambda measures structural integrity of the knowledge graph, this feature measures the health of the *curation process itself* — are agents and humans engaging critically with the knowledge base, or is drift accumulating unchecked?

This is a corpus-level behavioral signal, not an entry-level lifecycle mechanism. It answers: \"is the correction process working?\" — not \"which entries should be removed?\" (that is #363 / #370).

---

## Background

`context_correct` is the designed mechanism for knowledge curation. When agents encounter inaccurate entries, they correct them — maintaining a hash-chained correction history. When the system is working, corrections flow naturally from usage. When they don't, one of two things is true: the knowledge base is unusually stable, or agents are encountering errors and not correcting them.

These look identical in the data. But the *rate* and *source distribution* of corrections, measured over cycles and compared against the corpus's own baseline, surfaces the difference as a signal — without requiring domain-specific thresholds.

Two key observations that shaped this design:

**Corrections have two populations with different implications:**
- In-flow agent corrections: agent finds something wrong during a cycle, calls `context_correct` immediately. Healthy critical engagement.
- Post-hoc human corrections: human notices a failure pattern after the fact, intervenes. Reactive curation — something slipped through without being caught in-flow.

These are distinguishable via `trust_source`. A corpus where all corrections are human post-hoc and none are agent in-flow indicates agents aren't engaging critically, even if the aggregate correction rate looks healthy.

**Orphan deprecations are a discipline signal:**
`context_correct` atomically deprecates + creates a successor. `context_deprecate` alone does not. Entries with `status = Deprecated AND superseded_by IS NULL` represent knowledge that was removed without replacement. There are valid cases, but a consistently high orphan ratio indicates agents are deleting instead of correcting — breaking the correction chain discipline.

**σ-based comparison is self-calibrating:**
[...]

PR #572, opened 2026-04-21
fix: route curation snapshot queries through read_pool via store layer (#535)

### LINKED ISSUE #535: fix: compute_curation_snapshot() uses write_pool_server() for read-only ENTRIES queries
## Background

Introduced in crt-047 (#529, PR #534). `compute_curation_snapshot()` in `crates/unimatrix-server/src/services/curation_health.rs` runs read-only SQL queries against ENTRIES (corrections and deprecation counts) but uses `write_pool_server()` instead of `read_pool()`.

## Root cause

`SqlxStore::read_pool()` is `pub(crate)` to `unimatrix-store`. Server-side code in `unimatrix-server` cannot call it directly. The agent fell back to `write_pool_server()` as the established workaround (Unimatrix entry #3028), but this is incorrect for read workloads.

## Impact

- `write_pool_server()` is a single-connection serializer. Every call to `context_cycle_review` that computes a curation snapshot holds the write connection for the duration of the ENTRIES reads, blocking concurrent write operations.
- Under normal single-agent usage this is low impact. Under concurrent review calls (e.g. multiple agents calling `context_cycle_review` simultaneously) this serializes unnecessarily.

## Fix

Expose `read_pool()` with sufficient visibility for server-side reads, or introduce a dedicated `store.read_entries_for_curation_snapshot()` method on the store trait that runs on the read pool and returns the `CurationSnapshot` directly — keeping pool selection inside the store crate where `read_pool()` is accessible.

The store-method approach is preferred as it maintains the pub(crate) encapsulation and is consistent with how other read-heavy operations are exposed via the trait.

## Scope

- `crates/unimatrix-store/src/cycle_review_index.rs` — add `get_curation_snapshot(feature_cycle, cycle_start_ts, review_ts)` to the store trait
- `crates/unimatrix-server/src/services/curation_health.rs` — `compute_curation_snapshot()` calls the store method instead of issuing SQL directly
- No schema changes, no migration required

## Related

- crt-047 (#529) — introduced the issue
- Unimatrix entry #3028 — documents the `write_pool_server()` workaround pattern (this fix supersedes it for the curation snapshot case)

dug-21:
## Summary

- `compute_curation_snapshot()` was using `write_pool_server()` (single-connection serializer) for three read-only ENTRIES SELECT queries, blocking concurrent write operations during every `context_cycle_review` call
- Added `CurationSnapshotRow` struct + `SqlxStore::get_curation_snapshot()` method to `unimatrix-store/src/cycle_review_index.rs` using `self.read_pool()` — keeping pool selection inside the store crate where `read_pool()` is accessible (`pub(crate)`)
- Rewrote `compute_curation_snapshot()` as a thin delegating wrapper; removed unused `sqlx::Row` import; updated stale doc comments referencing the old workaround

Fixes #535. Introduced in crt-047 (#529, PR #534). Supersedes Unimatrix entry #3028 (write_pool_server workaround pattern — now entry #4342).

## Test plan

- [x] 3 new unit tests: `test_get_curation_snapshot_counts_corrections_by_trust_source`, `test_get_curation_snapshot_counts_deprecations_and_orphans_in_window`, `test_get_curation_snapshot_empty_store_returns_zeros`
- [x] 2915/2915 workspace unit tests pass
- [x] 7/7 targeted curation/cycle_review integration tests pass
- [x] Clippy clean on changed crates
- [x] No schema changes, no migration

### DIFF
--- a/crates/unimatrix-server/src/services/curation_health.rs
+++ b/crates/unimatrix-server/src/services/curation_health.rs
@@ -14,10 +14,9 @@
 //! - SQL uses parameterized binds only — no string interpolation of user input (SEC-01).
 //! - Population stddev (divide by n, not n-1) — matches `unimatrix_observe::baseline`.
 //! - Zero-stddev and zero-denominator guards prevent NaN propagation (NFR-02).
-//! - `write_pool_server()` is used because `read_pool()` is `pub(crate)` in
-//!   `unimatrix-store` and not accessible cross-crate (entry #3028).
+//! - `compute_curation_snapshot()` delegates to `store.get_curation_snapshot()` which
+//!   uses `read_pool()` — correct pool selection for a read-only workload (fixes GH #535).
 
-use sqlx::Row;
 use unimatrix_core::CoreError;
 use unimatrix_observe::{
     CurationBaselineComparison, CurationHealthSummary, CurationSnapshot, TrendDirection,
@@ -112,9 +111,9 @@ pub struct CurationBaseline {
 
 /// Query ENTRIES to compute curation counts for a single feature cycle.
 ///
-/// Three SQL queries against ENTRIES (no AUDIT_LOG join — ADR-003).
-/// Uses `write_pool_server()` because `read_pool()` is `pub(crate)` in
-/// `unimatrix-store` and not accessible from `unimatrix-server` (entry #3028).
+/// Delegates to `store.get_curation_snapshot()` which issues three read-only
+/// SELECT queries via `read_pool()` — correct pool selection for a read-only
+/// workload (GH #535).
 ///
 /// This function MUST be called before `store_cycle_review()` acquires the write
 /// connection, so the read completes before the write begins.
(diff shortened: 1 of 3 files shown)


dug-21 (commented):
## Security Review — 535-security-reviewer

**Risk Level: LOW**

No blocking findings. The change is minimal, targeted, and correct from both a security and architecture standpoint.

---

### What Was Changed

`compute_curation_snapshot()` in `curation_health.rs` previously issued three read-only SELECT queries directly against ENTRIES using `write_pool_server()`. The fix moves those queries into a new store method `get_curation_snapshot()` on `SqlxStore` in `cycle_review_index.rs`, which correctly uses the `pub(crate)` `read_pool()`. The server layer becomes a thin delegating wrapper.

---

### OWASP / Security Findings

**SQL Injection (OWASP A03)**
No risk. All three queries use parameterized binds only (`?1`, `?2`). No string interpolation of any external input. `feature_cycle`, `cycle_start_ts`, and `review_ts` are all bound as typed values. The module-level comment "SQL uses parameterized binds only — no string interpolation of user input (SEC-01)" is accurate and preserved.

**Access Control / Pool Segregation**
The fix improves this. `write_pool_server()` is documented as intended for server-internal writes against tables not covered by the high-level API. Using it for read-only queries against ENTRIES was a misuse that held the single write connection during read work, creating unnecessary contention. The new path goes through `read_pool()` (up to 8 concurrent connections, 2s acquire timeout vs 5s), which is the correct lane for these queries. Trust boundary is maintained: `read_pool()` remains `pub(crate)` — access from `unimatrix-server` is correctly mediated through a store method, not a raw pool reference.

**Input Validation**
No new external inputs introduced. `feature_cycle` is a caller-controlled string already used as a parameterized bind. `cycle_start_ts` and `review_ts` are i64 timestamps, also bound. No parsing or deserialization of untrusted data.

**Deserialization**
`CurationSnapshotRow` is a plain Rust struct with five i64 fields. No serde, no JSON deserialization, no external data parsing. No risk.

**Secrets / Credentials**
None present in the diff. No hardcoded values of any security concern.

**New Dependencies**
None introduced.

---

### Blast Radius Assessment

Worst case if the fix has a subtle bug: `get_curation_snapshot()` returns wrong counts (e.g., off-by-one in window bounds, wrong feature_cycle scoping). The caller (`context_cycle_review` step 8a) treats errors as non-fatal — it logs a warning and omits `curation_health` from the response. Silent wrong counts would be the failure mode, not a panic or crash. This was already the pre-fix risk profile; the fix does not change error handling semantics. The `max(0) as u32` clamp in the server layer is preserved and correct.

The `write_pool_server()` remains available for the test setup code that uses it to `UPDATE entries SET supersedes` and `UPDATE entries SET superseded_by` — appropriate since those are direct writes in test context.

---

### Regression Risk

**Low.** The SQL text is identical to what was previously in `curation_health.rs` — it was moved verbatim, not rewritten. The structural change (3 inline queries → 1 store method returning a struct) is purely mechanical. Existing test coverage in `curation_health.rs` (behind the `test-support` feature flag) exercises the full compute path end-to-end, and three new integration tests in `cycle_review_index.rs` exercise the store method directly:

- `test_get_curation_snapshot_counts_corrections_by_trust_source` — verifies feature_cycle scoping, trust_source bucketing, supersedes filtering
- `test_get_curation_snapshot_counts_deprecations_and_orphans_in_window` — verifies window bounds for deprecations_total and orphan_deprecations
- `test_get_curation_snapshot_empty_store_returns_zeros` — verifies the empty-table case returns zeros, not errors

One minor observation: the new store-layer tests call `sqlx::query(...).execute(store.write_pool_server())` for test setup (backdating `updated_at`, setting `supersedes`/`superseded_by`). This is correct per the established pattern for intra-crate test setup (Unimatrix entry #2744) — `write_pool_server()` is appropriate for direct writes in tests within the store crate.

---

### Summary

The change is minimal, confined to two files plus `lib.rs` re-export, and moves existing SQL verbatim. No security concerns identified. Pool selection is now correct. Tests adequately cover the new store method. Approved for merge.

PR #573, opened 2026-04-21
fix: byte-based content size enforcement for context_store (#561)

### LINKED ISSUE #561: feat: byte-based content size enforcement — context_store cap, context_status format:json risk
## Summary

The original issue framed this as enforcing \`max_tokens\` on \`context_briefing\`. Analysis during Wave 2 planning showed that framing is wrong on two counts:

1. **\`context_briefing\` is not the risk surface.** Maximum possible response is ~4.4 KB (20 entries × ~209 chars/entry + overhead). The 10 KiB Codex truncation threshold is not realistically reachable under normal usage.

2. **Tokens is the wrong metric.** Bytes are deterministic, model-agnostic, and enforceable without tokenizer dependencies. Unimatrix should not manage token counts.

## Correct framing

**Cap at store time, not read time.** \`context_store\` enforces a max byte limit on entry content. \`context_get\` needs no separate cap — it cannot return more than what was stored.

## Risk surface by tool

| Tool | Risk | Disposition |
|------|------|-------------|
| \`context_store\` | Oversized entry stored | **Fix — byte cap at write time** |
| \`context_get\` | Bounded by store cap | No action needed |
| \`context_briefing\` | ~4.4 KB max (SNIPPET_CHARS=150, 20 entries) | No action needed |
| \`context_search\` | k=5 normal usage | Accept risk |
| \`context_status\` default | Fine | No action needed |
| \`context_status format:json\` | Corpus-size dependent, could be large | Document as known risk |

### context_store byte cap
- Add `max_content_bytes` to `config.toml` under `[store]` (default: 8000 bytes — sufficient for a comprehensive ADR or pattern entry, leaves headroom below the 10 KiB Codex ceiling)
- Enforce at write time in `context_store` handler: return a clear error with the configured limit and received size
- Error message format: `"Content exceeds configured maximum of {N} bytes (received {M} bytes). Consider breaking this into multiple focused entries."`
- **Do not put the specific byte value in tool documentation** — state that a limit exists and is operator-configured; the actual value is revealed only in the error response at runtime

### context_store tool description update
- Add guidance that entries should be focused and specific (single decision, pattern, or lesson)
- State that a content size limit is enforced and returns an error if exceeded
- Do not publish the configured value — keeps documentation correct across operator configurations and avoids creating a specific target for adversarial content crafting

### context_status format:json
- Document in tool description that `format: json` responses are corpus-size dependent and may be large
- No enforcement — accept risk for this low-frequency, admin-facing call

## Files
- `crates/unimatrix-server/src/infra/config.rs` — add `max_content_bytes` field
- `crates/unimatrix-server/src/services/store_ops.rs` — enforce cap at write time
- `crates/unimatrix-server/src/mcp/tools.rs` — update `context_store` description; update `context_status` description for `format:json` risk
- `config.toml` (default/example) — document new field
- Drop the original `max_tokens` enforcement scope entirely — parameter remains reserved/ignored

dug-21:
## Summary

- Add `StoreConfig` with `max_content_bytes` (default 8,000 bytes, configurable range [1,000, 1 MiB]) to `UnimatrixConfig`
- Enforce byte cap at write time in `validate_store_params` and `validate_correct_params` — O(1) `str::len()` check fires before the O(n) `chars().count()` scan so oversized content exits early
- `StoreConfig::validate()` enforces bounds at startup; `merge_configs()` arm added (mandatory 5th site per established pattern)
- `context_store` description updated: focused-entry guidance, size limit exists without publishing the value
- `context_status` description updated: `format:json` response size scales with corpus size (no hard cap implied)
- `config.toml` documents new `[store]` section

## Test plan

- [x] `test_validate_store_params_content_byte_limit` — over-limit ASCII and multibyte emoji (🟩×2001 = 8,004 bytes, 2,001 chars — critical case: passes char limit, fails byte limit)
- [x] `test_validate_store_params_byte_check_fires_before_char_check` — byte error fires before char error
- [x] `test_validate_correct_params_content_byte_limit` — same for correct path
- [x] `test_validate_correct_params_byte_check_fires_before_char_check`
- [x] `StoreConfig::validate()` bounds tests (below min, above max, at boundaries)
- [x] 2926/2926 workspace unit tests passing
- [x] 23/23 integration smoke tests passing
- [x] 14/14 context_store integration suite passing
- [x] 10/10 input validation integration suite passing
- [x] No new clippy warnings

Closes #561

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -120,6 +120,19 @@ audit_log_retention_days = 180
 max_cycles_per_tick = 10
 
 # =============================================================================
+# [store] — Write-path limits for knowledge entry storage
+# =============================================================================
+
+[store]
+# Maximum byte length for content on context_store and context_correct calls.
+# The limit is byte-based (UTF-8), not character-based — multibyte characters
+# (e.g., emoji, CJK) count toward this limit by their byte width, not character count.
+# Entries should be focused and specific; large content should be split into
+# multiple targeted entries.
+# Range: [1000, 1048576]. Default: 8000
+# max_content_bytes = 8000
+
+# =============================================================================
 # [observation] — Domain pack registration
 # =============================================================================
 
--- a/crates/unimatrix-server/src/infra/config.rs
+++ b/crates/unimatrix-server/src/infra/config.rs
@@ -80,6 +80,8 @@ pub struct UnimatrixConfig {
     pub observation: ObservationConfig,
     #[serde(default)]
     pub retention: RetentionConfig,
+    #[serde(default)]
+    pub store: StoreConfig,
     // CycleConfig is intentionally absent (ADR-004: stub removed, rename is hardcoded).
 }
 
@@ -1570,6 +1572,60 @@ impl RetentionConfig {
 }
 
 // ---------------------------------------------------------------------------
+// StoreConfig (#561)
+// ---------------------------------------------------------------------------
+
+/// `[store]` section — write-path limits for knowledge entry storage.
+///
+/// All fields have compiled defaults via `#[serde(default)]` so an absent
+/// `[store]` block in config.toml applies defaults without error.
+#[derive(Debug, Clone, PartialEq, serde::Deserialize)]
+#[serde(default)]
+pub struct StoreConfig {
+    /// Maximum byte length for `content` on context_store and context_correct calls.
+    ///
+    /// Enforced as a byte check (O(1)) before the character count check (O(n)).
+    /// Applies to both the MCP write path and the UDS path via validation.rs.
+    ///
+    /// Range: [1_000, 1_048_576]. Default: 8_000.
+    pub max_content_bytes: usize,
+}
+
+fn default_max_content_bytes() -> usize {
+    8_000
+}
+
+impl Default for StoreConfig {
+    fn default() -> Self {
+        StoreConfig {
+            max_content_bytes: default_max_content_bytes(),
+        }
+    }
+}
+
+impl StoreConfig {
+    /// Validate all StoreConfig fields against their documented ranges.
+    ///
+    /// Called during server startup alongside InferenceConfig::validate()
+    /// and RetentionConfig::validate(). An out-of-range value aborts startup
+    /// with a structured error naming the field.
+    ///
+    /// Checks:
+    ///   - max_content_bytes in [1_000, 1_048_576]
+    pub fn validate(&self, path: &Path) -> Result<(), ConfigError> {
+        if self.max_content_bytes < 1_000 || self.max_content_bytes > 1_048_576 {
+            return Err(ConfigError::StoreFieldOutOfRange {
+                path: path.to_path_buf(),
+                field: "max_content_bytes",
+                value: self.max_content_bytes.to_string(),
+                reason: "must be in range [1000, 1048576]",
+            });
+        }
+        Ok(())
+    }
+}
+
+// ---------------------------------------------------------------------------
 // Preset enum
 // ---------------------------------------------------------------------------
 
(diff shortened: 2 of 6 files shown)


dug-21 (commented):
## Security Review — bugfix/561-byte-content-cap

**Risk Level: LOW**

Reviewed diff cold: `config.toml`, `infra/config.rs`, `infra/validation.rs`, `mcp/tools.rs`, `server.rs`, `main.rs`.

---

### OWASP / Input Validation

**Finding 1 — Information disclosure in error message (Low, Non-blocking)**

The rejection error message exposes the operator-configured limit to any caller with Write capability:

> "Content exceeds configured maximum of 8000 bytes (received 8001 bytes). Consider breaking this into multiple focused entries."

The byte count received (`byte_len`) is caller-controlled input echoed back verbatim. The configured limit (`max_content_bytes`) is a server-side constant. Neither value is sensitive — Write capability is already granted before validation runs. No injection vector exists: the values are formatted as integers from `usize` arithmetic, not from untrusted string input. The disclosure is intentional and useful for diagnostics.

**Verdict**: No security concern. The echo is safe (numeric only, no untrusted string interpolation).

---

**Finding 2 — Merge-config comparison against default may not pick up intentional config identity (Low, Non-blocking)**

`merge_configs()` uses `project.store.max_content_bytes != default.store.max_content_bytes` to decide which config wins. This is the established pattern (identical logic for `retention` and `inference` sections). The edge case: an operator who explicitly sets `max_content_bytes = 8000` (the default) in a project config gets the global value instead of the project value. This is not a security risk — the values are identical — but it is a documented behavioral quirk consistent with how all other config sections work in this codebase. No deviation from the established pattern.

**Verdict**: Consistent with existing design. No security concern.

---

**Finding 3 — Operator misconfiguration: minimum floor of 1,000 bytes prevents silent DoS (Low, Non-blocking)**

`StoreConfig::validate()` enforces `[1_000, 1_048_576]` at startup. A misconfigured value aborts startup with a structured error — no silent truncation, no runtime panic. The minimum of 1,000 bytes is meaningful (prevents an operator from accidentally setting `max_content_bytes = 1` and rendering the tool unusable), and the startup abort ensures the error is visible immediately.

**Verdict**: Well-handled. No risk.

---

### Blast Radius Assessment

If the byte check has a subtle off-by-one bug, the worst-case failure modes are:

- **Too strict (rejects valid content)**: Callers with exactly `max_content_bytes` bytes of content get a spurious error. Observable immediately. No data corruption, no information disclosure.
- **Too permissive (accepts 1 extra byte)**: The existing `MAX_CONTENT_LEN = 50_000` char check remains as backstop. Storage proceeds normally.

The fix does not affect reads, authentication, access control, or any persistent state. Failure mode is safe (error returned, not silent data corruption).

The test `test_validate_store_params_content_byte_limit` explicitly checks the exact boundary at both `byte_len == max` (passes) and `byte_len == max + 1` (fails), including the critical multibyte case (🟩×2000 = 8000 bytes passes; 🟩×2001 = 8004 bytes fails). Off-by-one is covered.

---

### Regression Risk

**Existing tests updated correctly**: All prior callers of `validate_store_params` and `validate_correct_params` in tests now pass `50_000` (above the production default of 8,000) so legacy tests remain valid under the new signature. No existing behavior changes unless an operator explicitly sets `max_content_bytes` below the size of content they were previously storing.

**Breaking change for existing operators**: An operator running the server against content they've previously stored that was, say, 12,000 bytes in size will now be blocked from re-storing identical content if they haven't customized the config. This is the intended behavior (the bug was the missing cap), but operators migrating from pre-fix versions who have existing entries over 8,000 bytes cannot use `context_correct` on those entries without either raising the limit or splitting the content. This is a UX consideration, not a security issue. The config.toml documentation explicitly states the range and default, so operators can raise the limit if needed.

---

### Dependency Safety

No new dependencies introduced. The fix is pure Rust stdlib (`str::len()`, `format!`, `serde::Deserialize`). No new deserialization of untrusted data beyond what already existed.

---

### Secrets Check

No hardcoded credentials, tokens, API keys, or secrets in the diff.

---

### Call Coverage Verification

`validate_store_params` and `validate_correct_params` are called in exactly two places in production code (both in `mcp/tools.rs`, lines 581 and 797). Both sites pass `self.store_config.max_content_bytes`. The UDS path mentioned in the PR description comment ("Applies to both the MCP write path and the UDS path via validation.rs") was verified — validation.rs is the single source of truth and both tool handlers call it.

---

### Summary

The change is minimal, correct, and well-tested. The validation is placed at the right boundary (before the expensive char scan, after capability check, before persistence). No blocking security findings. No injection, path traversal, access control, or deserialization risks introduced.

**Blocking findings: No**

Issue #580, opened 2026-04-23
[infra-001] test_quarantine_requires_write: context_quarantine still requires Admin, not Write


dug-21:
## Discovered By

Integration test: `suites/test_tools.py::test_quarantine_requires_write`

## Expected Behavior

The test (T-78) expects that `context_quarantine` requires Write capability (not Admin), per the vnc-014 IMPLEMENTATION-BRIEF.md capability table. An agent with only Write capability (auto-enrolled agents in permissive mode) should be able to quarantine entries.

## Actual Behavior

The server returns JSON-RPC error -32003: "Agent 'unknown-restricted-agent' lacks Admin capability." — confirming that `context_quarantine` still enforces `Capability::Admin` at line 1456 of `crates/unimatrix-server/src/mcp/tools.rs`.

## Server Logs

```
WARN rmcp::service: response error id=3 error=ErrorData { code: ErrorCode(-32003), message: "Agent 'unknown-restricted-agent' lacks Admin capability. Contact project admin.", data: None }
```

## Reproduction

```bash
cd product/test/infra-001
python -m pytest suites/test_tools.py::test_quarantine_requires_write -v
```

## Notes

Discovered during infra-001 integration testing (vnc-014 bug fix verification). The code change in `require_cap` at `tools.rs:1456` from `Capability::Admin` to `Capability::Write` was not applied as part of the vnc-014 audit_log migration (commit b5b30480). The capability table update is a separate pending change.

Not caused by the current feature under development (audit_fire_and_forget spawn_blocking → tokio::spawn fix).

dug-21:
# Bug Investigation Report: 580-investigator

## Bug Summary

`context_quarantine` enforces `Capability::Admin` at the gate (line 1456) and records `\"admin\"` in both audit log `capability_used` fields (lines 1500, 1538). The vnc-014 IMPLEMENTATION-BRIEF.md capability table (line 173) specifies `Capability::Write` for `context_quarantine`. Two integration tests were written against the spec and are marked `xfail` because the code contradicts it.

## Root Cause Analysis

[...]

PR #581, opened 2026-04-23
fix(audit): restore audit_log writes by fixing spawn_blocking + block_in_place panic (#579)

### LINKED ISSUE #579: bug: audit_log writes silently stopped ~2026-03-18
## Summary

The `audit_log` table has not received any new rows since **2026-03-18** (last `event_id` = 3759). The server is running normally, MCP tool calls are being processed, and the 4-column vnc-014 migration is present — but nothing is being written to the audit log.

## Evidence

- `SELECT MAX(event_id), datetime(MAX(timestamp), 'unixepoch') FROM audit_log` → `3759 | 2026-03-18 02:31:59`
- Server log shows active MCP tool calls (`unimatrix_server::mcp::tools`) with no corresponding audit write log lines
- GC cycle repeatedly logs `audit_log cleanup rows_deleted=0` across many maintenance ticks
- WAL file is 910KB — unrelated to audit writes (other tables)
- Related warning seen in log: `col-020: counter update failed: deserialization error: topic_delivery not found: vnc-014` — indicates the mcp::tools path is live but something in the write pipeline is broken

## Scope

Predates vnc-014 (merged 2026-04-22) by ~5 weeks. The regression window is somewhere around **2026-03-18** — need to identify what changed at that point.

## Likely Cause

The audit write is fire-and-forget (`spawn` or equivalent). A failure in the INSERT is silently swallowed with no log line, no panic, and no visible error to the caller. The INSERT may have broken due to a schema change, a missing column binding, or a query that references a column that no longer matches the struct.

## Investigation Starting Points

- `crates/unimatrix-server/src/mcp/tools.rs` — audit write call site
- Check git log ~2026-03-18 for schema or audit-path changes
- Add temporary error logging to the fire-and-forget audit spawn to surface the actual DB error
- Verify the `INSERT INTO audit_log` statement matches the current 12-column schema

## Impact

All MCP tool call attribution, credential tracking, and capability auditing has been blind since mid-March.

dug-21:
## Summary

- `audit_fire_and_forget` used `spawn_blocking + log_event`; `log_event` calls `block_in_place` internally, which panics on a blocking-pool thread. The panic was silently discarded by `let _ = ...`, causing all audit writes to fail since 2026-03-18.
- Fix: replace with `tokio::spawn + log_event_async` — the same pattern used by `insert_with_audit`, `correct_with_audit`, and both background audit emit functions.
- Added `tracing::warn!` on error and a doc comment acknowledging the JoinHandle drop trade-off.
- Added regression test `test_audit_fire_and_forget_persists_event` that polls the audit log with a 5s deadline; would have caught the original bug.

## Test plan

- [x] Regression test `test_audit_fire_and_forget_persists_event` passes
- [x] 4807 unit tests pass, 0 failures
- [x] 168 tools+lifecycle integration tests pass (1 pre-existing xfail filed as GH #580)
- [x] `cargo clippy -p unimatrix-server -- -D warnings` clean

Fixes #579

### DIFF
--- a/crates/unimatrix-server/src/server.rs
+++ b/crates/unimatrix-server/src/server.rs
@@ -508,7 +508,7 @@ impl UnimatrixServer {
     /// shutdown may be silently lost. Acceptable for observability logging (#579).
     pub(crate) fn audit_fire_and_forget(&self, event: AuditEvent) {
         let audit = Arc::clone(&self.audit);
-        let _ = tokio::spawn(async move {
+        tokio::spawn(async move {
             if let Err(e) = audit.log_event_async(event).await {
                 tracing::warn!(error = %e, "audit_fire_and_forget: write failed");
             }
--- a/crates/unimatrix-server/src/server.rs
+++ b/crates/unimatrix-server/src/server.rs
@@ -502,19 +502,17 @@ impl UnimatrixServer {
             .map_err(rmcp::ErrorData::from)
     }
 
-    /// Fire-and-forget audit event via `spawn_blocking`.
+    /// Fire-and-forget audit event via `tokio::spawn` + `log_event_async`.
     ///
-    /// Replaces direct `self.audit.log_event()` calls which would block the
-    /// async runtime thread on `store.lock_conn()` (#176).
+    /// The JoinHandle is intentionally dropped — audit writes in-flight at runtime
+    /// shutdown may be silently lost. Acceptable for observability logging (#579).
     pub(crate) fn audit_fire_and_forget(&self, event: AuditEvent) {
-        if tokio::runtime::Handle::try_current().is_ok() {
-            let audit = Arc::clone(&self.audit);
-            let _ = tokio::task::spawn_blocking(move || {
-                let _ = audit.log_event(event);
-            });
-        } else {
-            let _ = self.audit.log_event(event);
-        }
+        let audit = Arc::clone(&self.audit);
+        let _ = tokio::spawn(async move {
+            if let Err(e) = audit.log_event_async(event).await {
+                tracing::warn!(error = %e, "audit_fire_and_forget: write failed");
+            }
+        });
     }
 
     /// Insert a new entry and write an audit event.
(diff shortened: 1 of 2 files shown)


dug-21 (commented):
## Security Review — PR #581 (`bugfix/579-audit-fire-and-forget-spawn-blocking`)

**Risk Level: Low**

---

### Summary

This is a minimal, focused fix to `audit_fire_and_forget` in `server.rs`. The old implementation called `spawn_blocking` + `log_event`, which internally used `block_in_place` — an illegal combination that panicked silently and discarded every audit event routed through this path. The fix replaces it with `tokio::spawn` + `log_event_async`, matching the pattern already used by the inline audit spawns elsewhere in the same file and in `background.rs`.

No new dependencies. No new inputs. No access control changes.

---

### OWASP Assessment

**Injection** — no risk introduced. `AuditEvent` fields flow into parameterized sqlx queries with bound positional parameters (`?1` … `?12`). The `target_ids` field is serialized via `serde_json::to_string` before binding. No string interpolation into SQL. No change to this path in the diff.

**Access control** — not affected. `audit_fire_and_forget` is `pub(crate)` and called only from `async fn` tool handlers in `mcp/tools.rs`, all of which perform capability checks before reaching the audit call. This fix does not alter the capability-check path.

**Deserialization** — not affected. No new deserialization introduced.

**Input validation** — not changed. The store layer already applies a defensive guard on `AuditEvent.metadata` (empty → substituted with `{}`). The fix does not remove or bypass any existing validation.

**Secrets / credentials** — none present in the diff.

---

### Behavioral Change Analysis

The old code contained a runtime guard:
```rust
if tokio::runtime::Handle::try_current().is_ok() {
    tokio::task::spawn_blocking(move || { let _ = audit.log_event(event); });
} else {
    let _ = self.audit.log_event(event);
}
```

The new code removes that guard and calls `tokio::spawn` unconditionally. This is correct because every call site of `audit_fire_and_forget` is inside an `async fn` (verified across `mcp/tools.rs` — 9 call sites, all in `async fn context_*` handlers). There is no sync call site. If a future caller invokes this from outside a tokio runtime, `tokio::spawn` will panic; the old `else` branch is gone. This is a reasonable trade-off for an internal `pub(crate)` function, but worth noting.

The `background.rs` audit path is unaffected — it uses its own inline `tokio::spawn` + `log_event_async` and does not use `audit_fire_and_forget`.

---

### Blast Radius

Worst case: if `log_event_async` has a latent bug (e.g., write-pool exhaustion under high concurrency), all 9 audit call paths through `audit_fire_and_forget` log a `tracing::warn` and silently drop the event. This is the same silent-loss behavior as the old code (which panicked and dropped silently). The failure mode is **information loss in the audit log**, not data corruption or privilege escalation. The existing concurrent-starve regression test (`test_log_event_async_concurrent_does_not_starve`, 20 concurrent events) provides reasonable confidence that pool exhaustion is handled.

Shutdown-time loss is explicitly documented in the method comment and is an accepted trade-off for fire-and-forget observability logging.

---

### Regression Risk

Low. The fix narrows behavior: the old code had two paths (async runtime present / not present); the new code has one. All current callers are async, so the live path is unchanged. The regression test `test_audit_fire_and_forget_persists_event` uses a deadline poll loop with a 5 s timeout and `#[tokio::test(flavor = "multi_thread")]` — appropriate for async write-pool timing. It would have failed against the old code (verifying the regression was real).

---

### xfail Change in test_tools.py

`test_quarantine_requires_write` is marked `xfail` with a clear issue reference (GH #580) and a code location (`tools.rs:1456`). This is a pre-existing failure unrelated to the audit fix. The xfail annotation is correctly scoped (marks the test as expected-to-fail, not skipped) and does not suppress any new test that should be passing. No security concern.

---

### Findings

No blocking findings.

One observation (non-blocking): the removal of the `try_current()` guard means a future caller from a sync context will panic at `tokio::spawn`. The function is `pub(crate)` so this is an internal API contract. A short doc comment noting the caller-must-be-on-runtime assumption would make this contract explicit. Not required for merge.

---

**Blocking findings: none. Approved for merge.**

PR #586, opened 2026-05-10
fix(audit): atomic counter allocation eliminates UNIQUE constraint race (#584)

### LINKED ISSUE #584: bug: audit_log UNIQUE constraint violation on event_id — counter read+write races in WAL mode
## Summary

\`audit_fire_and_forget\` writes fail with:
\`\`\`
audit_fire_and_forget: write failed error=audit error: database error: error returned from database: (code: 1555) UNIQUE constraint failed: audit_log.event_id
\`\`\`

All tool-handler audit writes are affected. Pre-existing bug — fire-and-forget silently swallowed these errors before anyone was watching the daemon log.

## Root Cause

\`log_audit_event\` in \`crates/unimatrix-store/src/audit.rs:18-27\` uses a two-step counter read+write inside a \`BEGIN DEFERRED\` transaction:

\`\`\`rust
let current_id = counters::read_counter(&mut *txn, "next_audit_event_id").await?;
let id = if current_id == 0 { 1 } else { current_id };
counters::set_counter(&mut *txn, "next_audit_event_id", id + 1).await?;
\`\`\`

In SQLite WAL mode, \`BEGIN DEFERRED\` starts as a read transaction. Two concurrent \`tokio::spawn\` fire-and-forget writes can both open deferred transactions and read the same counter snapshot (say \`N\`) before either upgrades to a write lock. Both compute \`event_id = N\`, one inserts successfully, the other hits the UNIQUE constraint on \`audit_log.event_id\`.

## Evidence

- Error code 1555 = \`SQLITE_CONSTRAINT_UNIQUE\` on \`audit_log.event_id\`
- \`audit_fire_and_forget\` spawns a \`tokio::spawn\` per tool call — concurrent spawns race on the counter
- \`counters::set_counter\` uses \`INSERT OR REPLACE\` (not an atomic \`UPDATE ... RETURNING\`)
- \`write_pool_server()\` is used; if the pool allows >1 concurrent connection this compounds the race

## Fix

Replace the two-step read+set with a single atomic SQL increment:

\`\`\`sql
UPDATE counters SET value = value + 1 WHERE name = 'next_audit_event_id' RETURNING value
\`\`\`

This is atomic within SQLite regardless of transaction isolation level and eliminates the TOCTOU race. \`RETURNING\` is available since SQLite 3.35.0 (2021-03-12).

Alternatively: use \`BEGIN IMMEDIATE\` to acquire the write lock at transaction start, serializing all counter reads.

## Impact

- All audit_log writes from tool handlers (context_lookup, context_get, context_deprecate, context_status, context_briefing, context_quarantine, context_enroll, write_lesson_learned) silently fail — no audit trail is written for these operations
- Tool calls succeed normally; only the fire-and-forget audit path is affected
- Discovered while monitoring daemon logs during GH #582 diagnostic observation

dug-21:
## Summary

- Replaces two-step counter read+write in `log_audit_event` and `next_entry_id` with a single atomic `INSERT ... ON CONFLICT DO UPDATE SET value = value + 1 RETURNING value`
- Eliminates the TOCTOU race that caused `UNIQUE constraint failed: audit_log.event_id` (SQLite error 1555) when `write_max_connections=2`
- Fixes the same latent race in `next_entry_id` (identical pattern, same pool config exposure)
- Corrects `next_entry_id` counter seed from 1 to 0 in `db.rs` (first upsert returns 1, preserving contract)

## Changed files

- `crates/unimatrix-store/src/counters.rs` — added `next_counter` (atomic upsert-or-increment)
- `crates/unimatrix-store/src/audit.rs` — replaced 3-line read+set with `next_counter`
- `crates/unimatrix-store/src/db.rs` — counter seed corrected from 1 to 0
- `crates/unimatrix-store/src/pool_config.rs` — regression guard test: `write_max=2` is valid after this fix
- `crates/unimatrix-store/tests/sqlite_parity.rs` — parity assertion updated for seed=0

## New tests

- `test_next_counter_sequential_allocations` — basic sequential behavior
- `test_next_counter_concurrent_no_duplicate` — spawns two tasks on `max_connections=2` pool; asserts returned IDs are distinct (would have caught the original bug)
- `test_pool_config_write_max_2_is_valid` — regression guard documenting that `write_max=2` is safe after this fix

## Test plan

- [x] `cargo test -p unimatrix-store` — 294 passed, 0 failed
- [x] `cargo test --workspace` — all green
- [x] `cargo clippy --workspace -- -D warnings` — clean (pre-existing error in `auth.rs` unrelated)
- [x] Integration smoke suite — 23/23 passed
- [x] Integration tools + lifecycle suites — 168 passed, 0 failed

Fixes #584

### DIFF
--- a/crates/unimatrix-store/src/audit.rs
+++ b/crates/unimatrix-store/src/audit.rs
@@ -22,9 +22,7 @@ impl SqlxStore {
             .await
             .map_err(|e| StoreError::Database(e.into()))?;
 
-        let current_id = counters::read_counter(&mut *txn, "next_audit_event_id").await?;
-        let id = if current_id == 0 { 1 } else { current_id };
-        counters::set_counter(&mut *txn, "next_audit_event_id", id + 1).await?;
+        let id = counters::next_counter(&mut txn, "next_audit_event_id").await?;
 
         let target_ids_json = serde_json::to_string(&event.target_ids)
             .map_err(|e| StoreError::Serialization(e.to_string()))?;
--- a/crates/unimatrix-store/src/counters.rs
+++ b/crates/unimatrix-store/src/counters.rs
@@ -61,18 +61,41 @@ pub async fn decrement_counter(conn: &mut SqliteConnection, name: &str, delta: u
     set_counter(&mut *conn, name, current.saturating_sub(delta)).await
 }
 
+/// Atomically allocate the next value for a named counter.
+///
+/// Uses an upsert-or-increment form that is atomic within SQLite regardless
+/// of transaction isolation level, eliminating the TOCTOU race present in
+/// the two-step read_counter + set_counter pattern (bugfix-584).
+///
+/// Seeds the counter at 1 on first call (row absent or value=0 handled by
+/// the upsert INSERT path which always starts at seed=1).
+///
+/// Takes `&mut SqliteConnection` directly to avoid "not general enough" lifetime
+/// errors when the caller's future must be `Send` (e.g., inside `tokio::spawn`).
+pub async fn next_counter(conn: &mut SqliteConnection, name: &str) -> Result<u64> {
+    let val: Option<i64> = sqlx::query_scalar(
+        "INSERT INTO counters (name, value) VALUES (?1, 1)
+         ON CONFLICT(name) DO UPDATE SET value = value + 1
+         RETURNING value",
+    )
+    .bind(name)
+    .fetch_optional(&mut *conn)
+    .await
+    .map_err(|e| StoreError::Database(e.into()))?;
+
+    val.map(|v| v as u64)
+        .ok_or_else(|| StoreError::Database("next_counter RETURNING returned no row".into()))
+}
+
 /// Allocate the next entry ID within an open transaction.
 ///
-/// Reads `next_entry_id`, increments it, returns the pre-increment value.
-/// The returned value is always >= 1 (0 → 1 bootstrap).
+/// Atomically increments the `next_entry_id` counter and returns the allocated
+/// value. The returned value is always >= 1 (first call returns 1).
 ///
 /// Takes `&mut SqliteConnection` directly to avoid "not general enough" lifetime
 /// errors when the caller's future must be `Send` (e.g., inside `tokio::spawn`).
 pub async fn next_entry_id(conn: &mut SqliteConnection) -> Result<u64> {
-    let current = read_counter(&mut *conn, "next_entry_id").await?;
-    let id = if current == 0 { 1 } else { current };
-    set_counter(&mut *conn, "next_entry_id", id + 1).await?;
-    Ok(id)
+    next_counter(conn, "next_entry_id").await
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


dug-21 (commented):
## Security Review — Agent 584-security-reviewer

**Risk Level: LOW**

All tests pass (294/294). No blocking findings. Approving with observations.

---

### Finding 1: SQL Injection — NOT PRESENT
- **Severity**: n/a (confirmed safe)
- **Location**: counters.rs:76-80
- The `name` parameter is bound via `?1` placeholder, not interpolated into the SQL string. No injection surface.

### Finding 2: Integer Overflow on RETURNING cast — LOW, Non-Blocking
- **Severity**: low
- **Location**: counters.rs:86 (`val.map(|v| v as u64)`)
- SQLite stores INTEGER as i64. The `as u64` cast on a negative i64 (e.g., if someone externally sets `value = -1`) produces a very large u64 via wrapping, not a trap. In practice this counter only ever increases from 1, so the value would have to overflow i64::MAX (~9.2 × 10^18 rows) before the sign bit flips. This is not a realistic threat but the cast is unchecked.
- **Recommendation**: Not blocking for this fix. Future hardening: use `u64::try_from(v).map_err(...)` to surface the error explicitly rather than silently wrapping.
- **Blocking**: no

### Finding 3: Residual TOCTOU on `increment_counter` / `decrement_counter` — INFORMATIONAL
- **Severity**: informational
- **Location**: counters.rs:50-61; write.rs:87, 180-181, 218-219, 259
- This PR correctly atomicizes `next_counter` (ID allocation). The status-tracking counters (`active_count`, `total_deprecated`, etc.) still use the two-step `read_counter + set_counter` pattern. These are called inside transactions but with SQLite DEFERRED isolation, two concurrent writers under `write_max=2` could both read before either writes.
- **Mitigating factors**: (a) the default production config is still `write_max=1`; (b) status counters are advisory metrics, not primary keys — a count drift is observable and correctable, not a data-corruption risk; (c) this was pre-existing before this PR and out of scope for bugfix-584.
- **Recommendation**: File a follow-up issue to atomicize `increment_counter` / `decrement_counter` using the same `ON CONFLICT DO UPDATE SET value = value ± delta` pattern if `write_max=2` ever becomes the production default.
- **Blocking**: no

### Finding 4: Migration test fixtures seed `next_entry_id = 1` (not 0) — INFORMATIONAL
- **Severity**: informational
- **Location**: tests/migration_v10_to_v11.rs through migration_v24_to_v25.rs (all historical migration test scaffolding)
- The migration fixtures pre-seed `next_entry_id = 1` to represent a database that was already in use at that schema version. After upgrade, the first `next_counter` call on these fixtures returns 2 (not 1), skipping ID 1. ID gaps in monotonic sequences are safe — the system does not assume contiguous IDs. Fresh databases seeded at 0 correctly return 1 on first call.
- **Blocking**: no

---

### Blast Radius Assessment
The worst case for a subtle bug in `next_counter` is a UNIQUE constraint violation on `audit_log.event_id` or `ENTRIES.id` — the same failure mode that triggered bugfix-584 in the first place. The fix is additive (new function, callers switched), the upsert SQL is standard SQLite idiom, and the RETURNING clause is supported from SQLite 3.35.0 (bundled via sqlx 0.8). The atomic upsert form is safer than the two-step it replaced.

### Regression Risk
Low. The seed change (1 → 0 in db.rs) only affects fresh database initialization; it is load-bearing only for `next_entry_id`, and the new semantics (seed=0, first call returns 1) produce the same first allocated ID as before (was: read 1, allocate 1, write 2; now: upsert from 0, returns 1). Existing databases upgraded through migration are unaffected because `INSERT OR IGNORE` does not overwrite existing counter rows.

### Secrets / Dependencies
No hardcoded secrets. No new dependencies introduced.

---

**Summary**: The atomic upsert correctly eliminates the TOCTOU race for ID allocation counters. The change is minimal, well-tested (including a concurrency regression test), and does not introduce new OWASP-relevant risks. The informational findings are pre-existing or are acceptable tradeoffs documented in the PR.

PR #590, opened 2026-05-10
fix(audit): rename counter next_audit_event_id → next_audit_id (#587)

### LINKED ISSUE #587: bug: audit_log UNIQUE constraint violations persist after #586 — counter name mismatch (next_audit_event_id vs next_audit_id)
## Summary

The #586 fix (atomic counter allocation) is correct but incomplete. Audit writes are still failing with `UNIQUE constraint failed: audit_log.event_id` because the code uses the wrong counter name.

## Root Cause

The nxs-011 refactor (ff66537e) renamed the audit counter from `next_audit_id` → `next_audit_event_id` in code, but:
1. No migration updated the live database counter to the new name
2. `next_audit_event_id` was seeded at 0 as a fresh counter
3. `audit_log` already had 3759 rows from the `next_audit_id` era

**Live database state (observed 2026-05-10):**
```
next_audit_event_id = 0     ← what the code reads/writes
next_audit_id       = 3760  ← what the database actually tracks
MAX(audit_log.event_id) = 3759
```

So every new audit write allocates event_id=1, 2, 3, … from the stale counter, colliding with the 3759 existing rows.

## Evidence

```
sqlite3 unimatrix.db "SELECT name, value FROM counters WHERE name LIKE '%audit%';"
next_audit_event_id|0
next_audit_id|3760

sqlite3 unimatrix.db "SELECT MAX(event_id), COUNT(*) FROM audit_log;"
3759|3759
```

## Fix (investigation only — do not implement outside protocol)

In `crates/unimatrix-store/src/audit.rs`:
- Change counter name from `"next_audit_event_id"` → `"next_audit_id"`

In `crates/unimatrix-store/src/db.rs`:
- Update `INSERT OR IGNORE INTO counters ... VALUES ('next_audit_event_id', 0)` → `'next_audit_id'`

For existing databases: `INSERT OR IGNORE` is a no-op since `next_audit_id` already exists at the correct value. For fresh databases: seeds at 0, first `next_counter` call returns 1. Contract preserved.

## Note on #586 Fix

The atomic `INSERT … ON CONFLICT DO UPDATE … RETURNING` from #586 is correct and should be kept — it eliminates the real TOCTOU race. This issue is additive; both fixes are needed.

## Related

- Caused by: ff66537e (nxs-011 counter rename without migration)
- Blocked by: #584 / #586 (which fixed the race but not the name)

dug-21:
## Summary

- Counter name mismatch introduced by nxs-011 caused every audit write to collide with existing rows: code read/wrote `next_audit_event_id` (stuck at 0) while live databases track `next_audit_id` (at 3760+). This fix corrects both the call site and the DB seed, adds a shared const to make future renames compile-time-detectable, cleans the phantom row from live databases via schema v25→v26 migration, and ships the upgrade-continuity regression test that was missing across the entire #584/#586/#587 chain.
- Adds `pub const AUDIT_EVENT_COUNTER: &str = "next_audit_id"` in `counters.rs`; all sites bind the const.
- Schema v25→v26 migration DELETEs the phantom `next_audit_event_id` row; `INSERT OR IGNORE` on `next_audit_id` is a no-op for live DBs (row already exists at correct value).
- New test `test_audit_counter_upgrade_continuity`: pre-seeds N audit rows + counter at N, asserts next write returns N+1 with no UNIQUE violation.
- Pre-existing `test_restricted_agent_quarantine_allowed_write` failure xfailed with GH#589.

## Test plan

- [x] `test_audit_counter_upgrade_continuity` passes — reproduces and verifies the bug scenario
- [x] Full workspace unit tests: zero failures
- [x] Integration lifecycle+tools: 168 passed, 9 xfailed, 2 xpassed
- [x] Integration protocol+security: 31 passed, 1 xfailed (GH#589, pre-existing)
- [x] Clippy clean on changed crates
- [x] Gate 3 validator: PASS

Closes #587

### DIFF
--- a/crates/unimatrix-server/src/infra/audit.rs
+++ b/crates/unimatrix-server/src/infra/audit.rs
@@ -27,7 +27,7 @@ impl AuditLog {
     ///
     /// The caller provides all fields except `event_id` and `timestamp`,
     /// which are set by this method. The event_id is monotonically increasing
-    /// using counters["next_audit_event_id"].
+    /// using the `"next_audit_id"` counter (see `counters::AUDIT_EVENT_COUNTER`).
     pub fn log_event(&self, event: AuditEvent) -> Result<(), ServerError> {
         block_sync(self.store.log_audit_event(event))
             .map(|_| ())
--- a/crates/unimatrix-server/src/server.rs
+++ b/crates/unimatrix-server/src/server.rs
@@ -2272,13 +2272,13 @@ mod tests {
                 .await
                 .unwrap();
 
-            // Verify schema version is now current (25, vnc-014 audit_log migration)
+            // Verify schema version is now current (26, bugfix-587 audit counter rename)
             let version: i64 =
                 sqlx::query_scalar("SELECT value FROM counters WHERE name = 'schema_version'")
                     .fetch_one(store.read_pool_test())
                     .await
                     .unwrap();
-            assert_eq!(version, 25);
+            assert!(version >= 25, "schema_version must be >= 25, got {version}");
 
             // Verify backfill: quarantined entry should have pre_quarantine_status = 0
             let pre_q: Option<i64> =
@@ -2303,7 +2303,10 @@ mod tests {
                     .fetch_one(store.read_pool_test())
                     .await
                     .unwrap();
-            assert_eq!(version, 25, "schema version should remain 25 on re-open");
+            assert!(
+                version >= 25,
+                "schema_version must remain >= 25 on re-open, got {version}"
+            );
         }
     }
 
--- a/crates/unimatrix-store/src/audit.rs
+++ b/crates/unimatrix-store/src/audit.rs
@@ -13,8 +13,9 @@ use crate::schema::{AuditEvent, Outcome};
 impl SqlxStore {
     /// Append an audit event to the audit_log table.
     ///
-    /// Assigns `event_id` (monotonically increasing via `next_audit_event_id` counter)
-    /// and `timestamp` (current unix seconds). Returns the assigned event_id.
+    /// Assigns `event_id` (monotonically increasing via the `AUDIT_EVENT_COUNTER`
+    /// counter, `"next_audit_id"`) and `timestamp` (current unix seconds).
+    /// Returns the assigned event_id.
     pub async fn log_audit_event(&self, event: AuditEvent) -> Result<u64> {
         let pool = self.write_pool_server();
         let mut txn = pool
@@ -22,7 +23,7 @@ impl SqlxStore {
             .await
             .map_err(|e| StoreError::Database(e.into()))?;
 
-        let id = counters::next_counter(&mut txn, "next_audit_event_id").await?;
+        let id = counters::next_counter(&mut txn, counters::AUDIT_EVENT_COUNTER).await?;
 
         let target_ids_json = serde_json::to_string(&event.target_ids)
             .map_err(|e| StoreError::Serialization(e.to_string()))?;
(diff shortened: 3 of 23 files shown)


dug-21 (commented):
## Security Review — bugfix-587 (PR #590)

**Risk Level: LOW — No blocking findings.**

### Scope

Counter name mismatch fix: `next_audit_event_id` (phantom, stuck at 0) → `next_audit_id` (live counter). Schema v25→v26 migration. Changes touch `counters.rs`, `audit.rs`, `db.rs`, `migration.rs`, 15 migration test fixtures, `sqlite_parity.rs`, `server.rs`, and one Python xfail marker.

---

### OWASP Findings

**Injection — PASS.** All SQL in the v25→v26 migration uses literal string literals with no user-supplied input. The `AUDIT_EVENT_COUNTER` const is a compile-time Rust string bound via parameterized query in `db.rs`. No injection surface introduced.

**Access control — PASS.** The counter rename is internal to the store crate. No trust boundary changes. The xfail marker on `test_restricted_agent_quarantine_allowed_write` is correctly scoped to pre-existing GH#589; it does not suppress any new access control regression.

**Data integrity — PASS with one observation (non-blocking).** The v26 migration seeds `next_audit_id` from `COALESCE((SELECT MAX(event_id) FROM audit_log), 0)` using `INSERT OR IGNORE`. If a live database already has a `next_audit_id` row (possible on DBs that were never exposed to the nxs-011 bug), the INSERT is a no-op and the existing counter value is preserved — correct behavior. The DELETE of `next_audit_event_id` is idempotent on clean databases. Both paths are transaction-safe under ADR-003.

**Audit log integrity — PASS.** The append-only triggers (`audit_log_no_delete`, `audit_log_no_update`) created in v24→v25 operate on the `audit_log` table. The v26 migration's DELETE targets the `counters` table — no trigger conflict.

**Deserialization — NOT APPLICABLE.** No new deserialization of untrusted data.

**Secrets — PASS.** No hardcoded credentials, tokens, or API keys anywhere in the diff.

**Dependencies — PASS.** No new crate dependencies introduced.

---

### One Observation (Non-Blocking)

**Counter seed on the fresh-database path.** In `db.rs`, `create_tables_if_needed()` seeds `next_audit_id` at 0. The `next_counter()` atomic upsert then returns 1 on first call (INSERT path seeds at 1). This is correct and consistent with all other counters (`next_entry_id`, `next_log_id`, etc.). The new regression test `test_audit_counter_upgrade_continuity` explicitly exercises the live-database upgrade path (counter pre-seeded at N, next call returns N+1). Good coverage.

---

### Blast Radius

Worst case if the migration has a subtle bug: the v26 migration could fail to set `next_audit_id` correctly on a live database where both counter rows exist. Result would be UNIQUE constraint violations on audit writes — same symptom as the original bug, but caught immediately on the next audit write (not silent data corruption). The transaction boundary ensures schema_version stays at 25 on failure, making the failure detectable and retryable. No data loss path exists.

---

### Regression Risk

Low. All 15 historical migration fixture seeds now use `next_audit_id` (correct), which means they simulate databases that were always correct. Databases created during the nxs-011 bug window would have had `next_audit_event_id` as the active counter — those are exactly the live databases that the v26 migration remediates. The fixture change is consistent with the fix intent.

The loosening of schema version assertions in `migration_v24_to_v25.rs` and `server.rs` from `assert_eq!(version, 25)` to `assert!(version >= 25)` is appropriate — those tests exercise v24→v25 migration, not v25→v26, and should not be pinned to a specific final version. The idempotency test (`test_v25_migration_idempotent`) correctly checks that version does not change between the first and second open, which is the right invariant.

---

### Summary

The fix is minimal, correctly scoped, and follows established migration patterns (ADR-003). All SQL is parameterized or uses compile-time literals. The migration is idempotent and transactional. The regression test directly reproduces the original failure mode. No security concerns.

**Blocking findings: None.**

PR #618, opened 2026-05-20
fix: path_via_db() DB-backed BFS fallback for cold-start graph cache (#612)

### LINKED ISSUE #612: [infra-001] test_context_graph_path_found: no tick-force mechanism for path mode BFS graph rebuild
## Discovered By

Integration test: `suites/test_tools.py::test_context_graph_path_found` (vnc-020 AC-31)

## Expected Behavior

The integration test should be able to write entries, add graph edges, trigger a graph cache rebuild, and then call `context_graph(mode="path", ...)` to verify the BFS path is found.

## Actual Behavior

`path` mode uses the in-memory `TypedRelationGraph` which is rebuilt each tick (default 15 minutes). After writing entries and edges via MCP, the graph cache is not yet populated. Calling path mode immediately returns `{ found: false }` because the new edges are not yet in the in-memory snapshot.

The `fast_tick_server` fixture (UNIMATRIX_TICK_INTERVAL_SECS=30) reduces the wait but still requires a ~30 second sleep to be reliable — violating test determinism requirements.

## What Is Needed

A tick-force mechanism accessible through the MCP interface (or a test-mode `UNIMATRIX_TICK_NOW=1` env var) that immediately triggers a graph rebuild, allowing integration tests to validate path mode without sleeping.

## Server Logs

N/A — this is a test infrastructure limitation, not a server crash.

## Reproduction

```bash
cd product/test/infra-001
python -m pytest suites/test_tools.py::test_context_graph_path_found -v
```

## Notes

Discovered during infra-001 integration testing for vnc-020. Not caused by the current feature under development.
The path mode BFS logic is covered by unit tests (test_handle_path_1hop_from_id_not_in_hops, test_handle_path_2hop_from_id_not_in_hops, etc.).
The `test_context_graph_path_no_path_disconnected` (AC-14) is NOT affected — it tests `found: false` which is the default before graph rebuild.

dug-21:
## Summary

- `context_graph` path mode used BFS exclusively over the in-memory `TypedRelationGraph` snapshot, which is only rebuilt by the background tick (default 15 min). On cold-start or cycle-detected state, the snapshot is empty — path mode returned `{ found: false }` immediately after any `context_edge add` write.
- Added `path_via_db()`: BFS over live `GRAPH_EDGES` SQL using the existing `query_direct_neighbors` function, triggered when `state.use_fallback == true`.
- `handle_path` now extracts both `typed_graph` and `use_fallback` in the same lock guard, then branches before the in-memory BFS steps.
- Fan-out capped at `MAX_DB_NEIGHBORS_PER_NODE = 1000` via `.truncate()` post-fetch (consistent with filter/inverse mode caps).
- Removed `@pytest.mark.xfail` from `test_context_graph_path_found`; test now passes in ~8s without any sleep.

## Changed Files

- `crates/unimatrix-server/src/mcp/graph_read_path.rs` — `path_via_db()` + lock guard fix
- `crates/unimatrix-server/src/mcp/graph_read_path_tests.rs` — refactored to stay under 500 lines
- `crates/unimatrix-server/src/mcp/graph_read_path_db_tests.rs` — new cold-start regression test
- `product/test/infra-001/suites/test_tools.py` — xfail removed, stale comment updated
- `product/features/bugfix-612/` — agent reports

## Test Plan

- [x] `test_handle_path_db_fallback_cold_start_finds_path` — cold-start unit test (writes edges, no rebuild, asserts found=true)
- [x] Full workspace unit tests: all pass (3184 server lib tests)
- [x] Clippy: no new warnings in changed files
- [x] Integration smoke: 23/23 pass
- [x] `test_context_graph_path_found`: PASS (xfail removed, no sleep needed)
- [x] Graph integration suite: 30/30 pass

Closes #612


dug-21:
## Security Review — bugfix/612-path-db-fallback

---

### Scope

Cold read of `git diff main...HEAD` against the investigator and dev reports. Four changed artifacts:
- `crates/unimatrix-server/src/mcp/graph_read_path.rs` — new `path_via_db()` function + `use_fallback` branch
- `crates/unimatrix-server/src/mcp/graph_read_path_db_tests.rs` — new regression test
- `crates/unimatrix-server/src/mcp/graph_read_path_tests.rs` — test module linkage
- `product/test/infra-001/suites/test_tools.py` — xfail removal + comment cleanup

---

### OWASP Assessment

**SQL Injection (OWASP A03)**
Clean. `path_via_db()` passes edge_types to `query_direct_neighbors` as a `&[&str]` slice. The SQL in `graph_queries_neighbors.rs` builds `IN (?, ?, ...)` with numbered positional placeholders (`?1`, `?2`, ...) and binds each type string via `.bind(*t)`. The entry ID is also bound as `i64`. No string interpolation of untrusted values. The `relation_type` strings returned from the DB flow back to the caller as `PathHop.relation_type` — they do not re-enter any SQL. Injection risk: none, including second-order.

**Input Validation (OWASP A03)**
`path_via_db()` receives its inputs after all caller-side validation has already run (Steps 1–6 of `handle_path`): `from_id`/`to_id` validated as `u64`, depth clamped to `[1, 10]`, `edge_types` parsed through `parse_relation_types()` which rejects any unrecognized type name via an allowlist (`RelationType::from_str`). No new input surface introduced.

**Access Control (OWASP A01)**
No change to trust boundary. `use_fallback` is owned by the server process; MCP callers cannot influence it. The same `store.read_pool_server()` connection used everywhere else on the read path is reused — no escalation to the write pool.

**Denial of Service / Resource Exhaustion**
`MAX_DB_NEIGHBORS_PER_NODE = 1000` caps DB rows fetched per BFS hop via `.truncate()`. Combined with `max_depth` (max 10) and the visited set preventing re-enqueue, the effective frontier is bounded by distinct entry count. Identical risk profile to the in-memory BFS.

**Deserialization / Secrets / Dependencies**
No new deserialization of untrusted data. No hardcoded credentials, tokens, or API keys anywhere in the diff. No new crates added.

---

### Behavioral Divergence (documented, not a security concern)

`path_via_db()` documents an intentional behavioral gap: nodes with > 1000 outgoing edges of the requested type may have some neighbors silently skipped in DB mode, while the in-memory path has no per-hop fan-out cap. Failure mode if the cap fires is `found: false` (false negative), not information disclosure or privilege escalation. This divergence vanishes once the background tick fires and `use_fallback` becomes `false`.

---

### Blast Radius

If `path_via_db()` has a subtle bug (visited-set key regression, depth off-by-one):
- **Failure mode**: returns `found: false` when a path exists, or returns a path missing supersession resolution. Safe failure — no data corruption, no privilege escalation, no information disclosure.
- **Scope**: `context_graph path` calls only while `use_fallback = true` (cold-start window or post-cycle-detection). In-memory BFS is entirely untouched.
- **Blast boundary**: read-only path. No write paths touched.

---

### Regression Risk

Low. The change is additive — new branch fires only when `use_fallback = true`. Existing in-memory BFS (Steps 8–13 of `handle_path`) is unchanged. The xfail removal is correct: the assertion is identical, not weakened. The new unit test correctly simulates cold-start without `rebuild()` and asserts full path shape including ADR-005 invariant (from_id not in hops).

---

**No blocking findings. Cleared for merge.**

PR #620, opened 2026-05-20
fix(graph): reject resolve_supersessions on inverse/filter, add coverage tests (#616 #613)

### LINKED ISSUE #613: [infra-001] Add min_age_days end-to-end integration test for context_graph filter mode
## Summary

Gate 3c (vnc-020) flagged a non-blocking coverage gap: `min_age_days` is not exercised end-to-end in the infra-001 integration suite. The Q10 stale-Goal pattern (`min_age_days=30` combined with `max_edge_count=0`) is the primary use case for this parameter.

## What's missing

A `test_context_graph_filter_min_age_days` integration test in `test_tools.py` that:
1. Writes goal entries with different `created_at` timestamps (some old, some recent)
2. Calls `context_graph(mode="filter", category="goal", min_age_days=30, max_edge_count=0, edge_types=["Advances"])`
3. Asserts only old entries with no Advances edges are returned

## Context

- `max_edge_count=0` IS integration-tested (AC-29 passes)
- `min_age_days` SQL is correct (uses `strftime('%s','now') - ? * 86400` — INTEGER epoch)
- Gap is end-to-end combination only

Related: vnc-020 (#598)

### LINKED ISSUE #616: inverse/filter mode: deprecated entries produce false positives when superseded active entry has edges
## Summary

`context_graph(mode="inverse")` is the primary tool for Q9 uncited-source detection: find Source entries with no incoming Cites edges. The mode currently ignores `resolve_supersessions` (it silently has no effect on SQL-only modes).

This creates a correctness gap for the Q9 use case:

- Entry A (deprecated, superseded by B) has no incoming Cites edges
- Entry B (active successor of A) has incoming Cites edges
- Calling `inverse(category="source", missing_edge_types=["Cites"])` returns A — a false positive

A deprecated entry that *looks* uncited will appear in results when its active successor is correctly cited. Agents running gap detection would act on stale entries.

## Options

**Option A — Reject `resolve_supersessions` on inverse/filter modes** (strict): Return a validation error if `resolve_supersessions=true` is passed to inverse or filter. Callers know the limitation and cannot accidentally pass it expecting supersession behavior.

**Option B — Implement supersession-aware filtering** (correct): Add a WHERE clause: `AND e.superseded_by IS NULL` (or `e.status = 0`) so deprecated entries are excluded from inverse/filter results by default. This is different from resolve_supersessions; it just means "only active entries."

**Option C — Add `active_only` parameter** (explicit): New boolean flag defaulting to true that filters `e.status = 0`.

Note: the existing SQL already includes `AND e.status = 0`, which excludes deprecated entries. If this is already in the query, the issue may be moot — verify by checking `handle_inverse` and `handle_filter` SQL construction.

## Context

- Flagged during vnc-020 (#598) post-delivery review
- `inverse` and `filter` modes already use `AND e.status = 0` — verify whether this resolves the issue before implementing anything
- If `status = 0` already excludes deprecated entries, this issue can be closed as won't-fix with a note

Related: vnc-020 (#598)

dug-21:
## Summary

Closes #616 · Closes #613

**#616 — inverse/filter: deprecated entries false positive (won't-fix + hardening)**

The alleged bug does not exist: `handle_inverse` and `handle_filter` already enforce `AND e.status = 0`, which excludes deprecated entries before the antijoin runs. No false positive is possible.

Two additions close the correctness gap and prevent future confusion:

- **Behavioral change**: `validate_inverse_params` and `validate_filter_params` now reject `resolve_supersessions=Some(_)` with an explanatory error. Previously accepted as a silent no-op; this was inconsistent with chain mode which already rejects it explicitly. Error message follows chain mode's explanatory style: `"resolve_supersessions has no effect in {mode} mode — only active entries are returned"`.
- **Regression test**: `test_handle_inverse_deprecated_superseded_entry_absent` — inserts an entry, supersedes it to an active successor that has an incoming Cites edge, calls `inverse`, and asserts the deprecated entry is absent. Includes a new `set_superseded_by` helper (no such helper existed). Three validation unit tests also cover the new rejection behavior.

**#613 — filter mode: `min_age_days` end-to-end integration test**

`max_edge_count=0` was already integration-tested (AC-29). `min_age_days` had no end-to-end path through infra-001. Adds `test_context_graph_filter_min_age_days`: backdates two entries via direct SQLite UPDATE, calls `filter(min_age_days=30, max_edge_count=0)`, and asserts only the old entries are returned.

## Changed files

| File | Change |
|------|--------|
| `crates/unimatrix-server/src/mcp/graph_read_validation.rs` | Reject `resolve_supersessions` in inverse + filter validators |
| `crates/unimatrix-server/src/mcp/graph_read_tests.rs` | 3 validation unit tests for rejection behavior |
| `crates/unimatrix-server/src/mcp/graph_read_inverse_integration_tests.rs` | `set_superseded_by` helper + `test_handle_inverse_deprecated_superseded_entry_absent` |
| `product/test/infra-001/suites/test_tools.py` | `test_context_graph_filter_min_age_days` (#613) |

## Test plan

- [x] 5,146 unit tests pass (0 failed)
- [x] 185 tools suite tests pass (0 failed, 3 xfailed pre-existing)
- [x] Clippy clean
- [x] Gate 3 validation: PASS (13/13 checks)

### DIFF
--- a/crates/unimatrix-server/src/mcp/graph_read_inverse_integration_tests.rs
+++ b/crates/unimatrix-server/src/mcp/graph_read_inverse_integration_tests.rs
@@ -64,6 +64,19 @@ async fn deprecate_entry(store: &SqlxStore, entry_id: u64) {
         .expect("deprecate entry");
 }
 
+/// Deprecate an entry and record it as superseded by another entry.
+/// Sets status=1 (Deprecated) and superseded_by=successor_id directly in the DB.
+async fn set_superseded_by(store: &SqlxStore, deprecated_id: u64, successor_id: u64) {
+    sqlx::query(
+        "UPDATE entries SET status = 1, superseded_by = ?1, updated_at = 1700000002 WHERE id = ?2",
+    )
+    .bind(successor_id as i64)
+    .bind(deprecated_id as i64)
+    .execute(store.write_pool_test())
+    .await
+    .expect("set superseded_by");
+}
+
 fn inverse_params(
     category: Option<&str>,
     missing_edge_types: Option<Vec<&str>>,
@@ -249,5 +262,54 @@ async fn test_handle_inverse_empty_category_in_db_returns_empty_not_error() {
         .expect("handle_inverse");
     assert_eq!(resp.entries.len(), 0);
     assert_eq!(resp.total_returned, 0);
+    store.close().await.unwrap();
+}
+
+// ---------------------------------------------------------------------------
+// #616B — Deprecated-superseded entry absent from inverse results
+// ---------------------------------------------------------------------------
+
+#[tokio::test]
+async fn test_handle_inverse_deprecated_superseded_entry_absent() {
+    // #616B: A deprecated entry (even one that is part of a supersession chain)
+    // must NOT appear in inverse results. The AND e.status = 0 guard excludes
+    // all non-active entries regardless of edge presence.
+    //
+    // Fixture:
+    //   entry_a — category="source", deprecated, superseded_by entry_b
+    //   entry_b — category="source", active, has an incoming Cites edge
+    //   donor   — category="decision", active, provides the Cites edge to entry_b
+    //
+    // Expected:
+    //   entry_a absent — deprecated (status=1), excluded by status guard.
+    //   entry_b absent — active but has an incoming Cites edge, excluded by antijoin.
+    let (store, _dir) = open_test_store().await;
+
+    let donor_id = insert_entry(&store, "decision").await;
+    let id_a = insert_entry(&store, "source").await;
+    let id_b = insert_entry(&store, "source").await;
+
+    // Give entry_b an incoming Cites edge so it's excluded by the antijoin.
+    insert_edge(&store, donor_id, id_b, "Cites").await;
+
+    // Deprecate entry_a and mark it as superseded_by entry_b.
+    set_superseded_by(&store, id_a, id_b).await;
+
+    let params = inverse_params(Some("source"), Some(vec!["Cites"]), None);
+    let resp = handle_inverse(&store, &params)
+        .await
+        .expect("handle_inverse");
+
+    let ids: Vec<u64> = resp.entries.iter().map(|e| e.id).collect();
+
+    assert!(
+        !ids.contains(&id_a),
+        "deprecated-superseded entry_a must not appear (status guard); ids={ids:?}"
+    );
+    assert!(
+        !ids.contains(&id_b),
+        "active entry_b (has Cites edge) must not appear (antijoin); ids={ids:?}"
+    );
+
     store.close().await.unwrap();
 }
--- a/crates/unimatrix-server/src/mcp/graph_read_tests.rs
+++ b/crates/unimatrix-server/src/mcp/graph_read_tests.rs
@@ -299,6 +299,87 @@ fn test_node_index_for_unknown_node_returns_none() {
 }
 
 // -----------------------------------------------------------------------
+// #616: resolve_supersessions rejected on inverse and filter modes
+// Only active entries are returned by both modes; the parameter has no effect.
+// -----------------------------------------------------------------------
+
+#[test]
+fn test_validate_inverse_rejects_resolve_supersessions() {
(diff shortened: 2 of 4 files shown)


dug-21 (commented):
## Security Review — PR #620 (bugfix-616 + #613)

**Risk Level: Low**

Reviewed: graph_read_validation.rs, graph_read_inverse_integration_tests.rs, graph_read_tests.rs, test_tools.py (test_context_graph_filter_min_age_days)

---

### Finding 1 — Direct SQLite write in E2E test bypasses server isolation
- **Severity**: Low (test code only)
- **Location**: product/test/infra-001/suites/test_tools.py, test_context_graph_filter_min_age_days (~line 4822)
- **Description**: The test inlines the db_path derivation (sha256 digest, ~/.unimatrix/{hash}/unimatrix.db) rather than using the existing _compute_db_path helper that is already established in this file and used by every other direct-SQL test (lines 956-964). This is a copy-paste of private infrastructure knowledge. If the server ever changes its path derivation (e.g., encoding, hash length, base dir), this test will silently connect to a non-existent DB and create a new empty SQLite file at that path rather than failing loudly, causing a false-positive pass.
- **Recommendation**: Replace the inlined path derivation with _compute_db_path(server.project_dir). The helper already exists in the same file.
- **Blocking**: No — test-only, does not affect production code. But the inconsistency is a latent reliability risk.

### Finding 2 — resolve_supersessions rejection covers Some(false), which is correct
- **Severity**: Informational (positive finding)
- **Location**: graph_read_validation.rs:200, graph_read_validation.rs:242
- **Description**: The guard checks is_some() rather than == Some(true). This is the correct design. A caller sending resolve_supersessions=false on inverse/filter modes would be equally confused about the parameter's semantics; rejecting any presence of the field gives clear, unambiguous feedback regardless of value. The unit test test_validate_inverse_resolve_supersessions_false_also_rejected explicitly documents this intent.

### Finding 3 — No injection risk in validation layer
- **Severity**: Informational (positive finding)
- **Location**: graph_read_validation.rs
- **Description**: The new validation code operates entirely on already-deserialized GraphParams struct fields (Option<bool> checks). No string interpolation of user-controlled data into SQL, format strings, or shell commands. Error messages include the mode name (a constant string from the match arm) — not the user-supplied input — so there is no reflected injection surface.

### Finding 4 — Rust test helpers use parameterized queries
- **Severity**: Informational (positive finding)
- **Location**: graph_read_inverse_integration_tests.rs:69-78
- **Description**: set_superseded_by uses sqlx parameterized binding (?1, ?2) — no SQL injection risk. Bounded to test-support feature flag only (write_pool_test is #[cfg(feature = test-support)]).

---

### OWASP Assessment
- **Injection**: No risk. Validation is pure Option checks on deserialized struct fields.
- **Broken access control**: No change. The fix adds rejection, not new allowances.
- **Security misconfiguration**: No new configuration surface.
- **Input validation**: Fix improves the posture — previously a no-op parameter was silently accepted; now it is explicitly rejected with an actionable error message.
- **Deserialization**: No change to deserialization paths.

### Blast Radius
The fix is additive-rejection: callers who previously sent resolve_supersessions on inverse/filter modes (and received silently wrong results) now receive a clear error. The worst case if this guard has a subtle bug is that a valid caller receives an unexpected rejection. That is a recoverable user-experience error, not data corruption or privilege escalation. No write paths are touched.

### Regression Risk
Low. The change adds new Err arms to two validation functions. Existing callers that do not supply resolve_supersessions on these modes are unaffected — both new guards are gated on is_some() and the default for an Option field is None. The three new unit tests and the integration test collectively cover the new paths.

### Dependencies
No Cargo.toml or Cargo.lock changes. Zero new dependencies.

### Secrets
No hardcoded credentials, tokens, or API keys in any changed file.

---

**Summary**: The production change is minimal, correct, and additive. The single actionable item (Finding 1) is test-only: inline db_path derivation should use _compute_db_path(server.project_dir) for consistency and reliability. No blocking findings.

dug-21 (commented):
Security review posted — see full findings in the comment above. Risk level: Low. No blocking findings.

PR #624, opened 2026-05-21
fix(graph): subgraph and neighbors depth>1 fall back to DB on cold-start (#623)

### LINKED ISSUE #623: bug: context_graph subgraph mode returns depth_reached=0 for all seeds — in-memory graph cache not populated
## Summary

`context_graph(mode=\"subgraph\")` consistently returns `depth_reached=0` with no nodes or edges for any seed entry, including entries with confirmed edges in GRAPH_EDGES. The in-memory TypedGraphState graph cache appears not to be populated after maintenance tick completion.

## Reproduction

```
context_graph(mode="subgraph", seed_ids=[4545], max_depth=2, direction="both")
→ {"nodes":[{...seed only...}], "edges":[], "depth_reached":0, "truncated":false}
```

Same result for old entries with known edges (e.g. #405, ADR with co-access and Informs edges).

## Evidence

- `neighbors(mode="neighbors", id=4545, depth=1, direction="both")` returns correct edges from live DB:
  - #4545 → Advances → #4544
  - #4558 → Advances → #4545
  - #4559 → Advances → #4545  
  - #4560 → Advances → #4545
- `context_status` confirms maintenance tick fired (last maintenance: 2026-05-21T20:27:29Z, coherence: 1.0000)
- Subgraph returns `depth_reached=0` both before and after tick
- Affects ALL entries, not just newly created ones — old entry #405 also returns empty subgraph
- Per tool docs: "Seed IDs not present in the graph return an empty result" — confirming seed nodes are not being indexed in the cache

## Expected behaviour

After maintenance tick, the TypedGraphState in-memory graph should contain all entries present in GRAPH_EDGES as traversable nodes. `subgraph(seed=[4545], depth=2)` should return the seed, its neighbors, and their neighbors.

## Actual behaviour

BFS cannot start — seed node is not found in the graph index. `depth_reached` always 0. No edges returned regardless of seed, depth, or direction.

## Notes

- `neighbors` at depth=1 is unaffected (queries live DB directly)
- `neighbors` at depth>1 untested but likely affected (same cache dependency)
- Regression introduced at or after context_graph delivery (vnc-017/018/019/020)
- Workaround: use `neighbors(depth=1)` with incremental hops for multi-level traversal

dug-21:
## Summary

- `handle_subgraph` and `neighbors_bfs` (depth>1) both read `state.typed_graph` from the `TypedGraphState` lock without checking `state.use_fallback`. When `use_fallback=true` (cold-start or supersession cycle detected), `typed_graph` is an empty `TypedRelationGraph` — BFS never starts, `depth_reached=0` for all seeds.
- Added `subgraph_via_db` and `neighbors_via_db` DB fallback functions mirroring the pattern from `graph_read_path.rs` (GH #612). Both forward `edge_types` and `direction` explicitly, preserve edge dedup, dangling-edge filter, and `MAX_EDGES_UPPER` guard.

## Test plan

- [x] `test_subgraph_use_fallback_true_with_real_entries_falls_back_to_db` — real store, `TypedGraphState::new()` directly (cold-start), assert `edges.len()==1`, `depth_reached==1`
- [x] `test_subgraph_use_fallback_true_direction_incoming_forwarded` — verifies `direction` param is forwarded, not hard-coded
- [x] `test_neighbors_bfs_use_fallback_true_depth_gt1_falls_back_to_db` — equivalent for neighbors
- [x] 3,197 unit tests pass, 0 failures
- [x] 245 integration tests pass, 8 pre-existing xfailed (unchanged), 1 pre-existing xpass (unrelated, separate follow-up)
- [x] Bugfix gate: PASS (all 11 checks)

Closes #623

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3483,7 +3483,7 @@ checksum = "39ec24b3121d976906ece63c9daad25b85969647682eee313cb5779fdd69e14e"
 
 [[package]]
 name = "unimatrix-adapt"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "bincode 2.0.1",
  "ndarray",
@@ -3497,7 +3497,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-core"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "serde",
  "serde_json",
@@ -3510,7 +3510,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-embed"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "approx",
  "dirs",
@@ -3523,7 +3523,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-engine"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "dirs",
  "nix 0.31.2",
@@ -3542,7 +3542,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-learn"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "bincode 2.0.1",
  "ndarray",
@@ -3555,7 +3555,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-observe"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "bincode 2.0.1",
  "serde",
@@ -3570,7 +3570,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-server"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "bincode 2.0.1",
  "clap",
@@ -3610,7 +3610,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-store"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "base64 0.22.1",
  "bincode 2.0.1",
@@ -3625,7 +3625,7 @@ dependencies = [
 
 [[package]]
 name = "unimatrix-vector"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "anndists",
  "hnsw_rs",
--- a/crates/unimatrix-server/src/mcp/graph_read_neighbors.rs
+++ b/crates/unimatrix-server/src/mcp/graph_read_neighbors.rs
@@ -1,7 +1,9 @@
 //! Neighbors mode handler for context_graph (vnc-018, ADR-005).
 //!
 //! depth=1 → live SQL on GRAPH_EDGES (always fresh).
-//! depth>1 → BFS over in-memory TypedRelationGraph (tick-window staleness).
+//! depth>1 → BFS over in-memory TypedRelationGraph (tick-window staleness) when warm;
+//!           falls back to live SQL via `neighbors_via_db` when `use_fallback=true`
+//!           (cold-start or cycle-detected) — GH #623.
 //!
 //! Declared as a sub-module of `graph_read.rs` via `#[path]`.
 
@@ -171,6 +173,8 @@ pub(super) async fn handle_neighbors(
         )
         .await?
     };
+    // Note: neighbors_bfs reads use_fallback inside the lock guard and delegates to
+    // neighbors_via_db when use_fallback=true (cold-start / cycle-detected) — GH #623.
 
     Ok(NeighborsResponse { edges })
 }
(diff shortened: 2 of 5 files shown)


dug-21 (commented):
## Security Review — bugfix/623-subgraph-use-fallback

**Risk Level: LOW**

Reviewed by: uni-security-reviewer (fresh context, no prior session knowledge)

---

### Summary

The PR adds two DB-fallback BFS functions (`subgraph_via_db`, `neighbors_via_db`) that execute live SQL against GRAPH_EDGES when `use_fallback=true`. The change is minimal and well-scoped. Input validation from the existing `handle_subgraph`/`handle_neighbors` callers is preserved before the fallback dispatch. No injection vectors, no secret leakage, no new dependencies.

One low-severity finding is noted (no result-set cap in `neighbors_via_db`). All findings are non-blocking.

---

### Finding 1 — Missing result-set cap in `neighbors_via_db` (response size unbounded)

**Severity**: Low
**Location**: `graph_read_neighbors.rs:381-459`
**Blocking**: No

`neighbors_via_db` has `MAX_DB_NEIGHBORS_PER_NODE=1000` (per-node fan-out cap) and a visited-set dedup, but no ceiling on the total number of edges returned in `result_edges`. With depth=10 and 15 edge types plus direction=both, a sufficiently large graph during cold-start could produce a very large response vector.

For comparison, `subgraph_via_db` has `max_nodes` (enforced upper bound of 200) and `MAX_EDGES_UPPER=1000`. `neighbors_via_db` has neither.

The visited-set dedup (keyed by node_id) limits unique nodes, but does not cap the edge list. Practical impact is low because cold-start is transient and the visited set prevents exponential re-expansion, but the asymmetry with `subgraph_via_db` is worth noting for future hardening.

**Recommendation**: Add a response edge count cap (e.g., mirror `MAX_EDGES_UPPER`) to `neighbors_via_db` in a follow-up ticket, or add a note to the existing edge-count cap tracking issue. Not blocking for this fix.

---

### Finding 2 — `debug_assert` omitted in `subgraph_via_db` (minor contract gap)

**Severity**: Low
**Location**: `graph_read_subgraph.rs:519-527`
**Blocking**: No

The warm path (`handle_subgraph`) uses `debug_assert!(collected_edges.len() <= MAX_EDGES_UPPER)` before calling `fetch_edge_metadata`. The cold-start path (`subgraph_via_db`) instead uses a hard `truncate(MAX_EDGES_UPPER)` before the call — which is actually **stricter** than the debug_assert (a hard truncate fires in release builds; a debug_assert does not). This is acceptable. However, the comment says "excess edges lose metadata but are still present in the response shape" — since the truncate fires before computing `depth_reached` and building `edges`, edges beyond the cap are silently dropped from the response entirely, not just missing metadata. The comment is slightly misleading. Not a correctness bug, but worth a doc fix.

---

### OWASP Assessment

| Concern | Status |
|---|---|
| SQL Injection | Not present. `query_direct_neighbors` uses parameterized bindings exclusively. Edge type strings are validated via `RelationType::from_str` before reaching SQL (enum guard). |
| Input Validation | Preserved. All depth/direction/edge_type validation in `handle_subgraph` and `handle_neighbors` runs before the fallback dispatch. Fallback functions receive already-validated, typed parameters. |
| Path Traversal | Not applicable to this change. |
| Broken Access Control | Not applicable. The MCP capability check runs in `tools.rs` before `handle_graph`, unchanged by this diff. |
| Deserialization | Not applicable. No new deserialization of untrusted data. |
| Secrets / Credentials | None. No hardcoded tokens, keys, or passwords in the diff. |
| Error Handling | Errors propagate cleanly via `Result`/`ErrorData`. `tracing::error!` logs include node ID and error string — appropriate for an internal system. No internal state leaked to callers beyond error codes. |

---

### Blast Radius Assessment

Worst case: during cold-start, a caller with `depth=10, direction=both` on a high-degree node receives a large edge list that takes longer than expected to hydrate. This is a latency issue, not a security issue. The visited-set dedup prevents exponential BFS re-expansion. The per-node fan-out cap (1000) bounds each SQL round-trip.

If `subgraph_via_db` or `neighbors_via_db` has a subtle bug, the failure mode is: wrong graph results returned during cold-start only (no data corruption, no privilege escalation, no information disclosure beyond what the warm path would return). The fallback is read-only against the existing `read_pool_server`.

---

### Regression Risk

**Low.** The fallback path is conditional on `use_fallback=true`, which is set only during cold-start (server startup before first tick) and cycle-detected states. The warm path (`use_fallback=false`) is structurally unchanged — the lock guard now extracts one additional boolean field, which is a negligible change. Both regression tests (`test_subgraph_use_fallback_true_with_real_entries_falls_back_to_db`, `test_neighbors_bfs_use_fallback_true_depth_gt1_falls_back_to_db`) use a real `SqlxStore` and directly assert the fix.

---

### Dependency Safety

No new dependencies introduced. Version bump is 0.6.3 → 0.7.0 across all workspace crates (Cargo.lock only — consistent with a version bump commit, no external crate changes).

---

**Verdict: APPROVE — no blocking findings.**

PR #627, opened 2026-05-22
fix(init): write commented default config.toml on unimatrix init (#626)

### LINKED ISSUE #626: init: does not create a default config.toml
## Summary

When unimatrix initializes for a new project, `ensure_data_directory` creates `~/.unimatrix/{hash}/` and `~/.unimatrix/{hash}/vector/` but neither the global (`~/.unimatrix/config.toml`) nor per-project (`~/.unimatrix/{hash}/config.toml`) config file is ever written.

## Expected Behavior

`init` should write a commented default `config.toml` (at minimum the global path) that documents all available options and their compiled defaults, giving users a discoverable starting point for customization.

## Actual Behavior

No config file is created. The server silently falls back to compiled defaults. Users must read source or documentation to discover what can be configured and what the defaults are.

## Relevant Files

- `crates/unimatrix-engine/src/project.rs` — `ensure_data_directory` (creates dirs but not config)
- `crates/unimatrix-server/src/infra/config.rs` — `load_config` (reads config if present, otherwise uses compiled defaults)

## Acceptance Criteria

- On first init, a default `config.toml` is written to `~/.unimatrix/config.toml` (global) if one does not already exist
- The file is fully commented, documenting each field and its default value
- Existing config files are never overwritten

dug-21:
## Summary

- Adds `write_default_config_if_absent(path, force)` helper to `config.rs` that atomically writes a fully-commented default `config.toml` using `create_new` semantics (TOCTOU-safe; `AlreadyExists` treated as success)
- Adds `DEFAULT_CONFIG_TOML` static string documenting every config section, field, and compiled default
- Wires the helper into the `unimatrix init` CLI command targeting the **project-specific** path (`~/.unimatrix/{hash}/config.toml`)
- Adds `--force` flag to overwrite an existing config
- `load_config` is unchanged — remains a pure reader with no side effects

## Test plan

- [ ] `test_write_default_config_creates_file_when_absent` — file created, parses as valid `UnimatrixConfig`
- [ ] `test_write_default_config_does_not_overwrite_without_force` — existing content preserved
- [ ] `test_write_default_config_overwrites_with_force` — content replaced when `--force`
- [ ] `test_write_default_config_succeeds_even_if_write_fails` — no panic on read-only parent
- [ ] Full workspace test suite: 4398 pass
- [ ] Integration smoke tests: 23/23 pass

Closes #626

### DIFF
--- a/crates/unimatrix-server/src/infra/config.rs
+++ b/crates/unimatrix-server/src/infra/config.rs
@@ -2994,6 +2994,309 @@ fn check_permissions(path: &Path) -> Result<(), ConfigError> {
 }
 
 // ---------------------------------------------------------------------------
+// Default config TOML template
+// ---------------------------------------------------------------------------
+
+/// Fully-commented TOML template written by `unimatrix version --project-dir`.
+///
+/// Every section, field, and its compiled default is documented inline.
+/// This is a static string — it is NOT serialised from `UnimatrixConfig`
+/// (structs are `Deserialize`-only per existing conventions; no `Serialize` is added).
+///
+/// Keep in sync with struct defaults. When a new field is added to any config
+/// struct, add the matching commented line here.
+pub static DEFAULT_CONFIG_TOML: &str = r#"# Unimatrix configuration file.
+# All values shown are the compiled defaults. Uncomment and edit to override.
+# File: ~/.unimatrix/{project-hash}/config.toml
+#
+# Two-level config hierarchy:
+#   ~/.unimatrix/config.toml           -- global (applies to all projects)
+#   ~/.unimatrix/{hash}/config.toml    -- per-project (overrides global)
+# Per-project values replace global values field-by-field (replace semantics).
+# List fields replace entirely — no append.
+
+# ---------------------------------------------------------------------------
+# [profile] — knowledge-lifecycle preset
+# ---------------------------------------------------------------------------
+# Named presets map to preset weight tables.
+# Valid values: "collaborative" (default), "authoritative", "operational",
+#               "empirical", "custom"
+# Use "custom" with [confidence] weights AND [knowledge] freshness_half_life_hours.
+[profile]
+# preset = "collaborative"
+
+# ---------------------------------------------------------------------------
+# [knowledge] — categories and freshness configuration
+# ---------------------------------------------------------------------------
+[knowledge]
+# Allowed entry categories. Defaults are the 5 built-in INITIAL_CATEGORIES.
+# Each entry: lowercase letters (a-z), digits (0-9), hyphens (-), underscores (_).
+# Maximum 64 entries; each at most 64 characters.
+# categories = ["decision", "convention", "pattern", "lesson-learned", "procedure"]
+
+# Categories that receive a provenance boost in search re-ranking.
+# Must be a subset of categories above.
+# boosted_categories = ["lesson-learned"]
+
+# Categories eligible for automated lifecycle management.
+# Must be a subset of categories above.
+# adaptive_categories = ["lesson-learned"]
+
+# Optional override for the freshness half-life used in confidence scoring.
+# None = use the active preset's built-in value.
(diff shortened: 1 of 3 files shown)


dug-21 (commented):
## Security Review — bugfix/626-write-default-config-on-init

**Risk Level: LOW**

---

### Summary

This change adds `write_default_config_if_absent()` to write a static, fully-commented TOML template to `paths.data_dir.join("config.toml")` during `unimatrix version --project-dir`. The write target is derived entirely from `ensure_data_directory()`, which canonicalizes the input path and constructs the target as `~/.unimatrix/{sha256_hex_16}/config.toml`. No user-controlled string is joined into the final write path.

No blocking findings. Details below.

---

### Path Traversal

**Severity: none**

The config write path is constructed as:
```
paths.data_dir.join("config.toml")
```
`data_dir` is `~/.unimatrix/{hash}/` where `{hash}` is the first 16 hex chars of `SHA-256(canonical_project_root)`. The hash is computed over an already-canonicalized path (`detect_project_root` calls `.canonicalize()`), so symlink traversal and `..` sequences are resolved before the hash is computed. The output is restricted to `[0-9a-f]{16}`, making the resulting `data_dir` path injection-safe. The literal `"config.toml"` is hardcoded — no user-supplied filename participates in the join.

---

### OWASP Concerns

| Concern | Assessment |
|---------|-----------|
| Path traversal | Not possible — hash output is hex-only, literal filename hardcoded |
| Injection (shell/SQL/format) | Not applicable — pure filesystem write of a static string |
| Injection (prompt) | `server.instructions` in the template carries the comment `"Must not contain prompt injection patterns"` — the existing `ContentScanner` gate fires on `load_config`, not on write. The written file is all comments, so no injection surface is introduced at write time |
| Deserialization | The written TOML is a static constant (`pub static DEFAULT_CONFIG_TOML`). No deserialization of external input occurs in this change |
| Broken access control | `data_dir` is created with mode `0o700` (unix) by `ensure_data_directory`. The new config write uses `create_dir_all` on the same directory — already created and correctly permissioned before this call |
| Secrets / credentials | None present in `DEFAULT_CONFIG_TOML` or anywhere in the diff |
| New dependencies | None introduced |

---

### `--force` Flag

`--force` triggers `fs::write` (unconditional overwrite) instead of `OpenOptions::create_new`. Both paths write only `DEFAULT_CONFIG_TOML` — a compile-time constant. The only risk is user data loss (overwriting a customized config), which is the documented and expected behavior of the flag. The test `test_write_default_config_overwrites_with_force` covers this.

---

### Error Handling

`write_default_config_if_absent` never propagates errors — all failure paths log a `warn` and return. This is appropriate for a best-effort init step. No panic paths exist. The failure test (`test_write_default_config_succeeds_even_if_write_fails`) covers the read-only directory scenario.

The `if config_path.exists()` check in `handle_version` after the call is cosmetic (for the eprintln). A TOCTOU race here is inconsequential: worst case the path is logged even if deleted immediately after, or not logged even if created. No security impact.

---

### Blast Radius

If `write_default_config_if_absent` has a subtle bug:
- Worst case (force=true): corrupted config.toml in `~/.unimatrix/{hash}/`, causing `load_config` to fail on next server start. Failure is caught at startup with a clear error; no data corruption of the database or vector index.
- Worst case (force=false): if the create-new race window were exploitable, an attacker with local user access could pre-create the file. This would cause the function to silently skip writing (the `AlreadyExists` arm returns without error). Given that the file lives in `~/.unimatrix/` (user-private, mode 0700), this is not a credible threat.

---

### Regression Risk

Low. The only behavioral change to existing code paths is the `Version` command variant gaining a `force: bool` field (default `false`). All callers in `main.rs` are updated. Existing tests updated to `Version { .. }` pattern. `handle_version(None, false)` behavior is unchanged. No async path is touched.

---

No blocking findings. Safe to merge.

PR #639, opened 2026-05-25
fix(infra): daemon child opens log file explicitly, eliminating fd inheritance (#638)

### LINKED ISSUE #638: bug: daemon child early log lines lost when spawned via bridge
## Summary

When the daemon is spawned via the bridge auto-start path (`run_bridge` → `run_daemon_launcher`), the daemon child's early startup log lines (including config provenance from #635) are lost from the log file. The same lines appear correctly when running `serve --foreground` or `serve --daemon` directly from the terminal.

## Observed behavior

Log file (`~/.unimatrix/{hash}/unimatrix.log`) starts at PID guard acquisition:
```
INFO unimatrix_server::infra::pidfile: stale PID file found (process is dead); PidGuard will reclaim pid=67580
INFO unimatrix_server::uds::listener: removed stale socket file ...
```

Missing lines that should precede it:
```
INFO unimatrix: starting unimatrix daemon
INFO unimatrix: daemon project initialized ...
INFO unimatrix: global config not found; using compiled defaults ...
INFO unimatrix: project config loaded ...
INFO unimatrix: config loaded preset=Collaborative
INFO unimatrix: effective config categories=[...] ...
```

## Expected behavior

All daemon startup lines (from `tracing::info!("starting unimatrix daemon")` onward) should appear in the log file regardless of how the daemon was started.

## Reproduction

1. Kill all unimatrix processes
2. Remove the log file
3. Start via bridge: `/usr/local/cargo/bin/unimatrix` (no args, triggers bridge → auto-spawn)
4. Check log file — early lines missing

vs.

1. Kill all unimatrix processes
2. Remove the log file  
3. Start directly: `/usr/local/cargo/bin/unimatrix serve --foreground`
4. Check stderr — all lines present

## Root cause hypothesis

`run_daemon_launcher` (daemon.rs:44) opens the log file in append mode and passes it as the child's stdout/stderr via `std::process::Command`. The child initializes tracing to stderr. There may be a race between when the child starts writing to stderr and when the OS fully sets up the redirected file descriptors, or a buffering issue where early writes are lost before the file is fully wired.

Alternatively, `prepare_daemon_child()` calling `setsid()` may close or reassign file descriptors in some environments.

## Files

- `crates/unimatrix-server/src/infra/daemon.rs:44-95` — `run_daemon_launcher` (file handle setup + child spawn)
- `crates/unimatrix-server/src/infra/daemon.rs:132` — `prepare_daemon_child` (setsid)
- `crates/unimatrix-server/src/main.rs:449-458` — tracing subscriber init in `tokio_main_daemon`

## Related

- #635 / #636 — config load observability (the lines that are being lost)

dug-21:
## Summary

- **Root cause**: `tokio_main_daemon` initialized tracing with `.with_writer(std::io::stderr)`, relying on the launcher's fd 2 redirect to the log file. This implicit inheritance is fragile on macOS where `posix_spawn` + `setsid()` can lose early writes.
- **Fix**: When `cli.daemon_child` is true, the child now opens `paths.log_path` explicitly with `Mutex<LineWriter<File>>` + `.with_ansi(false)`. Falls back to stderr with warning if log file open fails (container safety). Foreground mode unchanged.
- **LineWriter wrapping** ensures each log line is flushed immediately — no buffered lines lost on crash.

## Changes

- `crates/unimatrix-server/src/main.rs` — moved `ensure_data_directory` before tracing init; daemon-child mode opens log file explicitly with fallback
- `crates/unimatrix-server/src/main_tests.rs` — 3 new tests

## Test plan

- [x] `test_daemon_child_writes_early_log_to_file` — spawns real binary, verifies early log lines in file
- [x] `test_daemon_log_writer_line_buffered` — confirms LineWriter flush behavior
- [x] `test_daemon_log_fallback_on_open_failure` — verifies stderr fallback on open failure
- [x] Full workspace: 5192 passed, 0 failed
- [x] Integration smoke: 23 passed
- [x] Integration protocol: 13 passed
- [x] Clippy clean on changed files
- [x] Gate: Bug Fix Validation — PASS

Closes #638

### DIFF
--- a/crates/unimatrix-server/src/main.rs
+++ b/crates/unimatrix-server/src/main.rs
@@ -446,22 +446,60 @@ fn run_stop(project_dir: Option<PathBuf>) -> i32 {
 /// reachable only from the daemon token cancellation path below.
 #[tokio::main]
 async fn tokio_main_daemon(cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
-    // Initialize tracing (daemon: log to stderr, redirected to log file by launcher).
+    // #638: Resolve project paths BEFORE tracing init so daemon-child mode can
+    // open the log file explicitly rather than relying on inherited fd 2.
+    // Foreground mode (cli.daemon_child == false) still uses stderr.
+    let paths = project::ensure_data_directory(cli.project_dir.as_deref(), None)
+        .map_err(|e| ServerError::ProjectInit(e.to_string()))?;
+
+    // Initialize tracing.
     // RUST_LOG override: RUST_LOG=info,unimatrix_server::obs=debug enables UDS observation logs
     // To silence obs logs (default): RUST_LOG unset or RUST_LOG=info
     let default_level = if cli.verbose { "debug" } else { "info" };
     let filter = tracing_subscriber::EnvFilter::try_from_default_env()
         .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new(default_level));
-    tracing_subscriber::fmt()
-        .with_env_filter(filter)
-        .with_writer(std::io::stderr)
-        .init();
+
+    if cli.daemon_child {
+        // #638: Daemon child — open log file explicitly with .with_ansi(false).
+        // This eliminates dependence on the launcher's fd 2 redirect, which is
+        // fragile on macOS with posix_spawn + setsid().
+        // LineWriter ensures each log line is flushed immediately (crash safety).
+        match std::fs::OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(&paths.log_path)
+        {
+            Ok(file) => {
+                let writer = std::sync::Mutex::new(std::io::LineWriter::new(file));
+                tracing_subscriber::fmt()
+                    .with_env_filter(filter)
+                    .with_writer(writer)
+                    .with_ansi(false)
+                    .init();
+            }
+            Err(e) => {
+                // MANDATORY FALLBACK: never exit silently because log file open
+                // failed — critical for container deployments (nan-014 W2-1).
+                eprintln!(
+                    "WARNING: failed to open daemon log at {}: {e}; falling back to stderr",
+                    paths.log_path.display()
+                );
+                tracing_subscriber::fmt()
+                    .with_env_filter(filter)
+                    .with_writer(std::io::stderr)
+                    .with_ansi(false)
+                    .init();
+            }
+        }
+    } else {
+        // Foreground mode (serve --foreground) — log to stderr unchanged.
+        tracing_subscriber::fmt()
+            .with_env_filter(filter)
+            .with_writer(std::io::stderr)
+            .init();
+    }
 
     tracing::info!("starting unimatrix daemon");
-
-    // Initialize project paths.
-    let paths = project::ensure_data_directory(cli.project_dir.as_deref(), None)
-        .map_err(|e| ServerError::ProjectInit(e.to_string()))?;
 
     tracing::info!(
         project_root = %paths.project_root.display(),
--- a/crates/unimatrix-server/src/main_tests.rs
+++ b/crates/unimatrix-server/src/main_tests.rs
@@ -999,3 +999,193 @@ fn test_foreground_appears_in_serve_help() {
         panic!("`serve` subcommand not found in CLI");
     }
 }
+
+// ---------------------------------------------------------------------------
+// #638: Daemon log file writer tests
+// ---------------------------------------------------------------------------
+
+/// #638: Daemon child process writes early startup log lines to the log file,
+/// not just to inherited stderr. This integration test spawns the real binary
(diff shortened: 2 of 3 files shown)


dug-21 (commented):
## Security Review: bugfix-638-security-reviewer

### Risk Level: LOW

### Summary

This fix is minimal, well-scoped, and introduces no new security concerns. The change moves `ensure_data_directory()` before tracing initialization and opens the log file explicitly in daemon-child mode instead of relying on inherited fd 2. The file path is internally computed (not user-controlled beyond `--project-dir`), the data directory has 0700 permissions, and the fallback path is safe.

---

### OWASP / Security Checklist

| Check | Result |
|-------|--------|
| **Path traversal** | PASS — `paths.log_path` is computed internally via `data_dir.join("unimatrix.log")` where `data_dir` is `~/.unimatrix/{sha256-hash}/`. The path is not directly influenced by untrusted input. `--project-dir` goes through `canonicalize()` before hashing, which resolves symlinks and `..` components. |
| **Symlink attacks** | LOW RISK — The data directory is created with `0700` permissions (user-only). An attacker with write access to `~/.unimatrix/{hash}/` already has local user access, which is outside the threat model for a local-only daemon. The `OpenOptions::append(true).create(true)` call follows the existing pattern used by the launcher (daemon.rs:46-55). |
| **Injection** | PASS — No shell commands, SQL, or format strings with untrusted input. The `eprintln!` in the fallback uses `Display` formatting, not string interpolation of untrusted content. |
| **Input validation** | PASS — No new external inputs. The `--daemon-child` flag is a hidden boolean set only by the launcher. The `--project-dir` validation is unchanged. |
| **Error handling** | PASS — Log file open failure falls back to stderr with a warning (no silent failure, no panic). `ensure_data_directory` failure returns an error (process exits with error message). |
| **Access control** | PASS — Data directory permissions (0700) are set by `ensure_data_directory` and are unchanged by this PR. |
| **Secrets** | PASS — No hardcoded secrets, API keys, or credentials in the diff. |
| **Dependencies** | PASS — No new dependencies added (confirmed: no Cargo.toml/Cargo.lock changes). |
| **Deserialization** | N/A — No deserialization of untrusted data. |

---

### Findings

#### 1. Pre-tracing error path is unlogged (informational, non-blocking)
- **Severity**: Low (informational)
- **Location**: `main.rs:452-453`
- **Description**: If `ensure_data_directory()` fails before tracing is initialized, the error propagates as a `Box<dyn Error>` back to `main()`. In daemon-child mode, there is no tracing subscriber active yet, so this error only reaches the caller's error handler (which prints to stderr). This is acceptable because: (a) the launcher still has the log file open on fd 2, so the child's stderr will land in the log; and (b) `ensure_data_directory` succeeded once already in the launcher path (`compute_paths_sync` at main.rs:348), so a second failure here would indicate a transient filesystem issue.
- **Recommendation**: No action needed. The existing error propagation is sufficient.
- **Blocking**: No

#### 2. Launcher still opens log file for fd 2 redirect (defense in depth, positive)
- **Severity**: Informational
- **Location**: `daemon.rs:46-55` (unchanged)
- **Description**: The launcher continues to redirect the child's stdout/stderr to the log file. The child now also opens the same file independently for tracing. This is intentional defense-in-depth: any output that bypasses the tracing subscriber (e.g., panic hook output, eprintln fallback messages) still reaches the log file via the inherited fd 2. This is good design.
- **Blocking**: No

---

### Blast Radius Assessment

**Worst case**: If the log file open succeeds but the `Mutex<LineWriter<File>>` writer has a subtle bug (e.g., mutex poisoning from a panic during write), subsequent log lines would be lost. The daemon would continue operating normally — only logging would be affected. This is a safe failure mode (no data corruption, no service disruption). The `LineWriter` flush-on-newline behavior is well-tested in stdlib and verified by the new unit test.

**Scope of change**: Only `tokio_main_daemon` is modified. The stdio mode (`tokio_main_stdio`) and bridge mode (`tokio_main_bridge`) are completely unaffected. The foreground mode path through `tokio_main_daemon` (when `cli.daemon_child == false`) preserves the original stderr-only behavior.

### Regression Risk

- **Foreground mode**: Unchanged — the `else` branch at line 494 preserves the exact original behavior (stderr writer with ANSI).
- **Stdio mode**: Completely untouched.
- **Bridge mode**: Completely untouched.
- **Daemon-child mode**: This is the only affected path. The new behavior is strictly better (explicit file open vs. fragile fd inheritance).
- **Order of operations**: `ensure_data_directory` is now called before tracing init instead of after. This means directory creation and permission setting happen slightly earlier in the startup sequence. This is safe — the same function was already called successfully by the launcher before spawning the child.

### Verdict

APPROVE — no blocking findings. Minimal, well-targeted fix. No security concerns. Good fallback behavior. Tests cover the happy path, line-buffering behavior, and open-failure fallback.

---
*Security review by bugfix-638-security-reviewer*

PR #642, opened 2026-05-26
fix(import): replace manual MAX(event_id)+1 with log_audit_event in record_provenance (#633)

### LINKED ISSUE #633: bug: audit_log event_id UNIQUE collisions after export/import — events silently dropped
## Summary
After export/import, the daemon logs repeated WARN: `audit_fire_and_forget: write failed: UNIQUE constraint failed: audit_log.event_id`. 18 audit writes were silently dropped in a single session.

## Context
- Audit log append-only completeness is a non-negotiable (PRODUCT-VISION.md, security cross-cutting concern #2)
- The import likely brings over `audit_log` rows with existing `event_id` values
- The daemon's event_id generator does not re-seed past the imported max, causing collisions on new writes
- Events are written fire-and-forget — collisions are logged as WARN but the audit event is permanently lost

## Scope
1. Confirm event_id generation strategy (UUID? sequence? timestamp-based?)
2. Determine whether import carries over audit_log rows (and whether it should)
3. If import does carry audit_log: ensure the event_id generator seeds past `MAX(event_id)` on startup
4. If import does not carry audit_log: determine why collisions occur — possibly stale WAL or cached state
5. Consider whether `INSERT OR IGNORE` is acceptable for audit (it should not be — audit gaps are a compliance risk)

## Why this matters
The audit log is the tamper-evident integrity layer. Silent event drops violate the append-only guarantee and could mask unauthorized operations. DDL triggers protect against UPDATE/DELETE but not against failed INSERTs that are silently swallowed.

dug-21:
## Summary

- **Root cause**: `record_provenance()` wrote audit_log entries using `SELECT COALESCE(MAX(event_id), 0) + 1`, bypassing the `next_audit_id` counter. After import, the counter held N but audit_log contained event_id N+1 (provenance record). Every subsequent `log_audit_event()` returned N+1, collided permanently — the transaction rolled back the counter increment, so it never advanced. All future audit writes were silently dropped.
- **Fix**: Replaced manual SQL with `store.log_audit_event()`, which uses the atomic `UPDATE...RETURNING` counter pattern. Also fixes a latent 8-of-12 column gap and incorrect `Outcome::Denied` (now `Outcome::Success`).
- **Tests**: Two targeted regression tests verify post-import `log_audit_event()` succeeds with monotonic IDs, and that provenance uses the counter (not MAX+1).

## Test plan

- [x] `test_gh633_log_audit_event_succeeds_after_import` — imports 3 audit rows, calls `log_audit_event()` twice, asserts monotonic IDs > all imported rows + provenance
- [x] `test_gh633_provenance_uses_counter_not_max_plus_one` — imports with counter=5, max(event_id)=2, verifies provenance at event_id 6 (counter-based, not MAX+1=3)
- [x] Workspace tests: 5194 passed, 0 failed
- [x] Clippy: no new warnings
- [x] Integration suites: 267 passed, 0 failures
- [x] Gate 3: PASS (7/7 checks)

Closes #633

### DIFF
--- a/crates/unimatrix-server/src/import/mod.rs
+++ b/crates/unimatrix-server/src/import/mod.rs
@@ -16,12 +16,11 @@ use std::fs::File;
 use std::io::{self, BufRead, BufReader};
 use std::path::Path;
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
 
 use sqlx::Row;
 use sqlx::SqlitePool;
 use sqlx::sqlite::SqliteConnection;
-use unimatrix_store::{SqlxStore, compute_content_hash};
+use unimatrix_store::{AuditEvent, Outcome, SqlxStore, compute_content_hash};
 
 use crate::format::{ExportHeader, ExportRow};
 use crate::project;
@@ -223,7 +222,7 @@ async fn run_import_async(
     crate::embed_reconstruct::reconstruct_embeddings(&store, &paths.vector_dir)?;
 
     // Phase 11: Record provenance
-    record_provenance(pool, input, &counts).await?;
+    record_provenance(&store, input, &counts).await?;
 
     // Phase 12: Summary
     print_summary(&counts, skip_hash_validation);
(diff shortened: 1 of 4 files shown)


dug-21:
## Security Review — LOW risk

**Reviewer**: bugfix-633-security-reviewer (inline — subagent timeouts)

### Findings (0 blocking)

1. **SQL injection surface reduced** — removes raw `SELECT COALESCE(MAX(event_id), 0) + 1` and routes through parameterized `log_audit_event()`. Strictly safer.
2. **No new trust boundaries** — same write access, proper API channel.
3. **No new dependencies, no secrets** in diff.
4. **Column coverage improved** — 8→12 columns, plus `Outcome::Denied`→`Outcome::Success` correction.
5. **Blast radius: minimal** — one call per import, failure logged as error.
6. **Test changes are test-only** — fix the same counter-bypass class in test helpers.

**Recommendation**: APPROVE — low risk, minimal change, eliminates a root cause class.

PR #654, opened 2026-05-29
fix: embedding model SHA-256 hash verification (#651)

### LINKED ISSUE #651: bug: embedding model ensure_model() does not verify SHA-256 hash
## Summary

`ensure_model()` for the embedding model does not verify SHA-256 hashes after download, while `ensure_nli_model()` has optional hash verification via `nli_model_sha256` config. This is a supply chain integrity gap — a replaced or corrupted embedding model file would be loaded without detection.

## Impact

An attacker or corrupted download could replace the embedding ONNX model file between runs. The binary would load it without verification, silently degrading or manipulating all vector search results. This contradicts ADR-002's SHA-256 hash-pinning requirement for model files.

## Fix

Add `embedding_model_sha256` config option (parallel to `nli_model_sha256`). Verify hash after download in `ensure_model()` and on load. Refuse to start with a hash mismatch and a clear error message.

## Origin

Discovered in ASS-061 Q4 (Model Distribution) out-of-scope finding #1.

dug-21:
## Summary

- Adds `embedding_model_sha256` config field to `InferenceConfig`, mirroring the existing `nli_model_sha256` pattern
- Extracts `verify_sha256()` from `nli_handle.rs` into shared `infra/hash.rs` module, used by both NLI and embedding handles
- Enforces verify-before-load ordering: `ensure_model()` → `verify_sha256()` → `OnnxProvider::new()` — hash is checked before the ONNX session loads into memory
- Global-wins merge semantics: per-project config cannot override a global hash pin (security-critical field)
- Adds SHA-256 output for embedding model in `unimatrix model-download` CLI
- Adds import path warning in `embed_reconstruct.rs` when hash verification is unavailable

## Root Cause

crt-023 added SHA-256 hash verification for the NLI model but never back-ported it to the embedding model loading path. A replaced or corrupted embedding ONNX model file would be loaded without detection.

## Test Plan

- [x] 4 hash verification unit tests (correct, wrong, missing file, case-insensitive)
- [x] 5 config validation tests (wrong length, 63 chars, non-hex, valid 64 hex, None)
- [x] 4 global-wins merge tests (global wins, project fallback, both None, same value)
- [x] Full workspace: 5293 tests passed, 0 failed
- [x] Integration: 301 tests passed, 0 failed
- [x] Clippy: no new warnings

Closes #651

### DIFF
--- a/crates/unimatrix-server/src/embed_reconstruct.rs
+++ b/crates/unimatrix-server/src/embed_reconstruct.rs
@@ -39,6 +39,14 @@ pub fn reconstruct_embeddings(
     vector_dir: &Path,
 ) -> Result<(), Box<dyn std::error::Error>> {
     // Step 1: Initialize ONNX provider
+    // NOTE (bugfix-651): The import path does not have access to InferenceConfig,
+    // so embedding_model_sha256 hash verification is not available here. The import
+    // path operates on the CLI sync path (no server config loaded). Log a warning
+    // so operators know hash verification was skipped.
+    eprintln!(
+        "WARNING: embedding model hash verification is not available on the import path. \
+         Verify model integrity manually if importing into a production instance."
+    );
     eprintln!("Initializing embedding model...");
     let embed_config = EmbedConfig::default();
     let provider = OnnxProvider::new(embed_config).map_err(|e| {
--- a/crates/unimatrix-server/src/eval/profile/layer.rs
+++ b/crates/unimatrix-server/src/eval/profile/layer.rs
@@ -250,7 +250,14 @@ impl EvalServiceLayer {
         // ----------------------------------------------------------------
         let embed_handle = EmbedServiceHandle::new();
         let embed_config = EmbedConfig::default();
-        embed_handle.start_loading(embed_config);
+        embed_handle.start_loading(
+            embed_config,
+            profile
+                .config_overrides
+                .inference
+                .embedding_model_sha256
+                .clone(),
+        );
 
         // ----------------------------------------------------------------
         // Step 6b: Build NLI handle for NLI-enabled profiles (crt-023, FR-26, ADR-006).
--- a/crates/unimatrix-server/src/infra/config.rs
+++ b/crates/unimatrix-server/src/infra/config.rs
@@ -261,6 +261,20 @@ pub struct InferenceConfig {
     pub rayon_pool_size: usize,
 
     // -----------------------------------------------------------------------
+    // Embedding model hash verification (bugfix-651)
+    // -----------------------------------------------------------------------
+    /// SHA-256 hash of the embedding model ONNX file as a 64-char lowercase hex string.
+    ///
+    /// When set, the hash is verified before the ONNX session is loaded into memory.
+    /// When `None`, hash verification is skipped with a `warn`-level log.
+    /// Must be exactly 64 hex characters if set.
+    ///
+    /// Security-critical: global-wins semantics in merge. A per-project config
+    /// MUST NOT override a global hash pin.
+    #[serde(default)]
+    pub embedding_model_sha256: Option<String>,
+
+    // -----------------------------------------------------------------------
     // NLI cross-encoder fields (crt-023)
     // -----------------------------------------------------------------------
     /// Enable the NLI cross-encoder (default `false`, opt-in).
@@ -702,6 +716,7 @@ impl Default for InferenceConfig {
         // On 20-core:    num_cpus = 20; 20/2 = 10; max(10, 4) = 10; min(10, 8) = 8.
         InferenceConfig {
             rayon_pool_size: (num_cpus::get() / 2).max(4).min(8),
+            embedding_model_sha256: None,
             nli_enabled: default_nli_enabled(),
             nli_model_name: None,
             nli_model_path: None,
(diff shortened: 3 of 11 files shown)


dug-21 (commented):
## Security Review: bugfix-651-security-reviewer

**Risk Level: Low**

Overall this is a well-structured, minimal change that closes a legitimate supply chain gap. The embedding model loading path now has hash verification parity with the NLI model path. The shared `verify_sha256` extraction is clean. No blocking findings.

---

### Findings

#### 1. TOCTOU Window Between Hash Verification and ONNX Session Load
- **Severity**: Low
- **Location**: `crates/unimatrix-server/src/infra/embed_handle.rs:107-132`
- **Description**: There is a time-of-check-to-time-of-use gap between `verify_sha256()` (step 2) and `OnnxProvider::new()` -> `Session::builder().commit_from_file()` (step 3). A local attacker with write access to the cache directory could swap the model file between verification and load.
- **Blocking**: No
- **Rationale**: This is the same pattern used by the pre-existing NLI code path. Mitigating would require reading the file once, verifying the hash on that buffer, and constructing the ONNX session from the same buffer — which would require changes to the `ort` crate's `Session::builder` API. The threat model here assumes the cache directory is not writable by untrusted processes, which is reasonable for a local-first tool. Documenting this accepted risk is sufficient.

#### 2. Validation Accepts Uppercase Hex But Error Message Says "lowercase"
- **Severity**: Low (UX, not security)
- **Location**: `crates/unimatrix-server/src/infra/config.rs:1043-1049`
- **Description**: `is_ascii_hexdigit()` accepts both uppercase (`A-F`) and lowercase (`a-f`) characters. The `verify_sha256()` comparison uses `eq_ignore_ascii_case`. So uppercase hashes are valid at runtime. However, the validation error message says "must be a 64-character lowercase hex string" — this is misleading. A user who provides uppercase hex will pass validation and work correctly, but if they provide a 63-char string, the error message incorrectly implies lowercase is required.
- **Blocking**: No
- **Recommendation**: Change the error message to "must be a 64-character hex string" (drop "lowercase"), or enforce lowercase by checking `c.is_ascii_lowercase() || c.is_ascii_digit()`. The NLI field has the same pattern, so this is pre-existing.

#### 3. NLI Hash Output Format Is a Breaking Change to stdout
- **Severity**: Low
- **Location**: `crates/unimatrix-server/src/main.rs:1506`
- **Description**: The `model-download --nli` command previously printed the bare NLI hash to stdout (`println!("{}", hash_hex)`). It now prints `nli: {hash_hex}`. Any operator scripts that pipe `model-download` stdout to capture the hash (e.g., `unimatrix model-download --nli | head -1`) will now get `embedding: {hash}` on line 1 and `nli: {hash}` on line 2 instead of the bare hash. This is functionally correct but is a stdout contract change.
- **Blocking**: No
- **Rationale**: This is a CLI tool run interactively by operators. The guidance text already went to stderr. The new prefixed format is strictly better for disambiguation. But anyone with existing automation should be aware.

#### 4. Import Path Skips Hash Verification (Documented, Accepted)
- **Severity**: Low
- **Location**: `crates/unimatrix-server/src/embed_reconstruct.rs:42-49`
- **Description**: The import/reconstruct path logs a WARNING but proceeds without hash verification. This is documented in the code comment as intentional — the import CLI path has no `InferenceConfig` available.
- **Blocking**: No
- **Rationale**: The import path is a CLI operation run by the operator. The warning is appropriate. The alternative (threading `InferenceConfig` through the import pipeline) would be a larger change with its own blast radius.

#### 5. NLI Hash Uses Project-Wins Merge; Embedding Hash Uses Global-Wins
- **Severity**: Low (design observation)
- **Location**: `crates/unimatrix-server/src/infra/config.rs:2748-2764` vs `2779-2782`
- **Description**: The new `embedding_model_sha256` correctly uses global-wins semantics (a per-project config cannot override a global hash pin). However, the pre-existing `nli_model_sha256` uses project-wins (`.or()`) semantics, meaning a per-project config CAN override the global NLI hash. This is an asymmetry — if the security reasoning for global-wins applies to the embedding model, it likely also applies to the NLI model.
- **Blocking**: No
- **Rationale**: The NLI merge logic is pre-existing and not part of this PR's scope. But this is worth a follow-up issue to align the NLI hash merge to global-wins as well.

---

### OWASP Assessment
- **Injection**: No risk. No shell commands, SQL, or format strings with untrusted input.
- **Path Traversal**: No risk. `onnx_filename()` returns a `&'static str` (`"model.onnx"`). `model_dir` comes from `ensure_model()` which uses `cache_subdir()` (slash-replaced model ID). No user-controlled path components.
- **Deserialization**: ONNX session construction from file is pre-existing. Hash verification now gates this, which is strictly an improvement.
- **Access Control**: Global-wins merge prevents per-project config from bypassing a hash pin. Correct.
- **Secrets**: No hardcoded secrets, API keys, or credentials in the diff.
- **Dependencies**: No new dependencies added. `sha2` was already in use.

### Blast Radius Assessment
- **Worst case if verify_sha256 has a subtle bug**: Model loads without verification (if the function incorrectly returns `Ok(())`). This is the same state as before this PR, so the blast radius is "no improvement" rather than "regression."
- **Worst case if config validation has a bug**: A malformed hash string passes validation but fails at runtime during `eq_ignore_ascii_case` comparison, causing the embedding model to enter `Failed` state. The retry monitor would re-attempt 3 times, then give up. Search would be unavailable but the server would remain operational for non-search operations.
- **Worst case if merge logic has a bug**: A per-project hash overrides a global pin. This would be a security regression but requires a specific multi-config deployment. Tests cover this case (4 merge tests).

### Regression Risk
- **Low**. The `start_loading()` signature change is the primary API break. All call sites (main.rs daemon, main.rs stdio, eval profile layer, test_support) are updated. The `None` default in test_support preserves existing test behavior.
- The formatting changes in `export_integration.rs` and `import_integration.rs` are whitespace-only (rustfmt). No logic changes.

### Verdict
**Approve.** The change is minimal, well-tested (5 validation tests, 4 merge tests, 4 hash verification tests), and closes a real supply chain gap. No blocking findings.

PR #664, opened 2026-05-29
fix(token): eliminate TOCTOU races in token file creation (#662)

### LINKED ISSUE #662: [vnc-021] TOCTOU race in token file creation
## Problem

`src/http/token.rs:44-59` uses `File::create()` then `set_permissions(0o600)`, leaving a brief window where the token file exists with umask-default permissions. The file is empty during this window, but the fix is trivial.

## Fix

Replace with `OpenOptions::new().mode(0o600).create_new(true)` — this atomically creates with correct permissions AND prevents the overwrite race (two servers starting simultaneously in the same data volume).

## Context

Found during vnc-021 security review (PR #661, Finding 1). Non-blocking — window is short and file is empty during it.

## Scope

Fast follow-up PR. Single file change.

dug-21:
## Summary

- Replaces `File::create()` + `set_permissions(0o600)` with `OpenOptions::new().create_new(true).mode(0o600)` for atomic file creation with correct permissions (`O_CREAT | O_EXCL`)
- Inverts `load_or_generate_token()` flow to generate-first with `AlreadyExists` fallback, eliminating the `path.exists()` TOCTOU race
- Adds cleanup guard: best-effort `fs::remove_file` on `write_all` failure to prevent 0-byte token files

Fixes #662

## Test plan

- [x] `test_no_overwrite_existing_token` — proves `create_new` rejects when file exists, loads existing token
- [x] `test_permissions_at_creation_are_0600` — verifies 0600 permissions set atomically at creation
- [x] `test_concurrent_creation_no_corruption` — two threads race to create, both succeed with same token
- [x] All 16 token tests pass (13 existing + 3 new)
- [x] Full workspace: 5416 passed, 0 failed
- [x] Integration smoke + security + protocol suites: 56 passed, 0 failed
- [x] Clippy clean on changed file

### DIFF
--- a/crates/unimatrix-server/src/http/token.rs
+++ b/crates/unimatrix-server/src/http/token.rs
@@ -3,6 +3,9 @@
 //! Manages the bearer token lifecycle: generate on first run, load on subsequent
 //! runs, validate format. The token is a 32-byte cryptographic random value stored
 //! as 64 hex characters in `{data_dir}/token` with mode 0600.
+//!
+//! File creation uses `O_CREAT | O_EXCL` (via `create_new`) with mode 0600 to
+//! atomically create-with-permissions and reject concurrent creators.
 
 use std::fs;
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review -- Agent 662-security-reviewer

**Risk Level: LOW**

### Summary

The TOCTOU fix is correct and well-implemented. The change replaces `path.exists()` + `File::create()` with `OpenOptions::create_new(true).mode(0o600)`, which maps to `O_CREAT | O_EXCL` at the kernel level. This is the canonical solution for this class of race condition. The permission-at-creation approach via `.mode(0o600)` eliminates the chmod window from the prior `set_permissions()` call. No new dependencies, no secrets in the diff, no injection vectors.

### Findings

#### 1. String-based error discrimination in `is_already_exists`
- **Severity**: Low
- **Location**: `crates/unimatrix-server/src/http/token.rs:44-52`
- **Description**: The `is_already_exists()` function matches `AlreadyExists` by searching for substrings `"(os error 17)"`, `"already exists"`, and `"File exists"` within a `ServerError::ProjectInit(String)`. This is fragile -- it depends on the OS locale and Rust's `std::io::Error` Display format, both of which could change. If a future Rust version changes the error message format, or the server runs in a non-English locale, this function could fail to recognize `EEXIST`, causing the function to return an opaque error instead of falling back to `load_existing_token`. The failure mode is safe (server fails to start, does not silently corrupt data), but it would be a confusing diagnostic.
- **Recommendation**: Consider preserving the `io::Error` as a structured variant in `ServerError` (e.g., `ServerError::TokenFileCreate(io::Error)`) so you can match on `e.kind() == ErrorKind::AlreadyExists` instead of string parsing. This is a hardening suggestion, not a blocking concern -- the current approach works correctly on all standard Linux/macOS platforms.
- **Blocking**: No

#### 2. Partial-write race window between `create_new` and `write_all`
- **Severity**: Low
- **Location**: `crates/unimatrix-server/src/http/token.rs:86-101`
- **Description**: There is a narrow window between `create_new` succeeding (file exists, 0 bytes, 0600 perms) and `write_all` completing (file has 64 bytes of hex). If a concurrent process hits the `AlreadyExists` fallback during this window, `load_existing_token` will read a 0-byte file and fail validation (`validate_token_format` rejects length != 64). The failure mode is safe -- the concurrent caller gets a `ServerError`, not a corrupt or empty token. The cleanup logic (`remove_file` on write failure) is also correct. In practice, this window is sub-microsecond and the function is called once at server startup, making concurrent invocation extremely unlikely.
- **Recommendation**: Acceptable as-is for the startup-only call site. If this function were ever called in a hot path, consider write-to-temp + atomic rename. But for server init, the current approach is sound.
- **Blocking**: No

#### 3. Non-Unix path: no permission enforcement on Windows
- **Severity**: Low (informational)
- **Location**: `crates/unimatrix-server/src/http/token.rs:80-83`
- **Description**: The `#[cfg(not(unix))]` branch uses `create_new(true)` without `.mode()`, so on Windows the token file will have default ACLs. Since the project targets Linux/macOS, this is informational only.
- **Blocking**: No

### OWASP Assessment
- **Injection**: No untrusted input flows into file paths or shell commands.
- **Access Control**: Token file permissions (0600) correctly restrict read access. Atomic creation eliminates the permission escalation window.
- **Deserialization**: Token loading uses `fs::read_to_string` + hex decode with strict format validation.
- **Input Validation**: `validate_token_format` is preserved and unchanged -- 64 hex chars, strict.
- **Secrets**: No hardcoded tokens or credentials. Token generated from CSPRNG.

### Blast Radius Assessment
- **Worst case**: If `is_already_exists` fails to match, the server fails to start with an error. Safe failure mode.
- **Affected components**: Only the HTTP bearer auth startup path. Stdio/UDS transports unaffected.

### Regression Risk
Low. Public API unchanged. Three new tests (T-TM-12, T-TM-13, T-TM-14) cover the fix.

### Verdict
The fix is minimal, correctly scoped, and addresses the reported TOCTOU race. No blocking findings. Recommending approval.

PR #668, opened 2026-05-29
fix(router): enforce body size limit on stream, not just Content-Length header (#663)

### LINKED ISSUE #663: [vnc-021] Chunked transfer encoding bypasses body size limit
## Problem

`src/http/router.rs:292-297` checks `Content-Length` header only for the 1MB body size limit. Chunked transfer encoding bypasses this check entirely. A determined attacker can push ~32 * (hyper read buffer limit) into memory via the 32-connection semaphore slots.

## Fix

Wrap the request body with `http_body::Limited` before passing to rmcp's `StreamableHttpService`. This enforces the size limit regardless of transfer encoding, cutting off the body stream once the limit is reached.

## Context

Found during vnc-021 security review (PR #661, Finding 2). The 32-connection semaphore bounds blast radius, but this is the one real gap in the HTTP transport security model. Shouldn't wait long.

## Scope

Single file change in `router.rs` — wrap body in `http_body::Limited::new(body, MAX_BODY_BYTES)` in the MCP dispatch path.

dug-21:
## Summary

- Chunked transfer encoding bypassed the 1MB body size limit because `McpAdapter::handle` only checked the `Content-Length` header — which chunked TE omits entirely
- Added stream-level `http_body_util::Limited` enforcement that pre-consumes the body with a size cap before passing to rmcp, returning **413 Payload Too Large** on limit exceeded
- Kept the Content-Length header check as a zero-cost fast-path (rejects without reading body bytes)
- Removed the `ReqBody` generic from `McpAdapter` — it now always passes `Full<Bytes>` to rmcp, simplifying type signatures
- Proper error discrimination: `LengthLimitError` → 413, other body errors (client disconnect) → 500

Closes #663

## Test plan

- [x] `test_chunked_body_under_limit_returns_200` — chunked body under limit passes through
- [x] `test_chunked_body_over_limit_returns_413` — Limited catches oversized chunked body
- [x] `test_content_length_over_limit_fast_path_413` — header fast-path still works
- [x] `test_midstream_body_error_returns_500_not_413` — error downcasting is correct
- [x] `test_internal_error_response_format` — response format verification
- [x] Full workspace: 5,421 unit tests passed
- [x] Integration security suite: 20/20 passed
- [x] Clippy clean (pre-existing warnings only)

### DIFF
--- a/crates/unimatrix-server/src/http/router.rs
+++ b/crates/unimatrix-server/src/http/router.rs
@@ -18,7 +18,7 @@ use bytes::Bytes;
 use http::{Method, Request, Response, StatusCode};
 use http_body::Body;
 use http_body_util::combinators::BoxBody;
-use http_body_util::{BodyExt, Full};
+use http_body_util::{BodyExt, Full, Limited};
 use rmcp::transport::streamable_http_server::session::local::LocalSessionManager;
 use rmcp::transport::{StreamableHttpServerConfig, StreamableHttpService};
 use tower::Service;
@@ -45,7 +45,7 @@ pub struct PathRouter<ReqBody>
 where
     ReqBody: Body + Send + 'static,
     ReqBody::Data: Send + 'static,
-    ReqBody::Error: std::fmt::Display,
+    ReqBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
 {
     project_router: ProjectRouter<ReqBody>,
 }
@@ -54,7 +54,7 @@ impl<ReqBody> std::fmt::Debug for PathRouter<ReqBody>
 where
     ReqBody: Body + Send + 'static,
     ReqBody::Data: Send + 'static,
-    ReqBody::Error: std::fmt::Display,
+    ReqBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("PathRouter")
@@ -67,7 +67,7 @@ impl<ReqBody> PathRouter<ReqBody>
 where
     ReqBody: Body + Send + 'static,
     ReqBody::Data: Send + 'static,
-    ReqBody::Error: std::fmt::Display,
+    ReqBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
 {
     /// Create a new PathRouter wrapping a ProjectRouter.
     pub fn new(project_router: ProjectRouter<ReqBody>) -> Self {
@@ -79,7 +79,7 @@ impl<ReqBody> Clone for PathRouter<ReqBody>
 where
     ReqBody: Body + Send + 'static,
     ReqBody::Data: Send + 'static,
-    ReqBody::Error: std::fmt::Display,
+    ReqBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
 {
     fn clone(&self) -> Self {
         PathRouter {
@@ -92,7 +92,7 @@ impl<ReqBody> Service<Request<ReqBody>> for PathRouter<ReqBody>
 where
     ReqBody: Body + Send + 'static,
     ReqBody::Data: Send + 'static,
-    ReqBody::Error: std::fmt::Display,
+    ReqBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
 {
     type Response = Response<BoxBody<Bytes, Infallible>>;
     type Error = Infallible;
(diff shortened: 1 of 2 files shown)


dug-21 (commented):
## Security Review: vnc-021-security-reviewer

**Risk Level: LOW** | No blocking findings. | **Recommendation: APPROVE**

### Summary

This fix correctly addresses GH #663 — chunked transfer encoding bypassing the 1MB body size limit. The two-layer enforcement strategy (Content-Length fast-path + `http_body_util::Limited` stream-level check) is sound and well-tested. Error discrimination via `downcast_ref::<LengthLimitError>()` is verified correct against `http-body-util` 0.1.3 source.

### Findings

#### F-01: Error downcasting is correct (verified)
- **Severity**: info
- **Location**: `router.rs:318-321`
- **Description**: Verified against `http-body-util` 0.1.3 source (`limited.rs:52`): `LengthLimitError` is boxed via `.into()` into `Box<dyn Error + Send + Sync>`, and `downcast_ref::<LengthLimitError>()` correctly identifies it. The upstream crate's own tests use the same pattern (line 139). Sound.

#### F-02: Trait bound change from `Display` to `Into<Box<dyn Error + Send + Sync>>` (non-blocking)
- **Severity**: low
- **Location**: `router.rs` (multiple generic bounds)
- **Description**: `ReqBody::Error` bound changed from `std::fmt::Display` to `Into<Box<dyn std::error::Error + Send + Sync>>`. This is required by `Limited<B>` (which has `B::Error: Into<Box<dyn Error + Send + Sync>>`). The sole production consumer (`hyper::body::Incoming`) has `hyper::Error` which satisfies both bounds. No API breakage.
- **Blocking**: no

#### F-03: Full body buffering before rmcp (non-blocking, advisory)
- **Severity**: low
- **Location**: `router.rs:310-331`
- **Description**: The body is now fully collected into memory (`collected.to_bytes()`) before passing to rmcp. This is bounded by `max_body_bytes` (default 1MB), so the worst case memory allocation per request is 1MB. With the existing 32-connection semaphore, maximum memory for in-flight bodies is 32MB. This is acceptable and actually *safer* than the previous design where rmcp could buffer unbounded chunked bodies internally.
- **Blocking**: no

#### F-04: No new dependencies (verified)
- **Severity**: info
- **Location**: N/A
- **Description**: `http_body_util::Limited` was already a transitive dependency at 0.1.3. The `Limited` import is the only new symbol used. No Cargo.toml changes, no new crate additions.

#### F-05: Unrelated report files included in PR (non-blocking)
- **Severity**: low
- **Location**: `product/features/vnc-021/agents/662-agent-1-fix-v2-report.md`, `vnc-021-docs-report.md`
- **Description**: Three agent report markdown files are included that are unrelated to GH #663 (they document the #662 TOCTOU fix and docs work). These are documentation-only files with no code impact. Harmless but noted per minimal-change principle.
- **Blocking**: no

### OWASP Assessment

| Check | Result |
|-------|--------|
| Input validation | **Pass** — body size enforced at stream level regardless of transfer encoding |
| Injection | **N/A** — no shell/SQL/format string operations in changed code |
| Deserialization | **Pass** — body collected as raw bytes, JSON deserialization handled by rmcp downstream with size-bounded input |
| Access control | **N/A** — auth layer (StaticTokenAuth) unchanged, upstream of this code |
| Error handling | **Pass** — `LengthLimitError` correctly distinguished from transport errors; 413 vs 500 separation is correct |
| Secrets | **Pass** — no credentials, tokens, or keys in diff |
| Dependencies | **Pass** — no new dependencies; `http-body-util` 0.1.3 is current |

### Blast Radius Assessment

**Worst case if fix has a subtle bug**: MCP requests that previously worked could fail with 413 or 500. The blast radius is limited to the HTTP transport path only (UDS/stdio MCP transport is unaffected). Failure mode is safe — the server returns an error response rather than silently corrupting data. The `McpAdapter` is behind auth, so only authenticated clients are affected.

**Memory**: Worst case is 32 concurrent requests each buffering 1MB = 32MB. This is bounded and acceptable.

### Regression Risk

**Low**. The `McpAdapter` generic removal (`McpAdapter<ReqBody>` to `McpAdapter`) is a simplification that moves the generic to the `handle` method. This preserves the same behavior while making the struct easier to clone and debug. All 22 router tests pass. The trait bound change is compatible with the sole production body type (`hyper::body::Incoming`).

### Verdict

**APPROVE recommended**. The fix is minimal, correctly targeted, and well-tested. The two-layer enforcement strategy follows the design reviewer's revised approach (F-02 from 663-design-reviewer-report). Error discrimination is verified correct against the upstream crate source. No blocking findings.

PR #711, opened 2026-06-08
fix: harden cargo test convention to stop orphaned process trees (#122)

### LINKED ISSUE #122: [infra-002] Zombie cargo test processes from subagents block workspace-wide testing
## Problem

Workspace-wide tests (`cargo test --workspace`) hang or fail during feature implementation because **zombie cargo processes from previous subagent sessions accumulate and hold file locks**.

This has been a recurring problem across the last 2-3 feature releases.

## Root Cause

When the implementation scrum-master spawns subagents (validators, testers, rust-dev), each may run `cargo test --workspace`. When the subagent completes or is interrupted:

1. The subagent's shell process exits
2. But the `cargo test` child process **survives as an orphan** (reparented to PID 1)
3. These processes hold:
   - Cargo build directory locks (`target/.cargo-lock`)
   - File handles on `.db` files (redb/SQLite test databases)
   - Memory (~80MB each)
4. Subsequent `cargo test --workspace` calls **block waiting for the lock** or contend for test database files

## Evidence (captured during crt-010 session)

- **26 zombie `cargo test --workspace` processes** found spanning 04:13 through 16:10
- All at 0% CPU — stuck, not working
- Processes from at least 6 different subagent sessions (different shell snapshot IDs)
- Current workspace test hung indefinitely until all zombies were killed

```
vscode    4947  0.0  0.3 972244 82556 ?  S  04:13  0:00  cargo test --workspace
vscode   12023  0.0  0.3 157720 79532 ?  S  04:20  0:00  cargo test --workspace
vscode   16128  0.0  0.3 157716 79536 ?  S  04:25  0:00  cargo test --workspace
vscode   25208  0.0  0.3 157728 79544 ?  S  04:40  0:00  cargo test --workspace
... (26 total)
```

## Impact

- **Implementation sessions**: Validators and testers can't run workspace tests reliably
- **Memory**: ~2GB consumed by zombie processes (26 × 80MB)
- **Developer experience**: Manual `pkill -f "cargo test"` needed before testing works
- **False failures**: Gate validators may report test failures that are actually lock contention

### Short-term (workaround)
- Add `pkill -f "cargo test"` as a pre-step before workspace test runs in validator/tester agents
- Or use `timeout` wrapper: `timeout 300 cargo test --workspace`

### Medium-term (process management)
- Subagent cleanup: when an agent completes, kill its process tree (not just the shell)
- Use process groups (`setsid`) so child processes can be killed as a group

### Long-term (test architecture)
- Run per-crate tests instead of workspace-wide: `cargo test -p unimatrix-server` etc.
- This avoids cross-crate lock contention and is faster
- Tests that passed in previous crates don't need re-running for a change in one crate

## Related

- #111: `test_store_1000_entries` rate limit issue (may be exacerbated by lock contention)
- #93: Integration test harness reliability

dug-21:
## Summary

Fixes #122 — zombie `cargo test --workspace` processes from subagents that reparent to PID 1, hold `target/.cargo-lock` and test `.db` handles, and hang or false-fail later workspace runs.

**Root cause:** the bare convention `cargo test --workspace 2>&1 | tail -30` runs cargo as the upstream writer of a non-killable pipeline. When the harness interrupts the Bash call, only the pipeline leader is signaled; cargo/rustc/test-binary children orphan to PID 1 and survive.

## Fix (Tier A — approved scope)

Hardened, byte-identical convention across every `.claude/` copy:

```bash
setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
```

- `setsid -w` — own process group (whole tree killable, no orphans) **and** propagates the child exit code.
- `timeout` (named `CARGO_TEST_TIMEOUT_SECS`, default 600) — hard ceiling; rc=124 = killed-not-failure (investigate the hang, do not blindly retry).
- file-not-pipe + `rm -f` — no live pipe to leak, PID-namespaced log cleaned on every path.
- **Single non-pipelined statement** so `rc=$?` captures cargo (via `timeout`), not `tail` (gate-integrity / false-green guard).

### Regression guard
`product/test/infra-002/check-cargo-test-convention.sh` — standalone shell scan of `.claude/` (not a cargo target, avoids circularity). Fails on **CLASS 1** (bare-pipe `cargo test` without `setsid`, the original #122) and **CLASS 2** (`setsid` without `-w`, GH#709). Ships `--self-test` proving all three cases.

## Mid-fix regression caught & fixed (GH#709, now closed)
The first commit used `setsid` **without** `-w`, which returns the fork status (always 0) — a silent false-green that inverted the gate-integrity guarantee. Caught at test verification, fixed in `03888c12`, guard strengthened to reject the form.

## Deferred (not in this PR)
- **Tier B** (SubagentStop hook reaping) — can't be built as proposed: hook is sub-50ms sync per ADR-002 and carries no pgid; deferred to a server-side follow-up.
- **Tier C** (per-crate routine testing) — the infra-002 strategic track.

## Verification
- Exit codes: `setsid -w timeout 1 sleep 5`→124, `... false`→1, `... true`→0
- Guard: 0 violations + `--self-test` PASS (deterministic, 2×)
- 12/12 setsid cargo lines use `-w`
- Integration smoke: 23 passed / 0 failed / 0 xfail
- 0 `.rs` files changed (convention + guard only); no-regression on workspace suite
- Gate: Bug Fix Validation — **PASS** (7/7)

### DIFF
--- a/.claude/agents/ndp/ndp-validator.md
+++ b/.claude/agents/ndp/ndp-validator.md
@@ -77,7 +77,8 @@ Read `.claude/skills/validate/SKILL.md` for the full procedure. Summary:
 ```bash
 cargo build --workspace 2>&1 | grep -A5 "^error" | head -20
 cargo build --workspace 2>&1 | tail -3
-cargo test --workspace 2>&1 | tail -30
+# CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
+setsid timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
 ```
 Plus anti-stub scan and deploy.sh integrity check (if deploy.sh was modified).
 
@@ -137,8 +138,9 @@ ALWAYS truncate cargo output:
 cargo build --workspace 2>&1 | grep -A5 "^error" | head -20
 cargo build --workspace 2>&1 | tail -3
 
-# Tests: summary only
-cargo test --workspace 2>&1 | tail -30
+# Tests: hardened workspace run — own process group + hard ceiling + file-not-pipe.
+# CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
+setsid timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
 
 # Clippy: first warnings only
 cargo clippy --workspace -- -D warnings 2>&1 | head -30
--- a/.claude/agents/ndp/ndp-validator.md
+++ b/.claude/agents/ndp/ndp-validator.md
@@ -78,7 +78,7 @@ Read `.claude/skills/validate/SKILL.md` for the full procedure. Summary:
 cargo build --workspace 2>&1 | grep -A5 "^error" | head -20
 cargo build --workspace 2>&1 | tail -3
 # CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
-setsid timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
+setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
 ```
 Plus anti-stub scan and deploy.sh integrity check (if deploy.sh was modified).
 
@@ -140,7 +140,7 @@ cargo build --workspace 2>&1 | tail -3
 
 # Tests: hardened workspace run — own process group + hard ceiling + file-not-pipe.
 # CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
-setsid timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
+setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
 
 # Clippy: first warnings only
 cargo clippy --workspace -- -D warnings 2>&1 | head -30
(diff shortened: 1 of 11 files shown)


dug-21 (commented):
## Security Review — PR #711 (infra-002, #122) — fresh-context

**Risk: LOW–MEDIUM. Non-blocking.** Convention/markdown + one standalone shell guard; **0 `.rs` files, 0 new deps, 0 secrets.** Exit-code semantics verified locally (`setsid -w timeout 1 sleep 5`→124, `…false`→1, `…true`→0; bare `setsid …false`→0 confirms the GH#709 false-green). Guard + `--self-test` pass on the branch.

OWASP injection surface: none meaningful. No untrusted input flows into the command — `CARGO_TEST_TIMEOUT_SECS` is operator-controlled and only ever used numerically by `timeout`; `$$` is the shell PID. Single-user dev container (`uid=1000 vscode`), sticky `/tmp`.

### Findings

**F1 — `/tmp/uni-test.$$.log` is a predictable, symlink-followable path (severity: low; non-blocking).**
The `>` redirect follows a pre-planted symlink at the target path — verified: a symlink at `/tmp/uni-test.<pid>.log` causes the redirect to truncate/overwrite the link target. `$$` is guessable and `/tmp` is world-writable (sticky only blocks *unlinking* others' files, not symlink creation). Threat model is narrow (single-user container, no hostile local user), so this is informational, not a merge blocker. If hardening later: use `mktemp` (`log=$(mktemp -t uni-test.XXXXXX)`), or write under a private dir (`${TMPDIR:-/tmp}` is the same; prefer `target/` or `$XDG_RUNTIME_DIR`). `rm -f` on a symlink also removes only the link, leaving the planted file — but harmless here.

**F2 — Guard false-negative: same-line `grep -v 'setsid'` masks a bare `cargo test |` pipe (severity: low–medium; non-blocking).**
CLASS-1 detection filters out any line containing the substring `setsid`. A line that legitimately mentions `setsid` (e.g. a per-crate hardened run, or a trailing `# …setsid…` comment) **and also** contains a second, bare `cargo test … | tail` is NOT flagged. Verified false-negative:
```
setsid -w timeout 600 cargo test -p foo > /tmp/x 2>&1; rc=$?; cargo test --workspace | tail -30
```
→ 0 hits. Also: a comment containing the word "setsid" on a bare-pipe line suppresses the catch. This is a guard *coverage* gap (lets a bad form merge), not a runtime vuln. Low likelihood given the byte-identical-copy discipline. Suggested tightening (follow-up OK): drop the line-level `grep -v setsid` and instead assert each real `cargo test` invocation is immediately preceded by `setsid -w` on the same statement, or match the bare-pipe pattern independent of whether `setsid` appears elsewhere on the line.

**F3 — Guard is line-oriented; multi-statement/heredoc forms could slip (severity: info).**
`grep -rnE` is per-line. Line-continuation (`\`) splitting `setsid` from the `cargo test | …` happened to be caught in testing (because `setsid` lands on the prior physical line), but the class is fragile. Informational only.

### Blast radius
Worst case is a **false-green gate**: if the convention silently reverts to a non-propagating form and the guard misses it (F2), a failing/killed `cargo test` reports rc=0 and a broken change merges. The PR's primary purpose is to *prevent* exactly this (GH#709), and the guard closes the common path; F2 is a residual narrow gap. No data-corruption / disclosure / privilege path — these are dev-time test-invocation conventions, not shipped server code.

### Regression risk
Low. The new form is a strict superset of the old (`setsid -w timeout … > file; tail; exit $rc`) and preserves the 30-line truncation contract. Requires `setsid` (util-linux) + `timeout` (coreutils) on the runner — present in the dev image; worth a one-line note that non-Linux/BusyBox runners may lack `setsid -w`. `rc=124` semantics are documented so gates won't misclassify a ceiling-kill as a flaky red suite.

### Verdict
Approve. F1/F2 are low-severity hardening items suitable for a follow-up, not merge blockers.

🤖 uni-security-reviewer (fresh-context security review)

PR #714, opened 2026-06-09
fix: deadline-poll test sync for fire-and-forget write races (#691, #452, #305)

### LINKED ISSUE #305: bug: test_retrospective_baseline_present — baseline_comparison is null with 3 prior MetricVectors
## Bug Report

**Test**: `suites/test_tools.py::test_retrospective_baseline_present`

**Failure**:
```
AssertionError: baseline_comparison should not be null with 3 prior MetricVectors
assert None is not None
```

**Pre-existing**: Confirmed failing on `main` before PR #304. Not related to bugfix-302.

**Steps to reproduce**:
```
cd product/test/infra-001
python -m pytest suites/test_tools.py::test_retrospective_baseline_present -v --timeout=60
```

**Context**: 
- Test seeds observation data for 4 synthetic features (col-801–col-804)
- Runs `context_retrospective` on first 3 to generate MetricVectors
- Expects 4th retrospective to return non-null `baseline_comparison`
- Server logs show `col-020: counter update failed: deserialization error: topic_delivery not found: col-801` — the synthetic features aren't registered in the delivery counter system, which may be causing the baseline lookup to fail

**Discovered during**: PR #304 (bugfix-302) integration suite run.

### LINKED ISSUE #452: [unit] col018 listener tests: assertion failure (left: 0, right: 1) — pre-existing
## Discovered By

Unit tests in `crates/unimatrix-server/src/uds/listener.rs`:
- `uds::listener::tests::col018_long_prompt_truncated`
- `uds::listener::tests::col018_prompt_at_limit_not_truncated`
- `uds::listener::tests::col018_topic_signal_null_for_generic_prompt`

## Expected Behavior

Tests should pass.

## Actual Behavior

Each test panics with: `assertion left == right failed, left: 0, right: 1`
Also observed: `A Tokio 1.x context was found, but it is being shutdown.`

## Reproduction

```bash
cargo test -p unimatrix-server --lib -- col018
```

## Notes

These tests exist on `main` branch and in `uds/listener.rs` which is not touched by vnc-012.
Not caused by the current feature under development.
Discovered during vnc-012 infra-001 integration testing.

### LINKED ISSUE #691: Flaky test: test_transcript_delta_in_batch_dropped_rest_persist — 50ms sleep races fire-and-forget batch write
## Discovered By

vnc-025 Stage 3c test execution (repeated runs of `cargo test -p unimatrix-server transcript`). Fails intermittently (~1 in 5-15 runs), including in isolation. Passes on retry.

## Test

`crates/unimatrix-server/src/uds/listener.rs::tests::test_transcript_delta_in_batch_dropped_rest_persist` (introduced in vnc-024, commit 514f2acf / 70b3aeb7).

## Expected Behavior

Batch with 2 normal events + 1 transcript_delta: exactly 2 observation rows persist.

## Actual Behavior (intermittent)

```
assertion `left == right` failed: batch: exactly the N=2 non-delta events persist; the delta is dropped
  left: 0
 right: 2
(listener.rs:5656)
```

## Root Cause

The batch observation write is `spawn_blocking_fire_and_forget`; the test synchronizes with `tokio::task::yield_now()` + `std::thread::sleep(50ms)`. Under scheduler/blocking-pool latency the write has not landed when the count assertion runs. Same fragile pattern as the sibling single-event tests, which use the identical sleep.

## Not Caused by vnc-025

- Test appears in zero +/- diff lines across all six vnc-025 commits (f1c14876..72464806).
- The vnc-025 tee on this path is a synchronous in-memory loop before the unchanged `spawn_blocking_fire_and_forget`; it adds no async work and does not delay the write.
- Failure mode is purely the pre-existing sleep-vs-spawn race.

## Suggested Fix

Replace the fixed 50ms sleep with a poll-until-deadline helper (e.g. poll `count_all_observations` every 10ms up to 2s) in these listener batch/single-event tests.

## Reproduction

```bash
for i in $(seq 1 15); do cargo test -p unimatrix-server test_transcript_delta_in_batch_dropped_rest_persist 2>&1 | grep -E "FAILED" && break; done
```

Discovered during vnc-025 integration testing. Not caused by the feature under development.

dug-21:
## Summary

Combined fix for three pre-existing, **test-only** failures — no production code changed.

### #691 + #452 — fire-and-forget write race in `#[tokio::test]` listener tests
`spawn_blocking_fire_and_forget` observation writes raced a fixed `yield_now() + std::thread::sleep(50ms)`, giving `left: 0` assertion failures. #452 additionally surfaced *"Tokio context being shutdown"* from `Handle::current().block_on(...)` inside `spawn_blocking` on a current-thread test runtime.

- New shared deadline-poll helpers `await_observations` (listener.rs) and `await_attribution` (stamp_read.rs), modeled on the in-tree canonical `transcript.rs:324` (5s deadline, 10ms poll, 100ms settle).
- **Positive** assertions (count==N / rows.len()==1) converted to poll-until-deadline.
- **Negative** assertions (count==0 / skip-observation) keep a fixed settle but swap `std::thread::sleep` → `tokio::time::sleep().await` so the reactor stays alive until any in-flight write lands. All 12 sleep sites resolved.

### #305 — analytics-drain write-visibility race (Python harness)
The analytics queue drains every 500ms; `test_retrospective_baseline_present` read col-804's baseline before 3 prior MetricVectors committed, so `compute_baselines` correctly returned `None`. Product behavior was correct; the col-020 warning is a cosmetic red herring.

- New `_wait_for_metrics_committed` poll helper waits until 3 `observation_metrics` rows commit before the 4th retrospective; `@pytest.mark.xfail` removed.

## Testing
- Bug-specific loops: col018 20/20, transcript 20/20, stamp_read 20/20, `test_retrospective_baseline_present` 3/3 — all green.
- `cargo test --workspace`: 3746 passed; the sole failure is the pre-existing unrelated flake `http::token::tests::test_concurrent_creation_no_corruption` (10/10 isolated, not in diff).
- Smoke 23/0; tools suite 188 passed.
- Gate: Bug Fix Validation — **PASS 7/7**. Diff is test-only across 3 files.

Fixes #691
Fixes #452
Fixes #305

### DIFF
--- a/crates/unimatrix-server/src/uds/listener.rs
+++ b/crates/unimatrix-server/src/uds/listener.rs
@@ -5305,6 +5305,46 @@ mod tests {
             .collect()
     }
 
+    /// Deadline-poll for a fire-and-forget `spawn_blocking` observation write to land,
+    /// then return the session's rows. Modeled on the canonical loop in
+    /// `tests/transcript.rs` (5s deadline + 100ms settle): polls `query_observations`
+    /// every ~10ms until it reaches `expected` rows, then settles 100ms so callers can
+    /// still assert the EXACT final count. Uses `tokio::time::sleep().await` (never
+    /// `std::thread::sleep`) to keep the current-thread `#[tokio::test]` reactor alive so
+    /// the in-flight `Handle::current().block_on` INSERT completes before teardown (the
+    /// "A Tokio 1.x context ... is being shutdown" race). Positive-assertion only —
+    /// never poll for an absence.
+    #[allow(clippy::type_complexity)]
+    async fn await_observations(
+        store: &Store,
+        session_id: &str,
+        expected: usize,
+    ) -> Vec<(
+        String,
+        i64,
+        String,
+        Option<String>,
+        Option<String>,
+        Option<String>,
+    )> {
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(5);
+        loop {
+            let rows = query_observations(store, session_id).await;
+            if rows.len() >= expected {
+                break;
+            }
+            assert!(
+                tokio::time::Instant::now() < deadline,
+                "timed out waiting for {expected} observation row(s) for {session_id} (got {})",
+                rows.len()
+            );
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+        }
+        // Settle so a trailing row would still fail an exact-count assertion.
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        query_observations(store, session_id).await
+    }
+
     #[tokio::test]
     async fn col018_context_search_creates_observation() {
         // T-01, AC-01: ContextSearch with valid session_id produces observation row
@@ -5338,11 +5378,7 @@ mod tests {
         )
         .await;
 
-        // Allow spawn_blocking to complete
-        tokio::task::yield_now().await;
-        std::thread::sleep(std::time::Duration::from_millis(50));
-
-        let rows = query_observations(&store, "sess-obs-1").await;
+        let rows = await_observations(&store, "sess-obs-1", 1).await;
         assert_eq!(rows.len(), 1, "expected exactly 1 observation row");
         let (sid, ts, hook, tool, input, topic) = &rows[0];
         assert_eq!(sid, "sess-obs-1");
@@ -5386,10 +5422,7 @@ mod tests {
         )
         .await;
 
-        tokio::task::yield_now().await;
-        std::thread::sleep(std::time::Duration::from_millis(50));
-
-        let rows = query_observations(&store, "sess-topic-1").await;
+        let rows = await_observations(&store, "sess-topic-1", 1).await;
         assert_eq!(rows.len(), 1);
         assert_eq!(rows[0].5.as_deref(), Some("col-018"));
     }
@@ -5427,10 +5460,7 @@ mod tests {
         )
         .await;
 
-        tokio::task::yield_now().await;
-        std::thread::sleep(std::time::Duration::from_millis(50));
-
-        let rows = query_observations(&store, "sess-generic-1").await;
+        let rows = await_observations(&store, "sess-generic-1", 1).await;
         assert_eq!(rows.len(), 1);
         assert!(
             rows[0].5.is_none(),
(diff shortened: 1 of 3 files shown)


dug-21 (commented):
## Security Review — Risk Level: LOW (non-blocking)

Fresh-context review of the full `main...HEAD` diff. **No blocking findings. Approve from a security standpoint.**

### Scope confirmation: genuinely test-only
Verified all changes land in test code — no production path altered:
- `listener.rs`: every changed line (5305–5989) is inside the `#[cfg(test)] mod tests` block that opens at line 3297 and runs to EOF. Dispatch/handler code untouched. (The `#[cfg(test)] enrich_topic_signal` at line 155 is *not* in the diff.)
- `listener/tests/stamp_read.rs`: a test submodule (`mod stamp_read;` under `mod tests`). Only the `settle()` helper and per-test waits changed.
- `test_tools.py`: new `_wait_for_metrics_committed` helper + removal of the GH#305 `xfail` marker so the test runs for real.

Diff stat: 3 files, all tests. No production source, no config, no CI, no dependency manifests.

### OWASP / input validation
- **Injection**: `_wait_for_metrics_committed` builds SQL with parameterized `?` placeholders and binds `feature_ids` as a tuple — no string interpolation of values. Clean. The `topic_with_sql_metachars_binds_parameterized` test (which exercises the parameterization guarantee) is preserved and now uses the deadline-poll helper.
- **Deserialization / access control / new external inputs**: none introduced.
- **Secrets**: no credentials, tokens, or keys in the diff.

### Regression / masking risk (the key thing to check here)
The spawn concern was whether an over-long poll deadline could hide a genuine write failure. It does not:
- Positive helpers poll `len() >= expected` then **settle 100ms** then assert the **exact** count — so a missing write still fails (timeout AssertionError), and an erroneous *trailing* row is still caught by the exact-count assertion. The deadline only shortens the happy path; it never converts a failure into a pass.
- Negative assertions (count must stay 0) were correctly **not** converted to polling — author kept a bounded fixed settle and swapped `std::thread::sleep` → `tokio::time::sleep().await`. This keeps the current-thread reactor alive so an erroneously-spawned write actually lands and gets *caught*, instead of being killed by runtime teardown. Net effect: the negative tests get **stronger**, not weaker.

### Blast radius
Worst case if these helpers have a subtle bug: a test flakes or hangs up to its 5s/10s deadline in CI — contained to the test harness. No production behavior, data integrity, or runtime path is reachable from these changes. The only trade-off is a genuinely broken write path now takes up to 5–10s to fail instead of 50ms — acceptable and standard for deadline-poll sync.

### Dependencies
None added. Rust side uses `tokio::time` (already a dep); Python side uses `time` (module-level import, line 7) and `sqlite3` (module-level import, line 951) — both already in scope.

**Verdict: LOW risk, no changes requested.**

Issue #574, opened 2026-04-22
fix(context_cycle): write cycle_events + sessions.feature_cycle via MCP handler, not hook path


dug-21:
## Problem

The \`context_cycle\` MCP tool handler does not write to \`cycle_events\`, does not update
\`sessions.feature_cycle\`, and does not update \`SessionState\` (feature, current_phase,
current_goal). All five of these mutations happen exclusively via the Claude Code hook path
(\`uds/listener.rs\` → \`handle_cycle_event\`). The MCP handler only validates, acknowledges,
drains the pending_entries_analysis bucket on Stop, and writes to \`audit_log\`.

The handler's own response text acknowledges this: *"Attribution is applied via the hook path
(fire-and-forget)."*

**Consequence for personal cloud (Codex CLI, Gemini CLI)**: These clients do not fire MCP tool
call hooks (upstream openai/codex#16732; Gemini similarly). As a result, calling
\`context_cycle\` via HTTP produces only an audit log entry. None of the behavioral state is
set. Every downstream capability that reads from \`SessionState\` degrades silently:

| SessionState field | Set by handle_cycle_event | Downstream consumers |
|--------------------|--------------------------|----------------------|
| \`feature\` | \`set_feature_force()\` on Start | PreCompact, topic signal resolution, majority-vote eviction |
| \`current_phase\` | \`set_current_phase()\` on Start/PhaseEnd; cleared on Stop | \`context_search\` phase-affinity boost, \`context_store\` phase tagging, \`context_get\` phase capture, \`context_briefing\` phase-conditioned ranking |
| \`current_goal\` | \`set_current_goal()\` on Start | \`context_briefing\` goal-cluster blending; SubagentStart routing to \`IndexBriefingService\` |

And two DB writes are also missing:

| DB write | Consumer |
|----------|----------|
| \`insert_cycle_event()\` | \`context_cycle_review\` primary path (cycle_events-first); goal embedding; behavioral signal infrastructure |
| \`update_session_feature_cycle()\` | Behavioral provenance chain: \`audit_log.session_id → sessions.session_id → sessions.feature_cycle = cycle_events.cycle_id\` |

## Additional Gap: CycleParams Has No session_id

Every other tool param struct (\`SearchParams\`, \`StoreParams\`, \`GetParams\`, \`LookupParams\`,
\`DeprecateParams\`, \`EnrollParams\`) carries \`session_id: Option<String>\`. \`CycleParams\` does not.

[...]

PR #715, opened 2026-06-09
fix: INSERT OR IGNORE in insert_session to stop compaction-resume attribution clobber (#300)

### LINKED ISSUE #300: [col-010] SessionStart hook resets started_at on context compaction, breaking retrospective attribution
## Summary

When a conversation hits the context limit and resumes (compaction), the `SessionStart` hook fires again with the same session ID. The handler uses `INSERT OR REPLACE INTO sessions`, which resets `started_at` to the resume time. This destroys the original session start time and — critically — overwrites the `feature_cycle` with whatever cycle was current at the moment of resumption.

Result: `context_retrospective` cannot find observations for the feature that was being delivered, because no session exists with that `feature_cycle`.

## Evidence

From nxs-011 delivery session `bde10d2b`:

```
ts_sec      event
1773762097  first observation (actual session start)
1773762117  cycle_start (nxs-011 set)
1773766260  cycle_stop
1773774858  cycle_start (infra-001 set — test harness run mid-delivery)
1773800040  session.started_at  ← RESET by SessionStart on compaction resume
```

- 6,000+ observations tagged `bde10d2b` spanning 10.5 hours
- `session.started_at` is 10.5 hours AFTER the first observation
- `session.feature_cycle = infra-001` (last cycle set before compaction)
- `context_retrospective(feature_cycle: "nxs-011")` → "No observation data found"

## Root Cause

`SessionStart` hook handler calls `INSERT OR REPLACE INTO sessions` unconditionally. On compaction resume, this:
1. Resets `started_at` to the resume time (losing original start)
2. Captures whatever `feature_cycle` was last set (losing nxs-011 attribution)

Session ID is **stable** across compaction — all observations use the same ID. The bug is purely in the re-insert logic.

## Fix

Change the `SessionStart` handler to use `INSERT OR IGNORE` instead of `INSERT OR REPLACE`. If a session with the same ID already exists, preserve the existing record entirely.

The `started_at` and initial `feature_cycle` from the first `SessionStart` are the authoritative values — they should not be overwritten on resume.

## Impact

Low frequency — only affects long sessions that exceed the context limit. nxs-011 was one of the largest features delivered, making it the first case where this was observed. Most sessions complete within one context window.

## Notes

- A second contributing factor in nxs-011: `context_cycle` was called with `infra-001` mid-delivery (running the infra-001 test harness as part of verification). Best practice: only call `context_cycle` for the primary feature being delivered, not for sub-systems or test harnesses run as part of verification. This is a protocol concern, not a code bug.

dug-21:
## Summary
Fixes #300. On a context-compaction resume the SessionStart hook re-fires with the same stable \`session_id\`. \`insert_session()\` used \`INSERT OR REPLACE INTO sessions\` (delete-then-insert), overwriting the existing row with inception/default values — resetting write-once \`started_at\`, clobbering a \`Declared\` \`feature_cycle\`, zeroing \`compaction_count\`, and able to revive a \`Completed\` row to \`Active\`. This broke retrospective attribution (\`context_retrospective\` / cycle review couldn't find the delivered feature's observations).

**Fix:** change the conflict clause to **\`INSERT OR IGNORE\`** so the re-fire is a no-op on an existing row. Columns that legitimately change mid-session flow through the separate \`update_session\` UPDATE path (SessionClose / cycle path), never through \`insert_session\` — so IGNORE suppresses no needed write and \`feature_cycle\` stays mutable-by-owner.

## Scope & framing
Conformance to **vnc-030 ADR-004** (FeatureSource ownership), not invention. #300 is the uncovered **persisted-table residue** of ADR-004's explicitly-accepted *in-memory* resume/compact deferral (named follow-up #4140): \`apply_stamp\` restores only in-memory \`feature_source\`, never the persisted row, and nothing restores \`started_at\`. The two column-invariants: \`started_at\` = write-once-for-everyone; \`feature_cycle\` = write-by-owner.

Two column-invariants, one statement — but REPLACE→IGNORE fixes both because every column SessionStart sends is an inception/default (the "should-refresh-on-resume" class is empty).

#574 (make the MCP handler authoritative for cycle writes; non-Claude-Code client parity) is the larger design-gated other half and is **deferred** — this PR regression-locks the invariant so #574 inherits it.

## Changes
- \`crates/unimatrix-store/src/sessions.rs\`: \`INSERT OR REPLACE\` → \`INSERT OR IGNORE\`; doc comment corrected (write-once \`started_at\`, owner-written \`feature_cycle\`); new \`#[cfg(test)] mod tests\` with 5 regression tests.

## Tests
- 5 new regression tests: \`started_at\` preserved on re-fire, \`Declared\` \`feature_cycle\` survives, closed row not revived to Active (NB-1), \`compaction_count\` not zeroed (NB-3), first-insert verbatim.
- Full \`cargo test --workspace\` — 0 failures. Integration smoke — 23 passed / 0 failed. Lib clippy clean.

## Validation
Bugfix gate: **PASS** (8/8). Diagnosis → architect design review (APPROVED WITH NOTES) → product review → human-approved → fix → gate PASS. Full trail on #300.

### DIFF
--- a/crates/unimatrix-store/src/sessions.rs
+++ b/crates/unimatrix-store/src/sessions.rs
@@ -118,15 +118,27 @@ fn session_from_row(row: &sqlx::sqlite::SqliteRow) -> Result<SessionRecord> {
 impl SqlxStore {
     /// Insert a new SessionRecord directly into the write pool.
     ///
-    /// Uses `INSERT OR REPLACE` semantics — if a record with the same
-    /// session_id already exists, it is fully overwritten.
+    /// Uses `INSERT OR IGNORE` semantics — if a record with the same
+    /// session_id already exists, the existing row is left fully intact and
+    /// this call is a no-op. This makes SessionStart idempotent: a
+    /// context-compaction resume re-fires the hook with the same stable
+    /// session_id, and the re-fire must NOT overwrite accumulated state (#300).
+    /// In particular `started_at` is write-once (no path ever rewrites it), so
+    /// the original session-age/attribution window is preserved.
+    ///
+    /// Columns that legitimately change on a live session are written via the
+    /// separate `update_session` UPDATE path, never here: `status`/`ended_at`/
+    /// `outcome`/`total_injections`/`compaction_count` at SessionClose, and
+    /// `feature_cycle` by the cycle path (`update_session_feature_cycle`).
+    /// `feature_cycle` is therefore owned-by-the-cycle-path (mutable-by-owner),
+    /// NOT immutable — IGNORE only prevents SessionStart from clobbering it.
     ///
     /// Writes directly (not via analytics drain) to ensure immediate read
     /// visibility. Session records are read immediately after insert by
     /// callers that need to verify or update them (e.g. `dispatch_cycle_start`).
     pub async fn insert_session(&self, record: &SessionRecord) -> Result<()> {
         sqlx::query(
-            "INSERT OR REPLACE INTO sessions (
+            "INSERT OR IGNORE INTO sessions (
                 session_id, feature_cycle, agent_role, started_at, ended_at,
                 status, compaction_count, outcome, total_injections, keywords
             ) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10)",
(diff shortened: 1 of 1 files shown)


dug-21:

Reviewed `git diff main...HEAD` cold (single file: `crates/unimatrix-store/src/sessions.rs`), issue #300, the full source file, and all production callers.

### OWASP
- **Injection**: No regression. The statement remains fully parameterized (`?1..?10` with `.bind(...)`); the edit changes only the conflict clause keyword (`OR REPLACE` → `OR IGNORE`). No string interpolation of session-controlled data introduced.
- **Access control / trust boundary**: Unchanged. `insert_session` is reached in production only from the SessionStart hook arm (`uds/listener.rs:785`); the only other `insert_session` call sites are `#[cfg(test)]`. No trust boundary moved.
- **Deserialization / input validation**: Unchanged. Row→record decode path (`session_from_row`) and the `SessionRecord` shape are untouched.
- **Secrets**: None introduced.
- **Dependencies**: None added (`tempfile`/`tokio` already in the dev/test graph).

### Blast radius — verified, not assumed
The risk of IGNORE is "a security/lifecycle-relevant write gets silently dropped." Checked each mutable column against an independent code path:
- `started_at` — grep confirms **no** production path writes it except `insert_session` (the lone non-test hit in `retrospective.rs:227` is a read/divide). Write-once claim holds; IGNORE simply makes the resume re-fire a no-op, which is the intended fix.
- `feature_cycle` — owner-written via `update_session_feature_cycle` (`listener.rs:2665`), which is a pure `update_session` UPDATE. It **errors** if the row is absent, so it can never resurrect a GC-deleted row, and IGNORE only blocks SessionStart from clobbering it. Test `..._preserves_declared_feature_cycle` covers this.
- `status`/`ended_at`/`outcome`/`compaction_count` — SessionClose UPDATE path. Tests `..._does_not_revive_closed_row` and `..._does_not_zero_compaction_count` confirm a closed/zombie row is not revived to Active by a re-fire. Good worst-case coverage.

Worst realistic case: if the *first* SessionStart recorded a wrong `feature_cycle`, IGNORE now preserves that wrong value instead of letting a later re-fire correct it. This is pre-existing attribution behavior, strictly better than the REPLACE bug in #300 (which captured whatever cycle was current at resume), and corrigible via the owner UPDATE path. Not a regression.

### Regression risk — low
No production code relied on REPLACE to overwrite an existing session row (only caller builds a fresh inception record). Swapping to IGNORE cannot break a legitimate replacement because none existed.

### Change minimality
Diff is minimal and on-scope: one keyword, one corrected doc comment, one `#[cfg(test)]` module. Test helper `open_test_store` is reused (cumulative test infra), not re-scaffolded. No unrelated edits.

Recommendation: approve from a security standpoint.

Issue #723, opened 2026-06-10
test-infra: pipeline_e2e tests silently skip — skip_if_no_model() model dir path mismatch (`--` vs `_`)


dug-21:
## Discovered during crt-053 (#717) Gate 3b

`skip_if_no_model()` (`crates/unimatrix-server/src/test_support.rs:87`) builds the model directory path using `--` as the separator, while the model downloader / `cache_subdir()` (`model.rs:70`) writes it with `_`. When only the `_`-named dir exists on disk (the normal case), every `pipeline_e2e.rs` test reports "ok" but its body prints "ONNX model not found ... skipping" and returns in ~0.03s — a **vacuous pass**.

## Impact
All `pipeline_e2e.rs` tests (pre-existing + the new crt-053 AC tests) silently skip on a default `cargo test`, providing no real coverage unless the model dir happens to be named with `--`. Verified: symlinking the dir to the `--` name makes the crt-053 tests execute and pass (9 passed in 2.68s, zero skip lines).

## Why not fixed in crt-053
Out of crt-053's locked C-01 scope (production change is the search.rs seed filter alone). Fixing the path mismatch has broad blast radius — it un-skips every other `pipeline_e2e` test, any of which may currently be silently-skipping-and-would-fail. Needs its own assessment.

## Fix sketch
Align `skip_if_no_model()`'s separator with `cache_subdir()` (`_`), then run the full `pipeline_e2e` suite to surface any tests that were masking failures.

Found-by: crt-053 #717 Gate 3b validation.

dug-21:
# Bug Investigation Report: 723-investigator

## Bug Summary
`pipeline_e2e.rs` (16 tests) and `graph_subgraph_integration.rs` (3 tests) silently skip on a default `cargo test`, reporting a green "N passed" while exercising none of the production path. Caused by a model-dir separator mismatch in the test skip-guard.

## Root Cause Analysis
`skip_if_no_model()` (`crates/unimatrix-server/src/test_support.rs:87`) **re-derives** the model directory name with `model_id().replace('/', "--")`, producing `sentence-transformers--all-MiniLM-L6-v2`. The canonical name is produced by `EmbeddingModel::cache_subdir()` (`crates/unimatrix-embed/src/model.rs:70`) using `.replace('/', "_")` → `sentence-transformers_all-MiniLM-L6-v2`.

[...]

Issue #585, opened 2026-05-10
fix: Path C candidate pipeline pulls deprecated entries into NLI/cosine scoring — filter upstream, not at skip guard


dug-21:
## Summary

Follow-up to #471 (orphaned-edge compaction for deprecated entries).

Path C in `nli_detection_tick` skips target entries not found in `category_map` (non-active entries) via a per-entry guard. But the skip happens *after* candidates have already been generated and scored. Deprecated entries with graph edges still appear as targets in the candidate query because their edges aren't compacted (#471), and the candidate pipeline doesn't filter by `status = Active` before scoring.

## Evidence (daemon log, 2026-05-10T17:19:00)

```
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
Path C: target entry not found in category_map (deprecated mid-tick?) — skipping tgt_id=4318
```

21 Path C skips in a single tick, including `tgt_id=4318` appearing 6 times. These entries were not deprecated mid-tick — they were already deprecated when the tick started. The log message ("deprecated mid-tick?") is misleading.

## Root Cause

Two issues compound:

1. **Candidate query doesn't filter `status = Active`** — deprecated entries with graph edges are included in the candidate set and run through cosine/NLI scoring before being discarded at Path C.
2. **No deduplication on `tgt_id` before Path C** — `tgt_id=4318` appears 6 times, likely because the deprecated entry has multiple graph edges, causing a join to multiply rows. The candidate pipeline deduplicates Informs candidates (per the tick log: `informs_candidates_after_dedup=1`) but the Path C candidate set is not similarly deduplicated.

## Fix

Add `AND e.status = 'Active'` (or equivalent) to the candidate generation query in `nli_detection_tick` so deprecated entries never enter the scoring pipeline. This makes the Path C skip guard a true safety net for genuine mid-tick deprecations only, not the primary filter for a known-deprecated population.

Deduplication on `tgt_id` in the candidate set is a secondary fix — removes the ×6 multiplier for entries with multiple edges.

## Impact

[...]

Issue #406, opened 2026-03-26
Pre-existing: test_search_multihop_injects_terminal_active fails — multi-hop supersession injection not implemented


dug-21:
## Summary

`suites/test_lifecycle.py::test_search_multihop_injects_terminal_active` fails — search results do not include the terminal active entry (C) in a multi-hop supersession chain (A→B→C).

## Failure

```
FAILED suites/test_lifecycle.py::test_search_multihop_injects_terminal_active
AssertionError: AC-13: terminal active entry C (id=3) must be injected into search results.
Multi-hop traversal (A→B→C) must follow to C, not stop at B. Got result IDs: [1]
assert 3 in [1]
```

## Analysis

The test creates a three-hop correction chain (A corrected to B, B corrected to C). When searching for content that matches A, the search result injection logic (`search.rs` Step 6b) is expected to follow the full supersession chain via `find_terminal_active` to C. Instead it either stops at the first hop or does not inject at all.

This test was added as part of crt-014 feature work. The `find_terminal_active` function or the search injection step may not traverse beyond a single superseded_by hop.

## Labels

`bug`, `testing`, `pre-existing`

dug-21:
Tracked by **crt-053 (#717)** — subsumed as **Q1 multi-correction / AC-03** (multi-hop chain A→B→C→…→head returns the terminal active head). crt-053 suspects the root cause is **graph-snapshot construction** (does the rebuilt graph carry the full chain), not `find_terminal_active` itself. **Closes via crt-053's PR.** crt-053 is HOLD behind nan-018 (#716).

dug-21:
## Disposition updated — does NOT reproduce as a retrieval defect (ass-073 #720)

My earlier "subsumed as crt-053 AC-03" framing is **superseded.** ass-073 ran: the multi-hop terminal-active redirect (`find_terminal_active` via 6b) **resolves correctly** in the eval graph rebuild — a chain A→B→C→head returns the terminal active head, with superseded members ranked below it. **#406 does not reproduce as a live retrieval bug.**

So the failing `test_search_multihop_injects_terminal_active` is most likely a **test-harness / graph-snapshot-construction artifact**, not a retrieval defect. 

**Reframed: this is a test investigation** — reconcile the failing test against actual retrieval behavior (is the test's graph snapshot built differently than the live rebuild?). It is **not** part of crt-053's retrieval work (crt-053 is now only the active-only PPR seed filter). Leave open as a test-investigation item.

Issue #405, opened 2026-03-26
Pre-existing: test_base_score_deprecated flaky — deprecated confidence can exceed active after immediate deprecation


dug-21:
## Summary

`suites/test_confidence.py::test_base_score_deprecated` fails intermittently when deprecated entry confidence exceeds active entry confidence.

## Failure

```
FAILED suites/test_confidence.py::test_base_score_deprecated
AssertionError: Deprecated conf 0.4817064500327556 should be <= active conf 0.47000000000000003
assert 0.4817064500327556 <= 0.47000000000000003
```

## Analysis

The test stores an entry, reads its confidence (active), then immediately deprecates it and reads confidence again (deprecated). The assertion `conf_deprecated <= conf_active` fails because:

1. The active confidence is computed at store time (~0.47 base score) 
2. Between the two reads, background confidence evolution (usage tracking, co-access scoring) may increment the entry's score
3. Deprecation adjusts the base score multiplier but does not re-run the full composite scoring immediately
4. Result: deprecated score slightly exceeds the stale active snapshot

## Labels

`bug`, `testing`, `pre-existing`

dug-21:
**SPLIT OUT** of the deprecation cluster (crt-053 Q4, locked): this is a confidence-pipeline / test-isolation **timing flake**, only tangentially a status-signal property. It is **NOT** part of crt-053 (#717) — it stays an **independent bugfix** so it doesn't dilute that feature's coherence. Re-tagging note: belongs with the flaky-test cleanup, not the retrieval-correctness cluster.

dug-21:
## Disposition confirmed — remains split out of crt-053 (post-rescope)

Re-confirming after the crt-053 rescope (2026-06-10): #405 stays **split out and standalone.** crt-053 is now only the active-only PPR seed filter; this is an independent confidence-pipeline / test-isolation **timing flake** (`test_base_score_deprecated`), only tangentially a status property. Not part of crt-053. Fix independently as an ordinary flaky-test bug. Leave open.

PR #724, opened 2026-06-10
[crt-053] Active-only PPR expansion seeds — surgical search-heuristic correctness

### LINKED ISSUE #704: fix(unimatrix): context_search surfaces deprecated entries instead of resolving the correction chain to its active head
## Symptom

During crt-052 scope creation, the researcher's **semantic search** surfaced two **deprecated** knowledge entries and cited them as authoritative, unaware that corrected active successors exist:

- **#4819** (vnc-030 ADR-007, `status: deprecated`) — active successor **#4834** (rev3)
- **#4746** (vnc-025 ADR-008, `status: deprecated`) — active successor **#4764**

Both are **clean `context_correct` supersession chains** — provenance is intact, not the broken-chain failure mode (cf. bugfix-473). The active entries point backward ("rev3 supersedes #4819"; #4764 carries the `clear_poison` amendment), but the **deprecated** entries carry no body-level "superseded by …" breadcrumb, and search ranked them anyway. An agent searching in good faith cited dead IDs.

The content was equivalent in both cases (#4819→#4834 changed only §1; §2/§3/§4 unchanged — #4746→#4764 carried forward + strengthened), so **no downstream constraint was actually wrong** this time. The risk is the general case: a future correction that *does* change the cited section would have an agent building on superseded guidance.

## Expected vs Actual

- **Expected:** `context_search` should resolve a correction chain to its **active head** — automatically surfacing/boosting the active successor and excluding or clearly demoting `status: deprecated` entries. Only `context_get` by explicit ID should return a deprecated entry as-is (raw, for provenance/audit).
- **Actual:** semantic search ranks deprecated entries alongside (or instead of) their active successors, with no redirect, so callers can land on and cite the dead ID.

## Regression signal

This **did not used to be the case** — search previously auto-surfaced the active version of a correction chain. Something in the search path stopped following `superseded_by` / stopped filtering `deprecated`. Treat as a likely regression, not a missing feature.

## Investigation pointers

- Correction-chain resolution in the `context_search` ranking/return path: is `status: deprecated` filtered or demoted? Is `superseded_by` followed to redirect to the active head?
- When did it change? Bisect against the point where supersession-aware search last worked.
- Confirm `context_get` is (and should remain) the **only** path that returns a deprecated entry verbatim.
- Decide the contract: does search **exclude** deprecated, **redirect** to active, or **return active with a deprecated-also-matched note**? (Redirect-to-active is the least surprising.)

## Impact

Citation-hygiene gap across the whole knowledge engine. Provenance is fine, but discoverability is broken: agents can cite superseded ADRs/patterns/lessons in scope and design work. Self-learning quality depends on search returning the *current* decision.

## Immediate mitigation (not a fix)

crt-052 is content-safe — its cited constraints (ADR-007 §2 close/sweep interface, §4 no-second-`TranscriptBuffer`-reader invariant; ADR-008 poison-recovery) are binding via the active #4834 / #4764. crt-052 should cite the active IDs; no ADR content needs changing.

## Related

- bugfix-473 lesson (broken correction chain from split deprecate+store) — this is the **inverse**: the chain is intact, but search ignores it.
- Surfaced during vnc-030 (#699) retro / crt-052 scope creation, 2026-06-08.

### LINKED ISSUE #717: feat(crt-053): active-only PPR expansion seeds — surgical search-heuristic correctness
## Rescoped 2026-06-10 (after ass-073 #720 and ass-074 #721 ran)

The original framing (status-trust re-evaluation + steepness sweep) is **retired**. The research changed it. See `product/features/crt-053/SCOPE.md`.

## TO THE DELIVERY TEAM — read before opening any code

**This is the bread-and-butter of the platform — search-result quality is the single most critical element.** The dimensions of this feature were **researched and decided by human judgment.** They are not open for re-optimization.

When you open the search pipeline you **will** see things that look like gaps (the injection-path penalty bypass, the vnc-017 redirect ceiling, deprecated entries appearing in results). **Do not close them.** Each was examined and deliberately left as-is, because:
1. **There is no tool that measures search-heuristic effectiveness** — you cannot empirically validate that a "more defensive" choice produces *better results for an agent*.
2. **The most technically defensive option is frequently NOT the most helpful one.** This is judgment, not theorem-proving.
3. **These calls were already made by a human with product context.** Do exactly what is in scope, nothing more. If you believe something in scope is wrong, **stop and raise it — do not "fix" it.**

## The change (one surgical edit)

**PPR / graph_expand expansion seeds from ACTIVE entries only.** In the expander seed collection (`services/search.rs` Phase 0, `seed_ids` ~`:915`), filter the candidate pool to `status == Active` before `graph_expand` walks from it. That is the feature — one filter on the seed set.

- PPR now expands only from current knowledge; it no longer anchors on a deprecated entry.
- **HNSW ranking is UNCHANGED** — deprecated entries still appear and are still penalized in Flexible. Two-mode design stands.
- PPR-only. No other stage, penalty, or mode modified.

## Locked (do not reopen / do not "complete")

1. Two modes stand (Flexible penalize-keep-visible; Strict evict). Returning deprecated in search is **not an error**.
2. The bar is "deprecated must not outweigh a comparable active," not "must be absent" (in Flexible). HNSW penalty + this seed filter meet it.
3. **No steepness work** — both spikes confirmed magnitude is not the lever. Q6/Q8 dropped.
4. **No injection-side redirect/penalty machinery.** The seed filter is the chosen mechanism; residual ceiling cases are knowingly accepted.
5. The vnc-017 redirect ceiling (50) is not this feature's problem.

## Cluster disposition

- **#704** — closes here (the seed filter removes the one path a deprecated entry could outrank a comparable active via PPR injection).
- **#406** — does NOT reproduce in the eval rebuild (ass-073); treat as a test/snapshot artifact to investigate, not a retrieval fix.
- **#585** — out of scope; separate edge-generation hygiene concern, track on #585.
- **#405** — split out (locked).

## Research basis

ass-073 (#720): steepness is not the lever; #406 doesn't reproduce; gate relevance on MRR not soft-GT P@5. ass-074 (#721): the expander works and is enabled in prod; the leak is latent (graph is all Active→Active); PPR seeds from the full Flexible pool with no status filter — the exposure this closes. Primary discovery: the platform cannot measure its graph-relational layer, which is why these are human-judgment calls.

## Validation

Behavior-based (crt-013 #703): assert the seed filter excludes deprecated/superseded from the expander seed set; bit-for-bit unchanged when `ppr_expander_enabled=false`; HNSW ranking untouched. Do **not** assert deprecated *absence* in Flexible (contradicts two-mode design).

---

## Delivery deliverables (synthesized 2026-06-10)

- Implementation Brief: `product/features/crt-053/IMPLEMENTATION-BRIEF.md`
- Acceptance Map (AC-01..AC-05 + anti-AC + gate trip-wires): `product/features/crt-053/ACCEPTANCE-MAP.md`
- ADR-001 (active-only seed predicate): `product/features/crt-053/architecture/ADR-001-active-only-ppr-seeds.md` (Unimatrix #4917)
- Alignment: 6/6 PASS, zero variances (`product/features/crt-053/ALIGNMENT-REPORT.md`).

**On merge: close #704** (the seed filter removes the one PPR-injection path by which a deprecated entry could outrank a comparable active). #406 → test-artifact investigation; #585 → separate edge-hygiene; #405 → split. The "no adlibs" / single-edit (C-01) directive is binding on delivery.

dug-21:
## Summary

Implements crt-053 per approved design: restrict the PPR / `graph_expand` expansion seed set to **Active** entries only. The entire production delta is a single typed-enum filter on the `seed_ids` build inside the `if self.ppr_expander_enabled` branch (`crates/unimatrix-server/src/services/search.rs`):

```rust
let seed_ids: Vec<u64> = results_with_scores
    .iter()
    .filter(|(e, _)| e.status == Status::Active)
    .map(|(e, _)| e.id)
    .collect();
```

This closes the one path (PPR injection at penalty 1.0) by which a deprecated entry could anchor graph expansion and outrank a comparable active. **HNSW ranking is unchanged** — deprecated entries still appear and are still penalized in Flexible (the two-mode design stands). Everything downstream of the BFS is untouched.

> **Binding scope (C-01):** this is a single-edit feature. The five Locked Decisions and the four-issue cluster disposition were carried verbatim through design → delivery. No injection-side redirect, no `penalty_map` change, no steepness tuning, no vnc-017 ceiling change, no #585 edge hygiene — all deliberately out of scope.

## Source Documents
- Scope: product/features/crt-053/SCOPE.md
- Architecture: product/features/crt-053/architecture/ARCHITECTURE.md (ADR-001, Unimatrix #4917)
- Specification: product/features/crt-053/specification/SPECIFICATION.md
- Risk Strategy: product/features/crt-053/RISK-TEST-STRATEGY.md

## Tests
Behavior-based acceptance (no penalty constants, no eval-harness gate — SR-01) in `crates/unimatrix-server/tests/pipeline_e2e.rs`:
- AC-01/AC-05 with **differential control arms** (R-04): forcing the deprecated seed Active makes the deprecated-only neighbor reappear — proving absence is filter-caused, not unreachability.
- AC-04 positive-retention asserts (R-02): 6b terminal-active heads / HNSW actives survive the filter.
- AC-02 off-path equivalence (expander off = bit-for-bit unchanged); AC-03 HNSW ranking unchanged.

Results: workspace 6022 passed / 0 failed; crt-053 ACs proven **non-skip** (real run, zero skip lines); infra-001 smoke 23/0; regression 289 passed / 9 pre-existing xfail.

⚠️ Pre-existing harness bug **#723** (`skip_if_no_model()` `--` vs `_` dir mismatch) silently skips `pipeline_e2e` tests on a default `cargo test`. Out of crt-053's C-01 scope; filed separately. Gate runners must assert these tests *run*, not merely that they're green, until #723 lands.

## Gate Results
- Gate 3a (Design Review): PASS (7/7)
- Gate 3b (Code Review): PASS (15/17, 2 non-blocking warnings)
- Gate 3c (Risk Validation): PASS (9/9)

## GH Issues
Closes #717
Closes #704

Cluster (not closed here): #406 (does not reproduce — test/snapshot artifact to investigate), #585 (separate edge-write hygiene), #405 (split-out flake), #723 (pre-existing test-infra skip bug).


dug-21:
## Security Review — crt-053 (PR #724)

**Reviewer:** uni-security-reviewer (fresh context) · **Risk level: LOW** · **Blocking findings: none**

Reviewed `git diff main...HEAD` cold against ARCHITECTURE.md and RISK-TEST-STRATEGY.md. Verified every claim against source rather than the narrative.

### Production change (verified)
- The only production delta is one `.filter(|(e, _)| e.status == Status::Active)` on the `seed_ids` build at `search.rs:917-921`, lexically inside `if self.ppr_expander_enabled` (`:911`). Everything else changed is test-only (`test_support.rs`, `pipeline_e2e.rs`) plus feature docs.
- **Narrowing, not widening — confirmed.** Adding a `.filter()` to a previously unfiltered iterator can only remove elements; the resulting `seed_ids` is a strict subset of the prior set. No path can widen what enters `graph_expand`.
- **Typed-enum predicate — confirmed.** `Status` is imported at module level (`search.rs:10`, `unimatrix_core::Status`) and compared by value. No string compare, so no status-spoofing / injection vector (closes R-12).

### Security boundary integrity
- **Quarantine gate UNTOUCHED.** `SecurityGateway::is_quarantined(&entry.status)` enforcement at `search.rs:956` (the per-expanded-entry gate; line shifted from ~950 only by the added comment) is byte-identical. `gateway.rs:325-326` (`*status == Status::Quarantined`) and `unimatrix-core` are not in the diff at all. The seed filter dropping Quarantined seeds is defense-in-depth, NOT a replacement for the gate — both remain in force (R-11 clear).
- Trust boundary respected: the predicate reads an already-loaded, in-memory `EntryRecord.status` on candidates already admitted by HNSW. No new external input, no new path/query-param/deserialization surface.

### OWASP scan
- Injection: none (typed enum, no shell/SQL/format-string with untrusted input). Access control: quarantine gate intact. Deserialization: no new deserialization. Input validation: no new system-boundary inputs introduced.

### Dependencies & secrets
- No `Cargo.toml` / `Cargo.lock` changes — zero new dependencies, no CVE surface change.
- No secrets. The `*token` strings flagged by a naive scan are test fixture query terms (`yttriumtoken`, `xeniumtoken`), not credentials.

### Off-path / regression
- Default-off path (`ppr_expander_enabled = false`) is structurally inert — the filter binding is lexically inside the enabled branch and referenced nowhere else, so the production-default path is bit-for-bit identical (C-02).
- `test_support` additions (`new_with_expander`, `embed_and_index`) are behind `#[cfg(any(test, feature = "test-support"))]` (`lib.rs:38`) and cannot reach a production build.
- `graph_expand` signature unchanged; only a narrower slice is passed. Crate compiles clean.

### Blast radius (worst case)
Bounded to the seed list of one optional, default-off pipeline branch. Worst realistic case if the predicate were subtly wrong is *over-drop* (a legitimate active seed excluded → expansion narrows → some current-knowledge neighbors not injected) — a recall/quality regression, NOT data corruption, DoS, info disclosure, or privilege escalation. The failure mode is safe-degraded (fewer results), not dangerous. 6b terminal-active heads pass by construction. Empty-seed boundary (all-deprecated pool) yields an empty BFS, no panic.

**Conclusion:** Minimal, in-scope, security-neutral-to-positive. Approve from a security standpoint.

Issue #728, opened 2026-06-10
fix(merge-settings): prune stale cross-matcher-group Unimatrix hooks during merge (root-cause of nan-016 switchover workaround)


dug-21:
> **vnc-031** — design complete; scope below mirrors the approved `SCOPE.md`. Planning-era content superseded.

## Summary

`mergeSettings` (`packages/unimatrix/lib/merge-settings.js`) reconciles Unimatrix hooks only inside the matcher group it manages per event (`EVENT_MATCHERS[event]`). It does **not** prune Unimatrix-owned hooks (`isUnimatrixHook`) living in *other* matcher groups for the same event. So a legacy `"*"` `PreToolUse` uni hook survives un-repointed alongside the fresh narrowed-matcher hook (pattern #4930). Affects every consumer — `init`, `init --remote`, and `scripts/dogfood-switchover.sh`. Discovered in nan-016 (#681); a script-level prune workaround was shipped there because C-8 froze this file. This is the root-cause fix.

Pruning the stale `"*"` hook **completes a data-source migration** already decided by vnc-027 ADR-004 (#4811): the broad per-tool-call observation duplicated the `PostToolUse` signal, and its richer replacement (transcript-fed cycle-review distillation, crt-052 #706) is already shipped on `main`. vnc-031 finishes the install-surface cleanup only.

## Goals
1. `mergeSettings` prunes stale uni-owned hooks across **all** matcher groups for each managed event — clean legacy-`"*"` → narrowed migration for all callers.
2. Preserve every existing semantic: `isUnimatrixHook` scoping, in-place repoint, dedup, foreign-hook preservation, emptied-group/event cleanup, idempotency.
3. Regression coverage in `test/merge-settings.test.js` for the cross-group migration.
4. Retire the bespoke prune in `scripts/dogfood-switchover.sh` (nan-016 ADR-003: one battle-tested path).

## Non-Goals
- No change to matcher-narrowing / SubagentStop opt-in (ADR-004), `isUnimatrixHook`/`UNIMATRIX_PATTERNS`, or foreign-hook handling.
- No new pruning of the opt-out path for non-registered events (already correct).
- No `mergeSettings` signature break; no new CLI flags.
- **No client / Rust / transport changes** — install-surface only.

## Dependencies
[...]

Issue #726, opened 2026-06-10
feat(vnc-032): container HTTPS serving wiring + network security posture (W2-2 follow-through)


dug-21:
## Summary

W2-2 (#658, **shipped**) added the HTTPS transport + static bearer-token auth to the server binary. But nan-014's container packaging (#629, predates the W2-2 merge) was never updated to use it — and the artifacts still carry stale "until W2-2" comments:

- `Dockerfile:134` — "No EXPOSE — no HTTP listener until W2-2 (C-10)"; `CMD … serve --foreground` (UDS-only).
- `docker-compose.yml:25` — "No port mappings — UDS mode only until W2-2 (HTTPS transport)".

Result: `docker compose up` today yields a **UDS-only instance with no network endpoint** — a remote client cannot reach it. This feature wires the container to actually serve a reachable, TLS-terminated HTTPS endpoint, with a deliberate security posture decided *as* the service is turned on (not after). It is the real prerequisite — alongside client-only init (#725) — for running the personal-cloud option.

**Recommend a design session**: this carries genuine security ADRs (auto-cert generation + client trust model, bind/publish posture, token+cert provisioning, connection-bundle UX), not just config edits.

## In scope — wiring

- Container `serve` mode that binds the HTTP/HTTPS listener (not UDS `--foreground`), driven by config/env.
- `EXPOSE` the service port; `docker-compose.yml` `ports:`/`expose:` wired per the termination model below.
- TLS configuration surface: **in-container TLS only** — auto-generate a self-signed cert+key on first boot (persisted to the data volume), or accept an operator-provided cert. Verify against W2-2's actual server surface. (Proxy-terminated is NOT in OSS — see security posture.)
- Refresh the stale Dockerfile/compose comments.

## In scope — security posture (the single OSS standard)

Decided product posture (uni-zero session 2026-06-10) — **one standard, nothing to misconfigure**:

- **In-container TLS, always. No plaintext-to-client mode. No proxy-terminated mode in OSS.** Proxy / K8s termination is deferred to the enterprise product — additive on the same config seam, zero re-architecture. (Leave the termination point as a config seam with one OSS value; do not hardcode so deeply that enterprise needs surgery.)
- **Cloud generates its own config.** On first boot the container auto-generates BOTH the bearer token AND the self-signed cert+key, persisted to the data volume. The operator invents nothing (may override with a provided token/cert).
[...]

Issue #725, opened 2026-06-10
feat(nan-019): client-only / remote init — Mac-arm + Windows pure-JS clients (split from nan-016/F5)


dug-21:
## Summary

The remote/client half of F5 (nan-016) split out into its own feature. `npx unimatrix init --remote <url> --token` produces a complete, ready-to-use **client-only** project on any Node platform — pure JS, no native binary. Targets: **Linux, macOS (Apple Silicon), Windows**, all connecting to a Linux cloud over HTTPS.

Split rationale (uni-zero 2026-06-10): the remote/client path has nothing to validate against until a cloud instance exists, so building it inside nan-016 would be speculative. It is built alongside the consuming cloud project so it validates end-to-end against a live cloud. nan-016 keeps only the UDS dogfooding switchover (Slice A) that starts the F6 (#682) soak clock.

## Scope (carried from nan-016, deferred)

- **Init unification (remote):** `init --remote` selects the TS+HTTP client; explicit mode selection layered on F3's `--remote` base flow.
- **Skills copy in remote mode** (parity with local init).
- **init ↔ uni-init boundary (OQ-4 resolved):** `init` does NOT append the CLAUDE.md knowledge block. It copies skills so `/unimatrix-init` is reachable and prints the next-step pointer; the block is `uni-init`'s job.
- **macOS (Apple Silicon) client:** `init --remote` works with **zero** native binaries (~200 KB JS).
- **Windows client (NEW):** explicit support — HTTP-remote only (UDS is Unix-only; Windows local mode stays out of scope). Validate the `node <path> hook <EVENT>` command form on Windows (no LD_LIBRARY_PATH, backslash/space path quoting), settings.json path handling, and a real Windows CI check.
- **Remote-install size gate:** remote-only install < 250 KB (no 31 MB binary, no 87 MB model).

## Out of scope

- **Server binaries / native packaging** — server is Linux-only (local binary or docker). No macOS/Windows server, no darwin cross-compile.
- **Arch-specific client packages** — pure-JS clients have no CPU-arch dimension; "Mac-arm / Windows" is a support claim, not a packaging constraint.
- **CLAUDE.md knowledge-block append** — owned by `uni-init`.

## Dependencies

- A **deployed cloud instance** to validate against. Docker distribution is already shipped — **nan-014 (#629)** publishes multi-arch images to GHCR per release tag with docker-compose — so a cloud instance is standable today; no docker-distribution work blocks this.
[...]

Issue #682, opened 2026-06-05
feat(vnc-028): Rust hook.rs client retirement gate (F6 — committed in principle)


dug-21:
## Summary

Retire the Rust `hook.rs` **client** path once TS+UDS reaches parity. **Committed in principle** (uni-zero finalization, 2026-06): the maintenance win that justified the TS direction is only real if the Rust hook client actually retires — ass-069 proved TS+UDS can reach parity, so the blocker to committing is gone.

## Scope clarity

- Retire the Rust **hook client binary path** (the `hook` subcommand routing in `init.js`), **not** the events. Discrete-signal events still fire — served by the TS client.
- The Rust binary **stays** as the MCP server entrypoint; only the hook-command routing changes.
- Reversible: the binary remains compiled; retirement is a config change in `init.js`, not code deletion.

## Gates (ass-068 Ch.5)

- Parity suite passes all events on both transports.
- Latency regression acceptable (Node ~12ms vs Rust ~6ms — measured).
- Event-queue replay + PreCompact transcript parity validated.
- **Soak evidence — two sources, named explicitly so the retirement decision rests on artifacts it actually exercised:**
  - **Runtime fidelity:** ≥4 weeks of this repo running hooks on the installed-path TS client (F5 local dogfooding switchover, #681), zero signal-loss incidents. This soaks client *behavior* (event delivery, attribution, PreCompact restoration), **not** the npm packaging path.
  - **Packaging fidelity:** the `npx … init` install gates from F5 (#681) validate the artifact users actually receive.
  - The retirement decision must cite **both**; a clean runtime soak alone does not clear the gate if packaging fidelity is unproven, and vice versa.
- Soak clock reset: any soak-resetting client change (per #681's definition) restarts the ≥4-week runtime window.

## Dependencies

F3, F4, F5 validated in production.

## Vision alignment

`personal-cloud` — collapses the two-client maintenance tax to one (JS/TS). The reason the whole client-unification effort exists.

dug-21:
## Standing input from vnc-030 (#699, F4b) for the F6 retirement gate

vnc-030 (PR #702) landed the additive `observations.topic_source` column — the **evidence base** this retirement gate reads to decide when the Rust `hook.rs` client can retire. Two constraints on how F6 must read that soak data:

### 1. Discount the absent MARKER tier (#700)
[...]

PR #729, opened 2026-06-10
[nan-016] UDS dogfooding re-release capability + switchover mechanism (F5)

### LINKED ISSUE #681: feat(nan-016): UDS dogfooding switchover — re-release + delivered switchover mechanism (F5, Slice A)
## Summary

**nan-016 — UDS dogfooding re-release capability (F5, rescoped 2026-06-10).**

This repo dogfoods the **Rust** hook (`.claude/settings.json` → `target/release/unimatrix hook <EVENT>`). The F6 (#682) retirement soak — which validates the TS hook client by running this repo's own hooks on it — cannot begin because there is **no reproducible local re-release** of the in-repo TS client to switch to, and no isolation guaranteeing that editing the in-repo client source won't perturb a running soak.

nan-016 delivers the **build + copy-install capability** and the **switchover mechanism + runbook** to repoint this repo's hooks at the installed client. It makes the switch **possible and repeatable** — it does **not** perform the switch. The flip (and therefore the F6 soak-clock start) is a deliberate later human action taken in a no-active-feature window, so it does not perturb live delivery sessions.

> **Deployment matrix settled:** server is **Linux-only** (local binary or docker, arm/intel); every non-Linux target is a **pure-JS remote client**. This repo dogfoods in **UDS-local** mode.

## Goals

1. **Local re-release capability** — reproducible build + copy-install (NEVER `npm link`) of the in-repo TS client (`packages/unimatrix`) to a stable absolute path outside the dev tree (`~/.unimatrix/dogfood-client/`). Re-running it is the F6 soak-reset point.
2. **Switchover mechanism + runbook, delivered but NOT executed** — tooling to repoint this repo's hooks at the installed-path client, plus rollback (revert to Rust `hook.rs`). The actual flip is deferred to a no-active-feature window — out of scope to execute here.
3. **Prove copy-install isolation** — edits to in-repo client source do not change installed-path behavior, verified WITHOUT flipping the live repo.
4. **No regressions** to `lib/hook-client/`, Rust `hook.rs`, or the `npx … init` local path.

> **Soak-clock note:** F6 clock starts when the switchover is **executed** (post-delivery human action), NOT at nan-016 merge.

## Non-Goals

- **Executing the live switchover** — deferred deliberate human action.
- **Rust `hook.rs` retirement / behavioral change** → F6 (#682).
- **TS hook client logic / transports / parity** → shipped in F4a/vnc-027 (#680); not modified here.
- **`npm link`-based dogfooding** — forbidden (source-mutation leak).
- **macOS local mode + darwin packages** — CUT (Mac is client-only; no Mac server binary).
- **`init --remote` unification + skills-in-remote + remote-install size gate** — DEFERRED → #725.
- **Container HTTPS serving + network posture** → #726.
- **F6 soak pass/fail duration** — F6 owns soak policy.

## Acceptance Criteria

- **AC-01**: Committed script builds + copy-installs `packages/unimatrix` to `~/.unimatrix/dogfood-client/` (copy install, never `npm link`); re-running is idempotent and is the named F6 soak-reset point.
- **AC-02**: Switchover script, when run, repoints hooks to `node <fixed-path>/lib/hook-client/index.js <EVENT>` — verified by **effect** (test fixture / dry-run / scratch dir), NOT by flipping this repo's live `.claude/settings.json`.
- **AC-03**: Copy-install isolation — editing in-repo `lib/hook-client/` does not alter the installed copy's behavior.
- **AC-04**: Runbook documents promotion (= re-run build+install = F6 reset point), rollback (revert to Rust `hook.rs`), and that the flip is deferred to a no-active-feature window.
- **AC-05**: `npx @dug-21/unimatrix init` (Linux local) byte-identical to pre-F5 (no regression).
- **AC-06**: C-04 hook-client size gate (stripped ≤ 100 KB / raw ≤ 160 KB) still passes.

## Key Decisions

- Install location: fixed dir `~/.unimatrix/dogfood-client/` (npm global prefix is node-version-pinned → breaks on container rebuild).
- Switchover lives in a committed `scripts/dogfood-switchover.sh` (one command = promotion = F6 reset point).
- Repoint via `mergeSettings` (eventual soak exercises shipped matcher semantics; narrows stale PreToolUse `"*"`), not a minimal command-string swap.
- Switchover does not manage the daemon lifecycle (UDS daemon already runs locally; hooks fail-open if absent).
- Project-root-hash detail: copy-installed client run from repo cwd resolves the **same** socket hash → no separate state dir needed.

## Constraints

C-6 copy-install only (never `npm link`) · C-7 fail-open hook posture · C-8 no `lib/hook-client/` changes (vnc-027) · C-9 dependency-free client · C-04 hook-client size gate.

## Dependencies / Tracking

Depends on F3 (#679, shipped) + F4a (#680/vnc-027, shipped). Branch `feature/nan-016`.

nan-016 **enables** the F6 (#682) clock by delivering the reset mechanism; it does **not start** it. Follow-up: the eventual switchover flip should be tracked (checklist on #682 or small issue).

Scope doc: `product/features/nan-016/SCOPE.md`.

dug-21:
## Summary

Delivers the **UDS dogfooding re-release capability** for the in-repo TS hook client (nan-016, F5 — rescoped). Provides a reproducible build + copy-install of `packages/unimatrix` to a fixed external path and a switchover mechanism + runbook to repoint this repo's hooks at the installed client — **delivered and proven by effect, but NOT executed against this repo's live settings.** The live flip (which starts the F6 #682 soak clock) is a deliberate later human action in a no-active-feature window.

Closes #681.

## What ships

| Component | Purpose |
|-----------|---------|
| `scripts/dogfood-install.sh` | `npm pack` + staged-extract + atomic-`mv` clean-replace install to `~/.unimatrix/dogfood-client/` (overridable `--target`). Idempotent, `rm`-guarded, platform-binary-absent, postinstall-inert. The named F6 soak-reset point. |
| `scripts/dogfood-switchover.sh` | `promote` / `rollback` / `--dry-run` — repoints a target settings file via the **installed** copy's shipped `mergeSettings`, plus a stale-uni-hook **prune** for a clean switch. |
| `packages/unimatrix/test/dogfood-effect.test.js` | `node --test` effect harness proving AC-02/AC-03 by a real re-fired hook against the installed path on a scratch project root, with mandatory non-vacuous negative controls. Never touches live settings. |
| `product/features/nan-016/RUNBOOK.md` | Promotion / switchover / rollback runbook, matcher-narrowing delta, fail-open posture, deferred-flip boundary. |

## Clean-switch prune (notable)

This repo's live `.claude/settings.json` has a legacy `"*"` PreToolUse Rust hook. Shipped `mergeSettings` only reconciles the matcher groups it manages, so that stale hook would survive a switchover — an unclean soak. The switchover now prunes uni-owned hooks not pointing at the switch target (foreign hooks preserved, idempotent, quote-aware whole-token match). **Root cause lives in shipped `mergeSettings` (frozen here by C-8) — tracked for a permanent fix in #728.**

## Constraints honored

- **C-8 (frozen surfaces):** `git diff main...feature/nan-016 -- packages/unimatrix/lib packages/unimatrix/package.json` is **empty** — zero changes to `lib/hook-client/`, `init.js`, `merge-settings.js`, `config.js`, or `package.json` runtime.
- **C-6** copy-install only (non-symlink asserted; never `npm link`). **C-7** fail-open hook posture. **C-04 / C-9** size + zero-deps gates pass.
- No write to the live `.claude/settings.json`; F6 soak **not** started; live flip **not** executed.

## Testing

- `node --test`: **112/112 pass** (effect harness 7/7 incl. proven negative controls; merge-settings 48/48; init 12, init-integration 8, init-remote 37).
- Gate scripts: hook-client size (76597/100000 stripped) and zero-deps both exit 0.
- Gates 3a / 3b / 3c all **PASS**. All 15 architecture risks + 9 scope risks covered.

## Not in scope

Executing the live flip / starting the F6 (#682) soak; retiring Rust `hook.rs` (→ F6); any client-logic change (→ vnc-027); macOS local mode / darwin packages (cut); `init --remote` unification (→ #725); container HTTPS (→ #726).

> Note: two early implementation commits reference `#682` in their message instead of `#681` (cosmetic; correct issue is #681).


dug-21 (commented):
## Security Review — nan-016 (fresh-context pass)

**Risk level: LOW. No blocking findings.** Reviewed the full diff cold (two shell scripts, the `node --test` effect harness, the runbook, plus design artifacts). Verified claims against actual code rather than the narrative.

### Scope confirmed
- **C-8 (zero production changes): VERIFIED.** `git diff main...feature/nan-016` touches no `packages/unimatrix/lib/**`, no `package.json`, no `crates/**`, no `Cargo.toml`. The only executable code is `scripts/dogfood-install.sh`, `scripts/dogfood-switchover.sh`, and `test/dogfood-effect.test.js`. Regression risk to existing functionality is therefore near-zero.
- **C-9 (no new dependencies): VERIFIED.** Scripts shell out to `npm`/`node`/`tar`/`git` only; the Node one-liners `require` only the *installed* `lib/merge-settings.js` + `config.js` (shipped contract) and Node stdlib (`fs`, `path`, `crypto`). No new packages.
- **No hardcoded secrets.** Full-diff scan for credentials/tokens/keys returns only the word "token" in the prune tokenizer and a documented "no secret handling (N/A)" justification. `model: claude-opus-4-8` is a default, not a credential.

### `rm -rf` / clean-replace target guard (the priority concern) — SAFE
Traced `guard_target` + `clean_replace` and exercised the guard against attack-shaped inputs. It runs **before any removal** and rejects:
- `$HOME`, `/`, `~/.unimatrix`, the repo root, and any **ancestor of the repo** (e.g. `/workspaces`) — each an explicit forbidden-path / `is_ancestor_of` reject.
- relative paths (`../../etc`) — rejected up front (must be absolute).
- empty / `.` / `..` / `/` basenames — rejected.

Coercion vectors tested and **defeated**:
- **Trailing slash** — `realpath -m` + `${%/}` normalization collapses `…/dogfood-client/` to the canonical path; no `//` double-slash.
- **Symlinked parent dir** — `realpath -m "$(dirname)"` resolves symlinks in the parent, so a symlinked path cannot smuggle the resolved target outside the guarded prefix.
- **Symlinked target itself** — `clean_replace` only ever `mv`/`rm -rf`s `$TARGET`, `.staging.$$`, `.old.$$` (siblings under `dirname "$TARGET"`). Confirmed empirically that `rm -rf`/`mv` on a symlink removes/moves the *link*, never the linked-to tree.
- `clean_replace` never traverses into an arbitrary parent; the `trap` cleanup is scoped to the same `$$`-suffixed siblings + the mktemp staging dir.

One **non-blocking design note** (already the documented intent, RISK-TEST-STRATEGY §Security Risks): the "explicit absolute `--target`" arm is permissive — it accepts any absolute path that isn't a forbidden/ancestor path (e.g. `/etc/foo`, `/home/user/x`). This is acceptable for operator-run dev tooling where the operator chose the path, the catastrophic targets are blocked, and the default (`~/.unimatrix/dogfood-client`) needs no flag. Flagging for awareness, not change.

### Injection / OWASP — no findings
- **No `eval`, no backticks, no command substitution on user input.** The two `<<NODE` heredocs are unquoted (so `$PRUNE_FRAGMENT`, a single-quoted script constant, expands), but **all user-controlled paths reach Node via `process.env` only** — never interpolated into the JS source text. No path-injection into the merge logic or into an unquoted shell command.
- Emitted hook commands are produced by the **shipped** `buildHookClientCommand` (quotes iff whitespace), not bespoke string concat.
- **No ReDoS:** the prune tokenizer `/"([^"]*)"|'([^']*)'|(\S+)/g` is linear (no nested quantifiers); 100k-char pathological input tokenizes in ~2ms.
- The platform-binary check (`find … -name 'unimatrix'`) targets the exact name `unimatrix`; the `bin/unimatrix.js` wrapper does not false-trigger it.

### Hard boundary (no live `.claude/settings.json` writes) — VERIFIED EMPIRICALLY
Ran the full effect suite (7/7 pass) with a pre/post `sha256sum` of `/workspaces/unimatrix/.claude/settings.json`: **byte-identical**. The harness reads live settings read-only (shape mirror only), routes every script `--settings` through `tmpdirGuard` (rejects the live path, asserted by T4), installs only into `os.tmpdir()`, and the `after()` hook re-hashes live settings, the real `~/.unimatrix/dogfood-client/`, and the in-repo `lib/hook-client/` to prove zero mutation even on failure paths.

### Blast radius (worst case if a subtle bug existed)
Local-only dev tooling, not network-facing. Worst case for the install script would be local data loss from a mis-targeted `rm -rf` — mitigated by the pre-removal guard above. Worst case for switchover would be writing a broken hook command into a settings file the operator explicitly pointed at; mitigated because hooks fail-open (C-7, asserted by T1/T1c re-fire exit-0) and rollback restores the exact shipped Rust form via the legacy arm (T2, idempotent, foreign-preserving). The deferred-flip boundary (scripts do not run against live settings unless a human explicitly passes `--settings`) is the load-bearing safety property and it holds.

**Recommendation: approve from a security standpoint.** No changes required.

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. Clean, well-scoped enabler for the F6 soak; the one real product question (the #728 prune workaround) is already correctly externalized.

### Vision / roadmap fit
This advances `goal:personal-cloud` — specifically the F5 slice of the OSS-cloud finalization arc (F1–F3 shipped, F4a/vnc-027 #680 shipped). It is the mechanical predecessor to F6 (#682, Rust `hook.rs` retirement), which cannot start its soak until there is a reproducible local re-release of the TS client to flip to. nan-016 delivers exactly that capability and *only* that capability — the live flip and the soak-clock start are deliberately deferred human actions. That boundary is held rigorously: live `.claude/settings.json` is byte-identical pre/post (sha256 verified independently by both 3c and the security pass), and no real `~/.unimatrix/dogfood-client/` is created by the test run.

This sits cleanly relative to in-flight work. The siblings under this goal — #725 (remote/client-only init), #726 (container HTTPS), #727 (multi-project routing) — were all explicitly carved out of nan-016's scope, so there is no sequencing collision. The deployment matrix decision (Linux-only server, every non-Linux target a pure-JS remote client) that frames this work is consistent with the standing "single edge language JS/TS" and "single binary server" principles. No vision drift.

### Approach commentary
Strengths worth noting:
- **C-8 frozen-surface discipline is exemplary.** Zero bytes touched in `lib/`, `package.json`, or `crates/`. The switchover and effect harness import the *installed* `mergeSettings`/`buildHookClientCommand`/constants rather than re-deriving them, so the dogfood path exercises shipped contract semantics, not a parallel reimplementation. This is the right call — it keeps the re-release honest.
- **The effect harness is non-vacuous**, which is the load-bearing quality property for a "prove isolation without flipping the live repo" feature. It re-fires a real installed entrypoint via `spawnSync`, and both mandatory negative controls (T1b broken-path, T1d no-prune) share the *same* assertion helper as their positive counterparts — so a regression to a no-op check surfaces. That is exactly the trap this kind of by-effect verification usually falls into, and it was avoided.
- **Security posture is appropriate to blast radius.** Local-only dev tooling; the `rm -rf` target guard runs before any removal and defeats the coercion vectors that matter (trailing slash, symlinked parent/target, ancestor-of-repo, relative paths). Independent re-run confirmed the LOW rating.

One concern the human should weigh — and it is already correctly handled, not a blocker:
- **The clean-switch prune is a workaround at the consuming layer.** The real root cause is that shipped `mergeSettings` only reconciles the matcher groups it manages and leaves stale cross-group uni hooks (this repo's legacy `"*"` PreToolUse Rust hook) firing. nan-016 was forced to add a script-level prune in `dogfood-switchover.sh` because C-8 froze `merge-settings.js`. This is the *right* short-term move (don't widen a frozen surface mid-slice), and it is properly externalized to #728 with full root-cause framing. The product risk is that #728 affects *every* `mergeSettings` consumer (`init`, `init --remote`), not just dogfooding — any user with a hand-written/legacy `"*"` uni hook hits the same unclean migration. I'd treat #728 as a real follow-up with its own design pass, not a nice-to-have, and ideally land it before the broader OSS-cloud `init` story is promoted to external users.

- **Q: Merge now, or hold for the #728 root-cause fix?** Recommend **merge now.** The workaround is contained to the dogfood script (a delivered-but-unexecuted tool), is idempotent and quote-aware, and does not touch any user-facing path. Coupling this slice's merge to a frozen-surface redesign would stall the F6 soak enabler for no safety gain.
- **Q: Is the Gate 3b WARN (708-line test harness over the 500-line guideline) acceptable?** Recommend **accept.** The cumulative-test-infra rule actively discourages splitting shared fixtures (`assertCleanPromoteState`, `reFire`, `makeScratchRoot`), and those shared helpers are precisely what make the negative controls trustworthy. The guideline targets production source; this is a test harness. Non-blocking.
- **Q: Does the permissive "explicit `--target`" arm (accepts any absolute non-forbidden path) need tightening?** Recommend **leave as-is for now.** This is operator-run dev tooling where the operator chose the path, catastrophic targets are blocked, and the default needs no flag. The security review flagged it for awareness, not change — concur.
- **Q: Should the eventual live flip be tracked?** Yes. Recommend a checklist item on **#682** (or a tiny dedicated issue) for the deferred flip, so the F6 soak-clock start is an explicit, attributable action rather than tribal knowledge. The PR body and runbook already name this boundary; just make sure it lands somewhere durable.

### Recommended actions
1. **Merge as-is.** Gates 3a/3b/3c PASS, security LOW with no blocking findings, C-8 verified empty, deferred-flip boundary intact.
2. **Keep #728 on the active follow-up list**, not parked — it is the root cause and it affects `init`/`init --remote`, not only dogfooding. Sequence it ahead of any external promotion of the OSS-cloud `init` flow.
3. **Add the deferred-flip + F6 soak-start as a tracked checklist item on #682** so the human action that starts the soak clock is explicit.
4. Cosmetic only: two early commits reference `#682` instead of `#681` (already noted in the PR body) — no action needed beyond awareness.

Issue #733, opened 2026-06-11
[vnc-034] Personal-cloud multi-project serving (umbrella: server + JS client + routing)


dug-21:
**Umbrella feature.** Personal-cloud multi-project serving: turn the partially-shipped substrate into a reachable, operator-run cloud — one Linux container serving N fully-isolated projects to N clients over pinned TLS HTTPS, single tenant, one bearer token, one connection bundle. Clients (pure-JS, any LLM CLI) attach *in*, each bound to exactly one project. The design locks six cross-cutting wire/isolation contracts (C1–C6) once so the two delivery waves slot together without interface drift, and every cloud mechanism reduces identically to the local-UDS single-project install.

Design complete (Session 1): SCOPE, SPECIFICATION, ARCHITECTURE, ADR-001..007, RISK-TEST-STRATEGY, ALIGNMENT-REPORT (PASS 6 · 0 variances). **Regenerated 2026-06-11** after the human locked four design decisions (see *Decisions — now CLOSED*).

## Delivery — ONE feature, TWO passes (NOT separate designs, NOT big-bang)

Deliver in two passes; each pass closes its wave's existing issues and references this umbrella. Do NOT close/duplicate the wave issues into this one — they are tracked *under* the umbrella per ADR-006.

| Pass | Scope | PR closes / references |
|------|-------|------------------------|
| **Wave 1 — serving + client** | Groups A + B against a single implicit project + the C1/C2 connection-contract **build-first** sub-deliverable. C1/C2/C3/C6 fully realized; C4 built minimal (`StoreResolver` trait + `SlugRouter` seam + `DefaultResolver`); C5 register/attach modeled. End state: container runs, emits a bundle, a Linux/macOS-arm/Windows client attaches over pinned TLS, knowledge flows. | PR **closes #726 + #725**, **references #733** |
| **Wave 2 — multi-project routing** | Group C against the validated Wave-1 base. Swap `DefaultResolver` → `ProjectRouter` behind the same `StoreResolver` trait + same `SlugRouter` call site; `[[projects]]` config + slug lifecycle + per-slug isolation. Purely additive — no Wave-1 client re-init. | PR **closes #727** |

[...]

Issue #727, opened 2026-06-10
feat(vnc-033): multi-project routing — path-prefix slugs, ProjectRouter, per-project isolation (W2-3)


dug-21:
## Summary

One cloud container serving **multiple** project knowledge bases, routed by an explicit project **slug** in the URL path. This is ass-060's (#643) multi-project data architecture — designed at directional confidence, deferred until the single-project cloud mechanics (#726 serve + #725 client hookup) are validated end-to-end.

Until this lands, a cloud serves exactly **one** project (the `[[projects]]`-absent default — current behavior). This feature makes one cloud serve N projects.

## The decided design (ass-060 #643)

- **Project identity = operator-declared slug** in server `[[projects]]` config (NOT a client-declared header — a typo would mint a phantom project; NOT per-project tokens — two auth mechanisms competing). Validation: `^[a-z0-9][a-z0-9-]{0,62}$` (lowercase alphanumeric + hyphen, 1–63 chars / DNS-label limit; **no underscore**). _Corrected 2026-06-11: this body previously read `^[a-z0-9][a-z0-9_-]{0,63}$`, which drifted from the locked ADR-004/FR-C5 grammar — underscore and the 64-char bound were never approved. ADR-004 is authoritative._
- **Routing = path-prefix**: `https://cloud/v1/{slug}/tools/...`. Bearer token validates at the **outer** middleware (authZ); the path slug resolves the project store at the **inner** layer (scoping); cert secures transport. Three separate concerns.
- **ProjectRouter** (ass-060 invariant #7): `slug -> Arc<ProjectContext>` (Store, VectorIndex, ServiceLayer) with a single-project default fallback. Everything downstream operates on a single-project ServiceLayer, unaware of multi-tenancy.
- **Per-slug isolation**: each project = its own `unimatrix.db`, `vector/`, `config.toml`, hash chain, and analytics under `/data/.unimatrix/{slug}/`. **No cross-project sharing in OSS** (enterprise-only). No docker-compose changes (subdirs in the one `unimatrix-data` volume).
- **MCP fit**: each project = its own MCP server entry / URL — which is *why* path-prefix beat a header (MCP clients support per-server URLs natively, not custom headers).
- **Client**: `init --remote https://host/v1/<slug>` — the cloud-wide bundle (url-base + token + cert-fingerprint) plus the per-project slug in the path.

## Scope

- ProjectRouter + path-prefix routing; `[[projects]]` config parsing + slug validation.
[...]

Issue #732, opened 2026-06-11
feat(metrics): authenticated Prometheus metrics endpoint (personal-cloud observability)


dug-21:
## Summary
Goal #4934/#4945 (`personal-cloud`) lists Prometheus metrics as a destination success criterion: *"non-root container, health checks, Prometheus metrics (not unauthenticated on a published port)."* The vnc-034 serving arc explicitly does **not** deliver a `/metrics` endpoint (Non-Goal: "No new `/metrics` endpoint … do not add an unauthenticated one"). This issue tracks metrics as a deliberate **future feature** so the goal and the delivery arc stay in lock-step.

## Scope
- An authenticated metrics surface (Prometheus exposition format) for the container deployment.
- Auth/transport posture decided at design — bearer-gated on the served port, or a separate auth domain / non-published port. **Never unauthenticated on the published TLS port** (the W2-2 prose mentioned `/metrics` aspirationally; the code has none — do not add an unauthenticated one).
- Core operational signals: request counts/latency, store-resolution outcomes, per-slug activity (post vnc-034 Wave 2), tick/rebuild health.

## Out of scope
- Dashboards / alerting (operator-side).
- Enterprise audit-confidence rework (separate, deferred — see goal "Known gap — audit confidence").

## Dependencies
- vnc-034 Wave 1 (container HTTPS serving + auth seam) — metrics binds to the same bearer/transport posture.
- Per-slug metrics depend on vnc-034 Wave 2 (`resolve_store` / `ProjectRouter`).

## Vision alignment
Advances `goal:personal-cloud` (#4934/#4945) — closes the one named success criterion the vnc-034 arc deliberately defers. Observability is part of "deploy and operate without ops friction." Re-trigger: after vnc-034 Wave 1 ships, or when a deployment needs operational visibility.

PR #734, opened 2026-06-11
[vnc-034] Wave 1 — personal-cloud single-project HTTPS serving + pure-JS remote client

### LINKED ISSUE #725: feat(nan-019): client-only / remote init — Mac-arm + Windows pure-JS clients (split from nan-016/F5)

### LINKED ISSUE #726: feat(vnc-032): container HTTPS serving wiring + network security posture (W2-2 follow-through)

dug-21:
## Summary
Implements **Wave 1** of the vnc-034 umbrella (personal-cloud multi-project serving) per the approved design: one Linux container serves a single implicit project over **pinned-TLS HTTPS** with one bearer token and one connection bundle, and a **pure-JS** client (`init --remote`) attaches over the pinned cert — no native binary, zero added runtime deps. The six cross-cutting contracts C1–C6 are locked; the C1 bundle codec + C2 cert-fingerprint parity corpus were authored once (build-first, ADR-006) and consumed byte-identically by both server and client.

The Wave-1↔2 boundary **is** the `StoreResolver` trait: the single store is served **through** the per-request `SlugRouter` seam (`PathRouter → SlugRouter → resolve_store(Default) → dispatch`), so Wave 2 (#727) swaps `DefaultResolver → ProjectRouter` at the same call site with no client re-init.

## What's included (Wave 1)
- **Server (#726):** `load_or_generate_cert` (idempotent, key `0600`, SANs from C3), `fingerprint_leaf_der` (C2 `sha256:`-leaf-DER) + Rust oracle/parity corpus, `derive_public_url` (C3 single derivation), `client-bundle` sync pre-tokio subcommand (stdout = opaque blob only; stderr = base-url+fp, **token redacted**), the C4 `StoreResolver`/`SlugRouter`/`DefaultResolver` per-request seam, TLS-only container posture + `UNIMATRIX_HTTP_ENABLED` env (ADR-007), and the cert-rotation runbook (FR-A11).
- **Client (#725):** pure-JS `init --remote <bundle> [--slug]` — 4 KB raw-string length cap **before** decode/parse, strict-schema load-bearing reject, C2 cert pin via `checkServerIdentity` with a diagnosable expected-vs-presented mismatch, skills copy, `< 250 KB` install gate, 1-client:1-project boundary unrepresentable.
- **C1/C2 contract:** committed parity corpus at `crates/unimatrix-server/tests/fixtures/c1c2-parity/`; JS goldens derived from the Rust oracle, never hand-written (SR-02).

## Source Documents
- Architecture: product/features/vnc-034/architecture/ARCHITECTURE.md (+ ADR-001..007)
- Specification: product/features/vnc-034/specification/SPECIFICATION.md
- Risk Strategy: product/features/vnc-034/RISK-TEST-STRATEGY.md
- Acceptance Map: product/features/vnc-034/ACCEPTANCE-MAP.md

## Gate Results
- Gate 3a (Design Review): **PASS**
- Gate 3b (Code Review): **PASS**
- Gate 3c (Risk Validation): **PASS**

## Tests
- Rust lib: 3945 pass (1 pre-existing `token.rs` parallel-load flake, passes in isolation; untouched by this feature)
- Rust integration: fingerprint_parity + cert_provisioner + bundle_codec + new `client_bundle_e2e` (drives the real binary: served-cert==fp, token-absent, rotation) — all green
- infra-001 smoke gate: 23 pass / 0 fail (fresh Wave-1 binary)
- JS: 840 pass / 0 fail; zero-deps + size gates pass
- Residual gaps are environment-bound live-container compose probes (AC-W1-S1/S2/S6/S7, per-OS C1), each degraded to a documented unit/e2e proof + flagged manual/CI follow-up — no silent drops.

## Scope
Wave 1 only. Wave 2 multi-project routing (#727 — `ProjectRouter`, `[[projects]]`, register/list/delete CLI) is deferred to a follow-up delivery session against this validated base.

## GH Issues
Closes #725
Closes #726
Refs #733


dug-21:
## Security Review — vnc-034 Wave 1 (fresh-context)

**Risk level: HIGH** · **Blocking: YES (1 finding)**

Overall the change set is careful and well-defended: secrets handling, the bundle trust-boundary parser, the `ProjectSlug` allowlist, the `resolve_store` funnel, dependency surface, and the container posture all hold up under review. One blocking defect: the cert-pin path rejects the server's own valid self-signed cert before the pin runs, so the OSS pinning trust model does not function over a live handshake.

---

### BLOCKING — F1: Cert pin rejects the valid self-signed cert before `checkServerIdentity` runs
**Severity: high · Location: `packages/unimatrix/lib/hook-client/cert-pin.js:75-82` (`applyCertPin`)**

`applyCertPin` sets `rejectUnauthorized = true` **and** `ca = undefined`, then attaches the custom `checkServerIdentity` pin. In Node, `checkServerIdentity` only runs *after* CA-chain verification succeeds. With `ca: undefined`, Node validates against its bundled Mozilla root store, so the server's self-signed leaf fails chain verification with `DEPTH_ZERO_SELF_SIGNED_CERT` **before** the pin function is ever called. The custom pin is dead code on the happy path: a correctly-configured client cannot connect to the correctly-configured server.

The code comment claims "CA-chain validation is bypassed by design (self-signed, no CA path)" — but the implementation does **not** bypass it. The two contradict.

Empirically confirmed against a real `https.createServer` with an openssl self-signed leaf:
```
rejectUnauthorized:true + ca:undefined + checkServerIdentity
  -> REJECTED: DEPTH_ZERO_SELF_SIGNED_CERT (pin never runs)
```

**Why the suite is green anyway:** the JS tests assert only the *options object shape* (`rejectUnauthorized === true`, `checkServerIdentity` is a function) and invoke the pin function directly with a synthetic `cert.raw`. No test performs a real TLS handshake — `test/helpers/real-server.js` uses plain `net.createServer()` (TCP, no TLS). So the live-path failure is invisible.

**Blast radius:** total loss of the remote HTTPS path — `init --remote` Ping fails, and every hook-client POST fails `connect`. Because the hook client is fail-open (resolves `connect`, never throws), production symptom is silent telemetry loss, not a loud error.

**Recommendation:** fingerprint-only pinning in Node cannot reuse the CA path. Two correct options:
1. Keep `rejectUnauthorized: true` and pass the **pinned leaf** as `ca` (requires shipping the cert/PEM in the bundle, not just the fingerprint — a wire-format change), **or**
2. Set `rejectUnauthorized: false` and perform the fingerprint check manually on the `secureConnect` event via `socket.getPeerCertificate(true).raw`, destroying the request on mismatch. NOTE: with `rejectUnauthorized: false`, `checkServerIdentity` is **not invoked at all** (verified — a wrong pin is silently accepted), so the current pin function would be a no-op there too. The check must move to `secureConnect`.

Add a real-TLS-handshake test (self-signed server, good-pin accepts / wrong-pin rejects) as the regression gate — the current shape-only assertions cannot catch this class of bug.

---

### Non-blocking observations

**N1 — Token on server first-boot stdout (pre-existing, out of scope).** `http/token.rs:101` prints `[UNIMATRIX TOKEN] {hex}` to stdout on first-boot generation. `token.rs` is unchanged in this PR and the `client-bundle` subcommand correctly uses a read-only `read_token_hex` (no print), so the bundle blob is uncontaminated. But NFR-06 ("token never in stdout") is technically violated by the server's own first-boot log line. Worth a follow-up to redact/relocate that to a 0600-file-only notice. Not introduced here.

**N2 — Pin comparison is non-constant-time** (`cert-pin.js:46`, `presented !== pinnedFp`). Acceptable: the fingerprint is public, not a secret, so a timing side-channel leaks nothing useful. No action needed.

**N3 — `cargo audit`: 1 pre-existing medium (RUSTSEC-2023-0071, `rsa` via `sqlx-mysql`).** Transitive through sqlx's mysql path (unused — project is sqlite). Not introduced by this PR; no fixed upgrade available upstream. The new deps `rcgen 0.13.2` and `base64 0.22.1` are clean. Confirmed `base64 0.22.1` is the only new lockfile entry.

---

### Verified GOOD (no action)
- **Bundle parser trust boundary** (`bundle.js` / `client_bundle.rs`): guard ordering correct — 4 KB raw-length cap *before* decode (over-cap non-base64url rejects on length), strict exactly-4-key schema is load-bearing, `base_url` https / token 64-hex / fp `^sha256:[0-9a-f]{64}$` all enforced. Error messages are token-free.
- **`ProjectSlug` allowlist** (`seam.rs`): `^[a-z0-9][a-z0-9-]{0,62}$` enforced at the parse edge via `TryFrom` before any path join; `.`/`/`/`\`/`%`/uppercase structurally cannot pass — path traversal is unrepresentable. ASCII-only charset makes the byte-length bound exact.
- **`resolve_store` funnel** (`default_resolver.rs` / `seam.rs`): `ProjectKey::Slug(_)` -> `UnknownProject`, never falls back to the default store. `ProjectKey` is transport-derived only (constructed from URL path in `parse_project_key`); no request payload names a project. `/observe` acquires its store *through* the funnel at boot.
- **`/health` + `/observe` regression**: `PathRouter` still splits `GET /health` (no auth) and `POST /observe` before the MCP arm; only the MCP fall-through goes through `SlugRouter`. `allowed_hosts` rmcp CVE-2026-42559 default is untouched (tests T-RO-05/06 assert it). Behavior unchanged.
- **Cert/key on disk** (`cert_provisioner.rs`): key written `O_CREAT|O_EXCL` mode `0600` at creation (no chmod window), cert `0644`, re-asserts `0600` on load, fail-loud on partial state, idempotent load on reboot (no silent cert rotation). Bounded 825-day validity. Token/cert/key are files on the data volume, never a DB.
- **stdout/stderr split** (`client_bundle.rs`): stdout = opaque blob only; stderr = base-url + fp echo, token absent. Unit-tested.
- **Container posture**: TLS-only `EXPOSE 8443`, no plaintext port; distroless nonroot UID 65532; secrets stay 0600 on /data, never baked into image layers; `UNIMATRIX_HTTP_ENABLED` env override is total (no panic on garbage).

dug-21:
## Security Re-Review (F1 verification) — vnc-034-security-reviewer-2

**F1 (HIGH, prior BLOCKING): CLOSED.** Verified with fresh eyes against commit `46d3aa6c` + current file state.

**1. Happy path restored.** `applyCertPin` now sets `rejectUnauthorized:false` (was `true`+`ca:undefined`), so the self-signed handshake completes and a client pinned to the server's real fingerprint connects and gets a 200. Real-TLS test (a)/(c') pass.

**2. Pin actually enforced.** With `rejectUnauthorized:false` Node skips `checkServerIdentity`, so the fingerprint is enforced manually on the TLS `secureConnect` event via `verifyPeerFingerprint(socket.getPeerCertificate(true).raw)`. A wrong/MITM cert is rejected diagnosably (expected-vs-presented + re-bundle guidance). Tests (b)/(c) pass.

**3. No token leak on mismatch (CRITICAL — verified).** In the pinned branch `req.end(body)` is called ONLY inside the `secureConnect` handler AFTER the pin passes; on mismatch the socket is `req.destroy(err)`-ed un-ended. I independently probed Node internals: the `socket` event fires before handshake completes (so `.once('secureConnect')` cannot miss it), and a request destroyed without `req.end()` transmits NO headers — the server never sees the request OR the `Authorization: Bearer` header. Tests (b)/(b') (20x hammer) confirm the token never reaches the server.

**4. No new weakness.** `rejectUnauthorized:false` is confined to the `isTls && pinnedFp` branch in `applyCertPin`; unpinned-TLS and plain-http keep Node defaults (CA verification). The transport `pinned` guard (`isTls && !!config.pinnedFp`) gates the manual-verify branch identically. Zero new runtime deps — only Node built-ins (crypto/http/https). No package.json change.

**5. Regression test is real.** `test/cert-pin-tls.test.js` spins up an actual `https.createServer` with a committed self-signed fixture and exercises a live handshake for match + mismatch + ping. 5/5 pass; remote-client unit suite 37/37 pass.

**Overall risk: LOW. Blocking: NO.** No new findings. (Pre-existing N1 token-on-first-boot-stdout and N3 transitive rsa RUSTSEC remain out of scope, non-blocking.)

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready as a product deliverable. The security review's one HIGH finding (F1) is genuinely fixed and independently re-verified closed; CI has since gone fully green except the GitGuardian dashboard incident, which is a stale-history artifact (no private key exists in the tree), not a live secret. Recommend: human dismisses the GitGuardian incident, then merge. One small follow-up worth filing (N1), not a blocker.

### Vision / roadmap fit

This PR delivers Wave 1 of vnc-034, the consolidation feature that goal #4946 (personal-cloud) names by ID: "cloud track to the destination, now consolidated under umbrella vnc-034 ... Wave 1 = #726 container HTTPS serving + security + #725 client init." The PR closes exactly #725 + #726 and refs the #733 umbrella — it lands the slice the goal asked for, no more. The four design decisions the human locked on 2026-06-11 (1:1 permanence, `UNIMATRIX_HTTP_ENABLED` env var, 4 KB bundle cap, the two elevated onboarding requirements: stderr redacted echo + cert-rotation runbook) are all present in the diff and proven by tests/AC. The delivery is faithful to the design that was approved at the design gate.

The load-bearing architectural bet — one `resolve_store` seam, single store served *through* the `SlugRouter` funnel rather than around it — is realized as designed (gate-3c verified `PathRouter` MCP edge → `SlugRouter::route_mcp` → `resolve_store`, with `DefaultResolver` returning `UnknownProject` for any slug). This is what makes Wave 2 (#727) a one-resolver-swap and keeps the cloud path byte-identical to local-UDS. The Wave-1 local-install regression test sits in the Wave-1 set, not deferred — the right guard, present. Nothing here forecloses the enterprise N:N relaxation (it stays additive on the `BearerValidator`/`unimatrix_project`-claim seam per goal #4946).

On sequencing: Wave 2 is correctly deferred to a follow-up session against this validated base, and #732 (authenticated metrics) stays fenced out as designed. No collision with in-flight personal-cloud threads.

### Approach commentary

**Strengths:**
- The security trust boundaries the design promised as fix-before-merge are real in the diff: 4 KB raw-string cap *before* base64url-decode/parse, strict-schema reject, slug allowlist at the parse edge before any path join (path traversal structurally unrepresentable), C1/C2 parity goldens derived from a single Rust oracle and consumed byte-identically by JS (never hand-written). These match the project's prior trust-boundary discipline.
- The fingerprint-pinning model is the harder thing to get right, and the security process caught the subtle failure: F1 was a *dead-code pin over a live handshake* (`rejectUnauthorized:true` + `ca:undefined` rejected the self-signed leaf before `checkServerIdentity` ran), invisible because no test did a real TLS handshake. The fix moves to a manual `secureConnect` peer-cert fingerprint check with deferred `req.end(body)` so the bearer token is never written until the pin passes — and the re-review independently probed Node internals to confirm an un-`end()`-ed request transmits zero bytes (no token leak to a MITM). A real-TLS regression test now exists. This is exactly the depth this feature's security warrants.
- Token hygiene is clean: stdout = opaque blob only, stderr = base-url + fingerprint with token redacted, proven by real-fd e2e capture.

**Concerns / things to weigh (none blocking):**
1. **N1 — server first-boot token on stdout (pre-existing, out of scope of this PR).** `token.rs` is untouched and the `client-bundle` subcommand correctly never prints the token, so this PR does not regress NFR-06. But the server's own first-boot `println!("[UNIMATRIX TOKEN] ...")` is a standing violation of the "token never in stdout" intent. It is the kind of thing that quietly undercuts the "fast + safe solo onboarding" promise (a token in scrollback / CI logs). Worth a small follow-up issue to relocate it to a 0600-file-only notice — not a reason to hold this PR.
2. **Residual coverage is environment-bound, honestly degraded.** The live-container compose probes (sibling-container HTTPS reachability, plaintext-port refusal, per-OS `init --remote`, image hardening) could not run in-session and are each degraded to a documented unit/e2e substitute plus a flagged manual/CI follow-up — not silent drops. The cross-stack contract risks those probes would touch (served-cert==fingerprint, token-absent) *are* proven end-to-end through the real binary in `client_bundle_e2e.rs`. This is acceptable, but the live-container HTTPS reachability path is the one thing genuinely unexercised by automation; I'd want the flagged manual/CI compose walkthrough actually run once before relying on the remote path in anger (it gates nothing in this PR, but it's the last untested seam).
3. **GitGuardian red is a dashboard-state artifact, not a live secret.** I verified independently: zero `.pem`/`.key` files are tracked anywhere in the tree, there is no committed `test/fixtures/tls/` directory, and the cert-pin TLS test generates its cert at runtime (agent-9's fix). The throwaway self-signed key was purged and history rewritten. The check stays red only because the GitGuardian incident is stateful and must be dismissed by a human in their dashboard — git/gh cannot clear it. This is a real but purely operational gate.

This PR carries no new open design questions — the design-gate OQs (A1, OQ-A/B/C/D) were resolved and locked before delivery, and the diff honors those resolutions. The only open items are merge-mechanical:

- **Q: Is the security posture acceptable to merge?** *Recommend: yes.* One HIGH finding, fixed, independently re-verified CLOSED with empirical no-token-leak proof and a real-TLS regression test sealing the gap that let it through. Risk level dropped high → low across the two reviews. The two non-blocking notes (N1 server-stdout token, N3 transitive `rsa` advisory via unused sqlx-mysql path) are both pre-existing and out of scope.
- **Q: Does the GitGuardian red block merge?** *Recommend: no, once a human dismisses the incident.* It is not a live secret — verified above. Do not merge while it is red without dismissing it first, so the audit trail shows a human adjudicated it rather than an override.

### Recommended actions

1. **Human dismisses the GitGuardian dashboard incident** (the purged throwaway key) — verified there is no key in the tree or current history. This is the one true remaining gate.
2. **Merge Wave 1.** CI is green across all hook-client suites (Node 18-24 × Linux/macOS/Windows — Windows has since completed and passes), zero-dep + size audits pass, Rust lib + integration suites green, security CLOSED. The delivery comment's "Windows in progress" caveat is now resolved.
3. **File a small follow-up issue for N1** (relocate the server first-boot token print to a 0600-file-only notice). Low-effort, directly serves the safe-onboarding aim; do not couple it to this merge.
4. **Run the flagged live-container compose walkthrough once** (manual or CI) before depending on the remote HTTPS path operationally — it is the only seam not exercised by automation. Not a merge gate.
5. **Proceed to Wave 2 (#727)** against this validated base in a follow-up delivery session, as scoped — the `DefaultResolver → ProjectRouter` swap behind the unchanged `StoreResolver` trait, no Wave-1 client re-init.

*Independence note: spawn prompt was clean (IDs, gate, artifact paths only) — no contract violation; this review was formed from the artifacts and my own orientation.*

Issue #735, opened 2026-06-11
[vnc-034 follow-up] token-to-stdout on first boot + Wave-1 cleanups


dug-21:
Follow-ups surfaced during vnc-034 Wave 1 delivery (PR #734). None block that PR.

### Security (review N1)
- `crates/unimatrix-server/src/http/token.rs:101` prints the bearer token to server **first-boot stdout**. Pre-existing (token.rs untouched by vnc-034) and the new `client-bundle` path is clean (token redacted from stdout/stderr/logs per NFR-06). But for the cloud HTTPS posture this stdout print is at odds with NFR-06 — evaluate redacting/gating it (it may land in container logs). Flagged by both the security reviewer and the product reviewer.

### Cleanups (Gate 3b warnings)
- `crates/unimatrix-server/src/http/router.rs` is ~562 lines (over the 500-line guideline) — pre-existing cumulative dispatch code; consider extracting.
- `crates/unimatrix-server/src/http/public_url.rs:19` retains a now-stale module-level `#![allow(dead_code)]` whose "until wiring lands" comment no longer applies (the module is consumed). Remove the allow + comment.

Refs #733.

dug-21:
The token-to-stdout / NFR-06 security item has been split out into its own dedicated issue: #736. This issue's remaining scope is the two minor code cleanups (router.rs >500 lines; stale public_url.rs `#![allow(dead_code)]`).

dug-21:
Confirmed during the first end-to-end personal-cloud dogfood (2026-06-17). The bearer token is printed to first-boot stdout and lands in `docker compose logs`:

```
[unimatrix] bearer token generated and stored (0600). Retrieve it with: unimatrix client-bundle
```

Real-world evidence that the stdout token is at odds with NFR-06 on the container HTTPS posture — it persists in container logs. Adding `goal:personal-cloud`.

dug-21:
Folded into vnc-038 and shipped via PR #772 (squash-merged to main). The three carry-items landed: token-to-stdout regression-guarded (already redacted in live code), router.rs extracted to 422 lines (≤500), and the stale public_url.rs `#![allow(dead_code)]` removed. See AC-11/12/13 in the vnc-038 acceptance map.

Issue #710, opened 2026-06-08
[flaky] http::token::tests::test_concurrent_creation_no_corruption flakes under full cargo test --workspace


dug-21:
## Summary
\`unimatrix-server\` unit test \`http::token::tests::test_concurrent_creation_no_corruption\` intermittently fails during full \`cargo test --workspace\` runs:

\`\`\`
thread 'http::token::tests::test_concurrent_creation_no_corruption' panicked at crates/unimatrix-server/src/http/token.rs:381:43:
thread 0 failed: ProjectInit("token file must contain exactly 64 hex characters, found 0")
\`\`\`

A concurrent reader observes the token file mid-write (0 hex chars) — a race in the concurrent-creation test path.

## Evidence
- Passed in one full \`--workspace\` run, failed in the next identical run (non-deterministic).
- Passes 3/3 in isolation: \`cargo test -p unimatrix-server --lib http::token::tests::test_concurrent_creation_no_corruption\`.
- Same class of full-\`--workspace\` shared-state/ordering flake as knowledge #3714.

## Triage
Pre-existing, unrelated to bugfix #122 (which changed zero \`.rs\` files; branch diffs no Rust source vs main). Not fixed in the #122 PR. Filed for separate tracking.

## Reproduction
\`\`\`bash
# flakes:
cargo test --workspace --jobs 1
# stable:
cargo test -p unimatrix-server --lib http::token::tests::test_concurrent_creation_no_corruption
\`\`\`

dug-21:
Consolidated into #742 — tracking and de-flake/hardening work moves there. Closing as consolidated.

PR #739, opened 2026-06-12
fix: redact bearer token from first-boot stdout (#736)

### LINKED ISSUE #736: [security] Eliminate bearer-token printing to stdout on server first boot (NFR-06)
## Summary
On first boot, the server prints the freshly-generated bearer token in cleartext to **stdout**, where it lands in container/process logs. This violates NFR-06 ("the token never appears in stdout, stderr, or any log line"). Surfaced by the vnc-034 security review (finding **N1**, low) and seconded by the product review on PR #734.

## Location
`crates/unimatrix-server/src/http/token.rs:101`
```rust
println!("[UNIMATRIX TOKEN] {hex_string}");
```
(doc comment at `token.rs:28` describes the same behavior)

## Why it matters
- Container stdout is routinely captured by `docker logs`, log shippers, and orchestrators — the bearer token (the sole auth credential) ends up persisted in plaintext outside the `0600` token file.
- It is **pre-existing** (this line predates vnc-034 and was untouched by it). The vnc-034 `client-bundle` subcommand is clean — it uses read-only `read_token_hex` with no print, and redacts the token from the bundle stdout/stderr per FR-A5b — so the cloud onboarding path is unaffected. This issue is specifically the server's own first-boot log line.

## Recommendation
Redact or relocate the first-boot token notice so the secret never hits stdout/stderr/logs:
- Persist the token to the `0600` file (already done) and print only a **non-sensitive pointer** — e.g. "token generated → run `unimatrix client-bundle` to retrieve it" — never the hex.
- If a local/UDS first-boot affordance is still wanted, gate any token echo behind an explicit opt-in flag and keep it out of the container default path.

## Scope notes
- Verify no other code path prints/logs the token (grep `token`, `hex_string`, `Bearer`).
- Add/extend a test asserting the token hex is absent from stdout+stderr on first boot (parallels the vnc-034 AC-W1-S5b token-absence assertions).

## Provenance
- Security review: `product/features/vnc-034/agents/vnc-034-security-reviewer-report.md` (N1)
- Product review: PR #734 advisory comment
- This supersedes the token-print item bundled in #735 (that issue's remaining scope is the two minor code cleanups).

Refs #733, #734

dug-21:
## Summary
Eliminates the cleartext bearer-token leak to stdout on server first boot (NFR-06). `write_new_token` printed the freshly-generated token hex via `println!("[UNIMATRIX TOKEN] {hex_string}")` at `token.rs:101`, where `docker logs`/log shippers persist the sole auth credential in plaintext outside the `0600` token file.

Surfaced by the vnc-034 security review (finding **N1**, low) and the PR #734 product review. Pre-existing — predates vnc-034.

## Change
- `token.rs:101` `println!`(token hex) → `eprintln!(render_first_boot_notice())` — a pointer-only message to **stderr**: `"[unimatrix] bearer token generated and stored (0600). Retrieve it with: unimatrix client-bundle"`. No hex, no file path.
- New pure `render_first_boot_notice()` builder (render-then-emit split, per established `client_bundle.rs` pattern) so the absence assertion is a true unit test.
- Doc comment (`token.rs:25-28`) updated; stale `(FR-08)` comment removed.
- Sole leak site — confirmed by investigator + architect scope checks (`main.rs` hash prints are model SHA-256s; "daemon token" traces are `CancellationToken`; client-bundle reads read-only and redacts).

## Test
- New unit test `test_first_boot_notice_token_absent_and_points_to_retrieval` asserts the rendered notice contains no token hex and includes the retrieval pointer.
- `cargo test -p unimatrix-server --lib http::token`: 17 passed.
- Workspace: 4002 passed; the 1 failure is pre-existing flake #710 (untouched code). Integration smoke 23 passed; security+protocol 35 passed.

## Reviews
- Diagnosis + design (APPROVED WITH NOTES) + product review + bugfix gate (PASS 7/7) — all on issue #736.

## Provenance note
On merge, strike the superseded N1 token-print item from #735 (its remaining scope is the two minor code cleanups).

Closes #736. Refs #733, #734.

### DIFF
--- a/crates/unimatrix-server/src/http/token.rs
+++ b/crates/unimatrix-server/src/http/token.rs
@@ -24,8 +24,10 @@ const TOKEN_HEX_LEN: usize = 64;
 
 /// Load an existing token or generate a new one.
 ///
-/// Returns raw token bytes (32 bytes), not hex. When generating a new token,
-/// prints it to stdout with a `[UNIMATRIX TOKEN]` label exactly once.
+/// Returns raw token bytes (32 bytes), not hex. On first generation, emits a
+/// non-sensitive pointer to the retrieval command on stderr; the token hex is
+/// never written to stdout, stderr, or any log line (NFR-06). The token is
+/// recoverable only from the 0600 file or `unimatrix client-bundle`.
 ///
 /// The flow is generate-first: attempt atomic file creation, and on
 /// `AlreadyExists` fall back to loading the existing file. This eliminates
@@ -97,10 +99,22 @@ fn write_new_token(
         )));
     }
 
-    // Print token to stdout exactly once (FR-08).
-    println!("[UNIMATRIX TOKEN] {hex_string}");
+    // Confirm generation on stderr with a non-sensitive pointer. The token hex
+    // is NEVER emitted (NFR-06) — render-then-emit keeps the message testable.
+    eprintln!("{}", render_first_boot_notice());
 
     Ok(token_bytes.to_vec())
+}
+
+/// Build the first-boot success notice shown after generating a new token.
+///
+/// Pure string builder (render-then-emit): it confirms the token was generated
+/// and stored, and points to the supported retrieval command. It MUST NOT
+/// contain the token hex, the token-file path, or any secret (NFR-06).
+fn render_first_boot_notice() -> String {
+    "[unimatrix] bearer token generated and stored (0600). \
+     Retrieve it with: unimatrix client-bundle"
+        .to_string()
 }
 
 /// Load and validate an existing token file.
@@ -147,6 +161,25 @@ mod tests {
     use super::*;
     use std::os::unix::fs::PermissionsExt;
     use tempfile::TempDir;
+
+    // T-TM-15: first-boot notice never contains the token hex (NFR-06) and
+    // points to the retrieval command. Mirrors client_bundle.rs token-absence
+    // unit test — render-then-emit makes this a true unit test (no binary spawn).
+    #[test]
+    fn test_first_boot_notice_token_absent_and_points_to_retrieval() {
+        // A representative 64-hex token; the notice must not echo it.
+        let hex_string = "ab".repeat(32);
+        let notice = render_first_boot_notice();
+
+        assert!(
+            !notice.contains(&hex_string),
+            "first-boot notice must never contain the token hex (NFR-06)"
+        );
+        assert!(
+            notice.contains("client-bundle"),
+            "notice must point operators to the retrieval command"
+        );
+    }
 
     // T-TM-01: test_generate_token_creates_file_with_correct_length
     #[test]


dug-21 (commented):
## Security Review — PR #739 (bugfix/736 token stdout redaction)

**Risk level: LOW. No blocking findings. Recommend merge.**

Fresh-context review of the full diff (`git diff main...bugfix/736-token-stdout-redaction`), the surrounding `token.rs`, the `main.rs` caller, and issue #736 / NFR-06.

### Leak closure — VERIFIED
- The sole leak site (`println!("[UNIMATRIX TOKEN] {hex_string}")`, `token.rs:101`) is removed and replaced by `eprintln!(render_first_boot_notice())`.
- `render_first_boot_notice()` is a pure string builder returning a static string: `"[unimatrix] bearer token generated and stored (0600). Retrieve it with: unimatrix client-bundle"`. It contains **no hex, no token-file path, no secret** — confirmed by reading the literal.
- No sibling leak path: `git grep` for `println!/eprintln!/print!` touching `token`/`hex`/`Bearer`/`TOKEN` across `unimatrix-server` returns only model-hash prints (unrelated) — the token is printed nowhere else.
- The caller (`main.rs:901`) consumes the returned `token_bytes` into a `[u8;32]` and never prints or logs it. `load_existing_token` emits only a `tracing::debug!` of the **path** (not the token), unchanged.

### stderr acceptability under NFR-06 — OK
NFR-06 forbids the *token* in stdout/stderr/logs, not all output. The emitted message is a non-sensitive operational pointer (exactly the issue's recommended remediation). Routing it to stderr is appropriate.

### Retrieval pointer — VALID
`unimatrix client-bundle` is a real CLI subcommand (`Command::ClientBundle`, `main.rs:354`; clap kebab-cases to `client-bundle`). The notice points operators to a working, token-clean retrieval path (client_bundle.rs uses read-only `read_token_hex` and has its own token-absence test).

### Blast radius — contained
Worst case of a subtle bug in the changed code: a wrong/empty notice string. That is a UX/operability regression (operator can't find the retrieval hint), **not** a secret disclosure or auth bypass — the token still persists to the 0600 file and auth is unaffected. No data corruption, DoS, or privilege path.

### Regression risk — none observed
- Return type (`Result<Vec<u8>, ServerError>`), signature of `write_new_token`/`load_or_generate_token`, file permissions (0600, atomic `create_new`), and the generate-first/TOCTOU-safe flow are all unchanged.
- All 17 `http::token` unit tests pass locally, including the new test.

### Test adequacy — sufficient for this leak
`test_first_boot_notice_token_absent_and_points_to_retrieval` asserts a representative 64-hex token is absent from the notice and that it points to `client-bundle`. Render-then-emit makes the emitted bytes the test subject (no fd capture / binary spawn needed), so it genuinely guards the regression. Mirrors the established `client_bundle.rs` token-absence pattern.

### Minor (non-blocking) observations
- **Stale docs (out of scope for this PR, worth a follow-up):** `README.md:452`, `docs/client-setup.md:7`, and several `vnc-021` artifacts still tell operators the token is "printed to stdout once with `[UNIMATRIX TOKEN]`". That guidance is now wrong; a doc update should redirect operators to `unimatrix client-bundle`. Code is correct — flagging so the behavior/docs don't drift.
- Test uses a fixed `"ab".repeat(32)` token. Since the notice is a constant with no hex substitution, the assertion is effectively structural — adequate here, but it would not catch a future change that interpolates a *different* value. Acceptable given render-then-emit.

Diff is minimal (single file, single commit) and matches the issue's recommended remediation exactly.

PR #740, opened 2026-06-12
fix: align skip_if_no_model() model dir to cache_subdir() + loud-fail guard (#723)

### LINKED ISSUE #723: test-infra: pipeline_e2e tests silently skip — skip_if_no_model() model dir path mismatch (`--` vs `_`)

dug-21:
Fixes #723.

## Root cause
`skip_if_no_model()` re-derived the model cache dir name with `model_id().replace('/', "--")`, while the canonical `EmbeddingModel::cache_subdir()` (used by both the downloader `ensure_model` and the ONNX loader) sanitizes with `_`. The `--` path never existed on disk, so all **19** model-gated integration tests (16 `pipeline_e2e.rs` + 3 `graph_subgraph_integration.rs`) silently early-returned — a vacuous green. Born in col-015 (#161); surfaced by crt-053 (#717) Gate 3b.

## Fix (two commits — symptom and disease, bisectable)
**Commit 1 — `ae57e2ef`** the one-liner: `test_support.rs` now derives the dir via the canonical helper (`cache_dir.join(config.model.cache_subdir())`). This is the bisectable un-skip.

**Commit 2 — `9acf18c7`** hardening of the actual defect class (a guard that *fails green*):
- `skip_if_no_model()` moved to `model_guard.rs` (re-exported from `test_support` for call-site compat; gated `cfg(any(test, feature = "test-support"))` — never in the production build).
- Replaces the boolean check with a **three-state classifier** (`Present` / `Absent` / `Mismatched`): canonical path present → run; model genuinely absent → clean skip (model-less CI); a dir for *this* model id present at a non-canonical (drifted) path → **`panic!`** naming found vs expected.
- Drift detection matches against an exact candidate set for the target model id, so a sibling dir (e.g. the NLI model) provably cannot trigger a false loud-fail.
- **Structural regression test**: asserts the guard's dir == the `cache_subdir()`-derived downloader dir. Runs without a model on disk; fails if anyone re-introduces separator drift.

> The split into "one-liner now, hardening later" was considered and rejected (see issue): the 19-test un-skip is the real blast radius regardless, the structural test is *this bug's own regression test*, and the fail-green guard is the defect that escaped twice. The clean-narrative concern is honored at the commit level, not by fragmenting the issue.

## Verification
- **Un-skip proven**: `pipeline_e2e` 16 passed @ **1.00s** (vs ~0.01s vacuous), `graph_subgraph_integration` 3 passed @ 0.25s, **zero** "model not found … skipping" lines. No previously-masked failures surfaced — all 19 pass when actually run.
- `model_guard` unit tests: **7 passed** (classifier + structural regression + sibling-NLI no-false-fail + canonical-precedence).
- Full workspace: **6204 passed, 0 failed**. Integration smoke: 23 passed.
- Clippy: the 54 errors are pre-existing Rust 1.95 lints in `unimatrix-observe`/`unimatrix-engine` (GH#688) — **zero in any touched file**.

## Changed files
- `crates/unimatrix-server/src/test_support.rs` — one-line alignment + re-export
- `crates/unimatrix-server/src/model_guard.rs` — new: classifier + loud-fail + tests
- `crates/unimatrix-server/src/lib.rs` — module declaration

Gate: Bug Fix Validation — **PASS** (7/7).

### DIFF
--- a/crates/unimatrix-server/src/lib.rs
+++ b/crates/unimatrix-server/src/lib.rs
@@ -38,6 +38,8 @@ pub mod snapshot;
 pub mod uds;
 
 #[cfg(any(test, feature = "test-support"))]
+pub mod model_guard;
+#[cfg(any(test, feature = "test-support"))]
 pub mod test_support;
 
 // Re-exports for external consumers (main.rs, integration tests).
--- a/crates/unimatrix-server/src/model_guard.rs
+++ b/crates/unimatrix-server/src/model_guard.rs
@@ -1,0 +1,262 @@
+//! Model-availability guard for server-level pipeline tests (#723).
+//!
+//! `skip_if_no_model()` decides whether the `pipeline_e2e` / `graph_subgraph_*`
+//! integration tests can run against a real ONNX model. It MUST derive the model
+//! directory the same way the downloader (`ensure_model`) and loader
+//! (`OnnxProvider::new`) do — via [`EmbeddingModel::cache_subdir`].
+//!
+//! Born wrong in col-015 (#161): the guard re-derived the dir name with a `--`
+//! separator while the canonical helper uses `_`. The drifted path never existed,
+//! so every guarded test silently skipped and reported a vacuous green (#723).
+//!
+//! This module makes the path derivation a single source of truth and replaces
+//! the fail-green behavior with a three-state classifier that fails LOUDLY when a
+//! model dir for the target id is present but at a non-canonical (drifted) path.
+
+use std::path::{Path, PathBuf};
+
+use unimatrix_embed::{EmbedConfig, EmbeddingModel};
+
+/// Canonical on-disk model directory for `model` under `cache_dir`.
+///
+/// Single source of truth for the guard's path derivation: delegates to
+/// [`EmbeddingModel::cache_subdir`] — the SAME helper the downloader and loader
+/// use. Exposed as a standalone fn (not inlined) so the structural regression
+/// test can assert guard-path == `cache_subdir()`-path WITHOUT a model on disk.
+fn canonical_model_dir(cache_dir: &Path, model: EmbeddingModel) -> PathBuf {
+    cache_dir.join(model.cache_subdir())
+}
+
+/// Three-state classification of a model's on-disk presence under `cache_dir`.
+enum ModelDirStatus {
+    /// Canonical path exists with the ONNX file — tests may run.
+    Present,
+    /// No directory for THIS model id exists under `cache_dir` — skip cleanly.
+    Absent,
+    /// A directory resolving to THIS model id exists, but NOT at the canonical
+    /// path (separator drift). Carries (found, canonical) for the loud-fail
+    /// message. NOT a skip — a hard failure.
+    Mismatched { found: PathBuf, canonical: PathBuf },
+}
+
+/// Plausible sanitized directory names for `model_id`'s org/repo `/` boundary.
+///
+/// `cache_subdir()` replaces `/` with `_`; historical drift used `--` (#723);
+/// `-` is included for completeness. These are the only forms that resolve to
+/// THIS model id. Matching against this exact set — rather than a fuzzy
+/// normalization — guarantees a sibling model dir (e.g. the NLI model's
+/// `nli-minilm2-l6-h768`) can never trigger a false mismatch: it is not in the set.
+fn drift_candidate_names(model_id: &str) -> Vec<String> {
+    ["_", "--", "-"]
+        .iter()
+        .map(|sep| model_id.replace('/', sep))
+        .collect()
+}
+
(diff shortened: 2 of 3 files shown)


dug-21:

**Risk level: LOW** · **Blocking findings: none**

Reviewed the true change set (`git diff $(git merge-base HEAD origin/main)..HEAD`), not plain `origin/main` — scope is exactly three files: `model_guard.rs` (new), `test_support.rs`, `lib.rs`. Unrelated #736/#739 advancement on origin/main was excluded.

### Scope & containment
- `model_guard` is gated `#[cfg(any(test, feature = "test-support"))]` in `lib.rs`. **Confirmed it never compiles into the production binary** — the entire blast radius is test-time only.
- No new dependencies. `tempfile = "3"` is already a dev-dependency; no Cargo.toml change.
- No secrets, tokens, or credentials. No `unsafe`. No shell, SQL, network, or deserialization surface.

### OWASP / input-validation
- **Path traversal: not applicable.** Paths are derived from `EmbeddingModel` — a closed `enum` with `model_id()` returning `&'static str` compile-time constants. `model_id` is never attacker- or operator-controlled, so `replace('/', sep)` cannot produce `..` or absolute-path injection. `cache_dir` comes from `resolve_cache_dir()` (config field → `UNIMATRIX_MODEL_CACHE` env → platform default) — same trust level as the downloader/loader already use.
- **Filesystem interaction** is read-only: `read_dir(cache_dir)` + `Path::exists()`. No writes, no deletes in the guard itself (writes are confined to `#[cfg(test)]` tempdirs). `read_dir` failure is handled (`if let Ok(...)`) and degrades to `Absent` (clean skip), not a panic.

### Correctness of the fix
- The fix collapses path derivation to a single source of truth: `cache_dir.join(model.cache_subdir())`, where `cache_subdir() = model_id().replace('/', \"_\")` — the **same** helper `download.rs` and the ONNX loader use (verified at `unimatrix-embed/src/model.rs:69` and `download.rs:12,77`). The old guard's `replace('/', \"--\")` is the #723 defect; it is gone.
- The three-state classifier correctly excludes the canonical name from drift matching (so a genuinely-missing ONNX at the canonical path → `Absent`/skip, not a false loud-fail), and the `drift_candidate_names` exact-set match (`_`, `--`, `-`) prevents a sibling model dir (e.g. NLI `nli-minilm2-l6-h768`) from triggering a false `Mismatched` — covered by `test_classify_model_dir_sibling_nli_dir_does_not_false_fail`.

### Blast radius (worst case if subtly wrong)
Test-only. Worst case is a misclassification: a false `Mismatched` would `panic!` and turn the integration tests red (loud, safe — fails toward CI noise, never toward silent green). A false `Absent` would re-introduce a vacuous skip — but that is strictly no worse than today's pre-fix behavior, and the new structural test (`test_canonical_model_dir_matches_cache_subdir_no_drift`) runs **without** a model on disk and guards against separator re-drift directly. No data corruption, disclosure, or privilege path exists.

### Regression risk
- `skip_if_no_model` is re-exported from `test_support` (`pub use crate::model_guard::skip_if_no_model`), so all existing call sites (`use crate::test_support::skip_if_no_model`) keep compiling — verified across `pipeline_e2e.rs` and `graph_subgraph_integration.rs`.
- The intended behavioral change is that ~19 previously-skipped tests now **execute** when a model is present. That is the point of the fix; risk is surfacing latent failures in those tests, not in this guard. On model-less CI they still skip cleanly (`Absent` → `true`).

**Recommendation: approve from a security standpoint.** No blocking findings.

Issue #747, opened 2026-06-12
[infra-002] Tier B — server-side subagent process-group reaping (design pass) (split from #712/#742)


dug-21:
## Summary
Split from #712 (closed/consolidated into #742). Tier B is **not a bugfix** — it needs its own design pass — so it was carved out of the #742 de-flake bugfix and is tracked here.

## Problem
Tier A (#122, knowledge #4874) stopped a single interrupted `cargo test` run from orphaning its process tree via `setsid -w timeout`. Tier B is defense-in-depth: reap a subagent's orphaned cargo/test process tree **server-side**, off the already-recorded `SubagentStop` event.

## Why it cannot be a quick hook change (from #712)
- The `SubagentStop` hook is a sub-50ms sync CLI per ADR-002 (fire-and-forget telemetry) — inline process-killing violates that contract.
- `HookInput` (`crates/unimatrix-engine/src/wire.rs`) carries no PID/PGID — only `session_id`/`cwd`/`transcript_path`. "Kill by stored pgid" has no source for the pgid.
- Over-broad `pkill -f "cargo test"` would kill a sibling agent's legitimate in-flight run in a swarm.

## Correct design (needs its own pass)
Capture each subagent's process-group id **at spawn**, correlate by `session_id`, and reap **server-side** off the recorded `SubagentStop` event — keyed on an exact owned pgid, never an argv/name match. Touches the server's session model.

## Note
This bugfix-cluster session (#742) was itself bitten by orphaned/parallel-agent state — a strong motivator for getting Tier B's owned-pgid reaping right.

Split from #712 / #742.

Issue #746, opened 2026-06-12
test: test_ac14 cond.1 (~2%) — HNSW approximate-retrieval membership flip; widen eval retrieval breadth (follow-up to #742)


dug-21:
## Summary
Follow-up split from #742. During #742's high-sample verification (n≥285), `eval::runner::sweep_tests::test_ac14_correlated_sweep_non_vacuous` was found to still flake **~2%**, exclusively at **cond.1** (`non_vacuous_trust_seen`, `sweep_tests.rs:218`). This is a **distinct defect** from the cond.4 bit-for-bit drift fixed in #742 (commits c2edbda9 + 2f6e73fc). cond.4 is green (0 failures / 285 runs).

## Root cause (adjudicated in #742, three agents + reproduced 195/200)
HNSW **approximate top-k membership flip** in the eval retrieval path:
- The eval corpus is ~25 entries; queries retrieve **top-5 approximate** (`hnsw_k = params.k = 5`; NLI off by default per `config.rs:1014`; expander off per `search.rs:5516`).
- `eval/corpus/embed.rs::embed_and_write_vectors` rebuilds the sweep index per test invocation via `VectorIndex::new` (hnsw_rs 0.3.4). `LayerGenerator` reseeds from OS entropy each build — **no public seeding API** on the pinned version.
- ~2% of builds drop an anchor from the approximate top-5 before it ever reaches the sort → `pr.entries` vacuous → cond.1 fails. No `services/search.rs` sort/tie-break can recover an anchor HNSW never returned (mechanically disproven in #742: id tie-break on the final sort left the rate unchanged at 147/150).

## Proposed fix (minimal, from #742 architect adjudication)
Decouple **eval retrieval breadth** from metric k: exhaustive / wide-enough retrieval in `replay.rs` / `sweep.rs` so the tiny corpus is fully returned and approximate search cannot drop an anchor; **metric k stays 5** (`compute_p_at_k` takes its own k). Eval-harness only — no `embed.rs`, no production retrieval defaults, no hnsw_rs seeding. **ADR-002-conformant** (nan-018 #4895: embed-at-load + shape-hash guard; seeding/sidecar would violate it).

### Open design question (resolve before implementing)
One-k vs two-k: trust/cost metrics also read `pr.entries`. Widening retrieval breadth for the present/rank verdict must not silently change what trust/cost see. Decide whether retrieval breadth and metric-k are independently parameterized.

### Adjacent question carried from #742
[...]

Issue #738, opened 2026-06-11
test: two pre-existing non-deterministic flakes in unimatrix-server lib suite


dug-21:
## Summary

Two pre-existing tests in the `unimatrix-server` lib suite fail intermittently under the full parallel `cargo test --lib` run but **pass deterministically in isolation**. Surfaced repeatedly during vnc-034 Wave 2 (#727) delivery; neither is caused by that feature (it touches no token or eval code). Filing so they're tracked and de-flaked rather than re-diagnosed each cycle.

### 1. `http::token::tests::test_concurrent_creation_no_corruption`
- Symptom: intermittent failure under the parallel suite; passes in isolation.
- Nature: a concurrency test — likely a race in the test setup/assertion (or a genuinely racy token-creation path worth confirming is test-only).

### 2. `eval::runner::sweep_tests::test_ac14_correlated_sweep_non_vacuous`
- Symptom: `assertion left == right failed: cond.4: swept baseline (default penalties) must reproduce behavior bit-for-bit`. Observed left/right differ only by ~1e-9 in values AND by HashMap key ordering.
- Nature: a "bit-for-bit" reproduction assertion over floating-point penalty maps with non-deterministic iteration/accumulation order. Bit-for-bit equality is too strict for an order-sensitive float computation.
- Reproduced isolated: `cargo test -p unimatrix-server --lib eval::runner::sweep_tests::test_ac14_correlated_sweep_non_vacuous` -> ok.

## Suggested fixes
- (1) Make the concurrency test deterministic (serialize the racy section, or assert the invariant rather than a timing-sensitive observable). Confirm the production token path is not actually racy.
- (2) Fix the underlying iteration-order non-determinism (canonicalize/sort before the reproduction compare) or relax the assertion from bit-for-bit to epsilon + order-independent comparison — whichever preserves the test's intent.

## Context
- Both pass in isolation, so CI red/green is scheduling-dependent — latent flakiness, not a regression.
- Repeatedly noted across recent deliveries (vnc-034 Wave 2 #727 and adjacent runs). Not blocking, but accumulating noise that masks real signal.

dug-21:
Consolidated into #742 — tracking and de-flake/hardening work moves there. Closing as consolidated.

Issue #712, opened 2026-06-08
[infra-002] cargo-test reliability follow-ups: Tier B server-side reaping + guard/path hardening


dug-21:
Follow-up to #122 (Tier A merged in #711). Tier A stopped a single interrupted `cargo test` run from orphaning its process tree via `setsid -w timeout`. This issue tracks the deferred defense-in-depth and hardening items.

## Tier B — server-side subagent process-group reaping
Originally proposed as inline killing in the `SubagentStop` hook, but that **cannot** be built as specified:
- The hook runs as a sub-50ms sync CLI per ADR-002 (fire-and-forget telemetry) — inline process-killing violates that contract.
- `HookInput` (`crates/unimatrix-engine/src/wire.rs`) carries no PID/PGID — only `session_id`/`cwd`/`transcript_path`. "Kill by stored pgid" has no source for the pgid.
- Over-broad `pkill -f "cargo test"` would kill a sibling agent's legitimate in-flight run in a swarm.

**Correct design (needs its own pass):** capture each subagent's process-group id at spawn, correlate by `session_id`, and reap **server-side** off the already-recorded SubagentStop event — keyed on an exact owned pgid, never an argv/name match.

## Hardening from the #711 security review (low severity, non-blocking)
- **F1** — `/tmp/uni-test.$$.log` is a predictable, symlink-followable path; the `>` redirect overwrites a pre-planted symlink target. Narrow threat model (single-user container, sticky /tmp) but `mktemp`/private dir would harden it. Apply to the canonical convention in `.claude/rules/rust-workspace.md`.
- **F2** — `check-cargo-test-convention.sh` CLASS-1 detection filters any line containing the substring `setsid`, so a line that mentions `setsid` *and also* contains a second bare `cargo test … | tail` is not flagged (verified false-negative). Tighten the guard regex.

## Tier C — per-crate routine testing (= infra-002 strategic track)
Default routine validation to per-crate `cargo test -p {crate} --lib`; reserve `--workspace` for pre-PR. Avoids cross-crate lock contention and is faster.

## Knowledge
- #4874 — hardened cargo test convention (Tier A)
- #4873 — setsid needs -w to propagate exit codes

dug-21:
Consolidated into #742 — tracking and de-flake/hardening work moves there. Closing as consolidated.

PR #748, opened 2026-06-12
fix: de-flake test-integrity cluster + harden cargo-test process (#742)

### LINKED ISSUE #742: test: consolidate flaky test-integrity cluster — de-flake server lib races + harden cargo-test process (infra-002)
## Summary

Consolidates the open test-integrity / flaky-test cluster into one tracked bugfix: **#738, #710, #705, #712**. These overlap heavily (the `http::token` concurrent-creation flake appears in three of them) and have been re-diagnosed each delivery cycle (vnc-034 Wave 2 #727, crt-052, bugfix #122). Working them once, together, as a single de-flake + process-hardening pass.

### A. De-flake `unimatrix-server` lib races (from #738, #710, #705)
All pass deterministically in isolation; fail only under full-parallel `cargo test --workspace`. Scheduling-dependent latent flakiness, not regressions.

1. **`http::token::tests::test_concurrent_creation_no_corruption`** — concurrent reader observes the token file mid-write (`token file must contain exactly 64 hex characters, found 0`). Race in the concurrent-creation test path / token write. (#738 #710 #705)
2. **`http::listener::tests::test_semaphore_recovery_sequential_connections`** and **`..._after_malformed_http`** — semaphore-recovery timing races under contention. (#705)
3. **`uds::listener::tests::stamp_read::*`** (e.g. `stamp_read_site_a_records_declared`, `stamp_read_site_b_records_declared`, `stamp_read_batch_n_declared_rows`) — async stamp-read ordering races. (#705)
4. **`eval::runner::sweep_tests::test_ac14_correlated_sweep_non_vacuous`** — `bit-for-bit` reproduction assertion over float penalty maps with non-deterministic HashMap iteration/accumulation order (left/right differ ~1e-9 + key ordering). Assertion is too strict for an order-sensitive float computation. (#738)

### B. Harden the cargo-test process (from #712, follow-ups to #122 Tier A / knowledge #4874)
5. **F1** — `/tmp/uni-test.$$.log` is a predictable, symlink-followable path; `>` overwrites a pre-planted symlink target. Harden with `mktemp` / private dir. Apply to the canonical convention in `.claude/rules/rust-workspace.md`.
6. **F2** — `check-cargo-test-convention.sh` CLASS-1 detection filters any line containing the substring `setsid`, so a line that mentions `setsid` AND a second bare `cargo test … | tail` is not flagged (verified false-negative). Tighten the guard regex.
7. **Tier C** — default routine validation to per-crate `cargo test -p {crate} --lib`; reserve `--workspace` for pre-PR. Avoids cross-crate lock contention; faster.

## Scope — OUT (deferred, still tracked here)

- **#712 Tier B — server-side subagent process-group reaping.** Per its own analysis this "needs its own design pass": the `SubagentStop` hook is a sub-50ms sync CLI (ADR-002) and `HookInput` carries no pgid, so inline killing can't be built as a bugfix, and over-broad `pkill -f` would kill sibling swarm agents. Correct design = capture each subagent's pgid at spawn, correlate by `session_id`, reap server-side off the recorded SubagentStop event. **Not in this bugfix** — to be cut as a fresh design issue at close-out.

## Relevant prior knowledge
- #4874 — hardened cargo-test convention (Tier A, #122)
- #4873 — `setsid` needs `-w` to propagate exit codes
- #4724 — fixture emitter+reader race: write atomically (temp + rename)
- #4881 / #4884 / #4882 — fire-and-forget write race; verified recipe converting fixed-sleep assertions to deadline-poll
- #4195 — parallel-run time-window flap; use explicit timestamp offsets
- #3714 — pre-existing full-`--workspace` shared-state flake class

## Reproduction
```bash
# flakes under full parallel/workspace:
setsid -w timeout 600 cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
# each family is stable in isolation, e.g.:
cargo test -p unimatrix-server --lib http::token::tests::test_concurrent_creation_no_corruption
```

Consolidates and supersedes #738, #710, #705, #712.

dug-21:
Consolidates and fixes the open test-integrity / flaky-test cluster (#738, #710, #705, #712 → consolidated into #742, all closed into it).

## What this fixes

**De-flake `unimatrix-server` lib races** (passed in isolation, failed only under full-parallel runs):
- **Item 1 — `http::token` concurrent creation (production race):** the loser of the `O_CREAT|O_EXCL` election read the token file while the winner wrote 64 hex bytes in place (`found 0`). Now the winner writes to a 0600 sibling temp file then `fs::rename`s atomically; the loser's load path has a bounded read-retry. Only the winner writes, so both racers converge on the same token. New `test_concurrent_creation_forced_interleave_converges`. — 25/25
- **Item 2 — `http::listener` semaphore recovery (tests):** deadline-poll for eventual recovery instead of fixed sleeps. — 25/25
- **Item 3 — `uds::listener::stamp_read` (tests):** keep the verified settle, add quiescence-on-top. — 45/45
- **Item 4 / cond.4 — `eval` bit-for-bit (production micro-nondeterminism):** `services/search.rs` summed `total` over HashMap iteration order (non-associative f64 → ~1e-9 drift). Now sums over the source `results_with_scores` Vec in a single pass + `ppr_only_candidates` id tie-break; `test_ac14` reframed to a single materialization. Bit-for-bit assertion preserved verbatim. — 285/285

**Harden the cargo-test process** (follow-ups to #122 Tier A / knowledge #4874):
- **Item 5 (F1):** replace predictable `/tmp/uni-test.$$.log` with `mktemp` across all 11 byte-identical convention copies (preserves `setsid -w` / single-statement / `exit $rc`).
- **Item 6 (F2):** tighten `check-cargo-test-convention.sh` CLASS-1 exemption from substring `setsid` to anchored `setsid[^;|&]*cargo test` (closes a verified false-negative); add a self-test fixture; add `--exclude-dir=worktrees` so `grep -r` doesn't false-fail on a gitignored active worktree.
- **Item 7 (Tier C):** document per-crate `cargo test -p {crate} --lib` as the routine default, reserving hardened `--workspace` for pre-PR.

## Verification
Server lib **4014/0**; other crates green; no new clippy warnings; no `unsafe`; convention scanner + `--self-test` both exit 0; integration smoke **23/0**. Gate: Bugfix Validation **PASS** (6/6).

## Deferred (tracked, not in this PR)
- **#746** — `test_ac14` **cond.1** ~2% non-vacuity flake. Distinct root cause (HNSW approximate-retrieval membership flip, eval-harness layer), gated on an eval-semantics design decision. Was never in the original reports (those were cond.4); surfaced only at n≥285. A prototype final-sort id tie-break was deliberately dropped (fixed no observed flake, silently altered AC-04/FR-08/NFR-03 semantics) — that question is carried in #746.
- **#747** — #712 **Tier B** server-side subagent pgid reaping (needs its own design pass).

Closes #742.

### DIFF
--- a/.claude/agents/ndp/ndp-validator.md
+++ b/.claude/agents/ndp/ndp-validator.md
@@ -78,7 +78,7 @@ Read `.claude/skills/validate/SKILL.md` for the full procedure. Summary:
 cargo build --workspace 2>&1 | grep -A5 "^error" | head -20
 cargo build --workspace 2>&1 | tail -3
 # CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
-setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
+log="$(mktemp -t uni-test.XXXXXX.log)"; setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > "$log" 2>&1; rc=$?; tail -30 "$log"; rm -f "$log"; exit $rc
 ```
 Plus anti-stub scan and deploy.sh integrity check (if deploy.sh was modified).
 
@@ -140,7 +140,7 @@ cargo build --workspace 2>&1 | tail -3
 
 # Tests: hardened workspace run — own process group + hard ceiling + file-not-pipe.
 # CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
-setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
+log="$(mktemp -t uni-test.XXXXXX.log)"; setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > "$log" 2>&1; rc=$?; tail -30 "$log"; rm -f "$log"; exit $rc
 
 # Clippy: first warnings only
 cargo clippy --workspace -- -D warnings 2>&1 | head -30
--- a/.claude/agents/uni/uni-rust-dev.md
+++ b/.claude/agents/uni/uni-rust-dev.md
@@ -146,7 +146,7 @@ Include in your agent report:
 ## Self-Check (Run Before Returning Results)
 
 - [ ] `cargo build --workspace` passes (zero errors)
-- [ ] `cargo test --workspace` passes (no new failures) — run via the hardened convention in `.claude/rules/rust-workspace.md` (`setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc`), never as a bare `cargo test ... | tail` pipe
+- [ ] `cargo test --workspace` passes (no new failures) — run via the hardened convention in `.claude/rules/rust-workspace.md` (`log="$(mktemp -t uni-test.XXXXXX.log)"; setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > "$log" 2>&1; rc=$?; tail -30 "$log"; rm -f "$log"; exit $rc`), never as a bare `cargo test ... | tail` pipe
 - [ ] No `todo!()`, `unimplemented!()`, `TODO`, `FIXME`, or `HACK` in non-test code
 - [ ] All modified files are within the scope defined in the brief
 - [ ] Error handling uses project error type with context, not `.unwrap()` in non-test code
--- a/.claude/agents/uni/uni-scrum-master.md
+++ b/.claude/agents/uni/uni-scrum-master.md
@@ -153,7 +153,7 @@ cargo build --workspace 2>&1 | tail -3
 
 # Test: hardened workspace run — own process group + hard ceiling + file-not-pipe (see rust-workspace.md).
 # CARGO_TEST_TIMEOUT_SECS: hard ceiling so an interrupted run cannot orphan cargo children
-setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > /tmp/uni-test.$$.log 2>&1; rc=$?; tail -30 /tmp/uni-test.$$.log; rm -f /tmp/uni-test.$$.log; exit $rc
+log="$(mktemp -t uni-test.XXXXXX.log)"; setsid -w timeout "${CARGO_TEST_TIMEOUT_SECS:-600}" cargo test --workspace > "$log" 2>&1; rc=$?; tail -30 "$log"; rm -f "$log"; exit $rc
 
 # Clippy: first warnings only
 cargo clippy --workspace -- -D warnings 2>&1 | head -30
--- a/.claude/agents/uni/uni-tester.md
+++ b/.claude/agents/uni/uni-tester.md
@@ -100,7 +100,7 @@ test-plan/
 
 ### What You Do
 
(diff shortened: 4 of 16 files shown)


dug-21:
## Security Review — PR #748 (#742 de-flake + cargo-test hardening)

Fresh-context review. Read the full `origin/main...HEAD` diff (3 commits: c2edbda9, 2f6e73fc, b4593ae7), all source files in full, and ran the scanner self-test + full scan.

**Risk level: LOW. No blocking findings. Recommend merge.**

This change *reduces* attack surface on two fronts (token write, cargo-test temp path). The diff is minimal and matches the stated scope (16 files; the `model_guard.rs`/`test_support.rs` lines that show up against a stale local `main` are from #740, already on `origin/main` — not part of this PR).

### Highest-value target — `http/token.rs` temp+rename: SOUND
- **0600 holds creation→rename (no world-readable window):** temp file created via `create_new(true).mode(0o600)` = `O_CREAT|O_EXCL` at 0600; `fs::rename` preserves the inode's mode, so the published `token` file is 0600 with no intervening chmod window. Verified by `test_concurrent_creation_forced_interleave_converges` (asserts 0600 survives rename).
- **Symlink-attack on the predictable `.token.<pid>.tmp` name: mitigated by `O_EXCL`.** The name is predictable, but `create_temp_token_file` opens with `create_new(true)` → `O_CREAT|O_EXCL`, which POSIX guarantees fails (does not follow) if the path is an existing symlink. A pre-planted symlink causes the create to return `Err` (handled → `ProjectInit` error, no write through the link). The pre-existing `create_token_file` for the final path already had this same profile; no regression, no new exposure.
- **No more-exposed location:** the temp sibling lives in the same private data dir as the final token (which was already 0600 there). No token bytes leak to `/tmp`, stdout, stderr, or logs — `render_first_boot_notice()` is a pure builder that never contains the hex (covered by `test_first_boot_notice_token_absent_and_points_to_retrieval`); load path is `tracing::debug!` only.
- **Loser read-retry is bounded and non-spinning:** 50ms ceiling, 1ms sleep between reads, `Instant::now() >= deadline` break that surfaces the last read to the canonical length error. No infinite loop, no busy-spin. DoS-safe.
- **Stale-temp `let _ = fs::remove_file(&tmp_path)` before create:** worst case a same-PID concurrent writer races, but `O_EXCL` on the subsequent create still rejects collision and errors safely rather than overwriting. Belt-and-suspenders, not a hole.

### `services/search.rs` deterministic sum — SAFE
Pure numerical determinism fix (sum `total` over the source `Vec` instead of `HashMap.values()`, plus an id tie-break before truncate). Value-set is provably identical (unique ids); only ordering is canonicalized. No injection/IO/trust-boundary surface. Blast radius of a subtle bug here is a *retrieval-ranking* change (eval bit-for-bit drift), not data corruption, disclosure, or privilege — and the zero-sum guard (FR-08/FM-05) is preserved. Three new unit tests pin bit-for-bit stability.

### Scanner regex (F2) — CORRECTLY TIGHTENED
`setsid[^;|&]*cargo test` exemption now drops a line ONLY when `setsid` governs that `cargo test` in the same separator-free segment. Verified live: `setsid -w true && cargo test | tail` is now correctly **flagged** (the `&&` stops the `[^;|&]*` anchor) — exactly the GH#742 item-6 false-negative. Self-test (4 cases incl. the new `setsid_substr` fixture) and full scan both exit 0. `--exclude-dir=worktrees` is a correct fix for `grep -r` not honoring `.gitignore` on the active nested worktree.

### mktemp convention (F1) — REMOVES the predictable-path/symlink surface
`mktemp -t uni-test.XXXXXX.log` replaces `/tmp/uni-test.$$.log` (PID-predictable, symlink-followable by a local user) with an `O_EXCL`, 0600, unpredictable name. The load-bearing exit-code guard is intact: `mktemp` is its own prior statement, the capture line is a single non-pipelined `cargo test` so `rc=$?` captures cargo (via `timeout`/`setsid -w`), not `tail` or `mktemp`. 11 copies are byte-identical to the canonical rule.

### Test-only changes (listener/stamp_read/eval) — NO security surface
Deadline-poll-for-observable replacing fixed sleeps; single-materialization eval refactor. All `#[cfg(test)]` / test modules. No production path, no `unsafe`, no new deps, no secrets.

### Non-blocking observations
1. **Scanner residual (informational, not a vuln):** a line of the form `setsid -w cargo test --workspace | tail` (setsid governing cargo test AND piping live, no separator) is exempted by the anchor despite using a pipe. It is not the hardened convention and `setsid` still mitigates the #122 orphan concern, so the gap is narrow. Consider, in a future pass, also requiring the file-redirect (`> "$log"`) rather than exempting purely on `setsid…cargo test` adjacency. Out of scope for this PR.
2. **Minimal-change note:** confirmed the PR diff against `origin/main` is exactly the 16 in-scope files. No unrelated changes rode along.

OWASP sweep (injection / access control / input validation / deserialization / secrets): no findings. Token format validation (length + hex) preserved; no shell/SQL/format-string injection introduced; no hardcoded credentials.

Issue #757, opened 2026-06-15
[flake] cert_provisioner::test_concurrent_first_boot_converges fails under parallel workspace test pressure


dug-21:
## Discovered By

Unit test: `crates/unimatrix-server/tests/cert_provisioner.rs::test_concurrent_first_boot_converges`

## Symptom

During a `cargo test --workspace` run, the test failed:

```
thread 'test_concurrent_first_boot_converges' panicked at crates/unimatrix-server/tests/cert_provisioner.rs:180:33
test result: FAILED. 8 passed; 1 failed; ...
```

The test spawns 4 threads that race to `load_or_generate_cert` against a shared TempDir behind a Barrier, then asserts all threads converge on the same cert/key pair (panic at the `.unwrap_or_else(|e| panic!("thread {i} failed: {e}"))` on line 180).

## Flake confirmation

Re-run in isolation **3/3 PASS**:

```bash
cargo test -p unimatrix-server --test cert_provisioner test_concurrent_first_boot_converges -j 1
# run 1 rc=0 / run 2 rc=0 / run 3 rc=0
```

It only failed once, during a full parallel `--workspace` run while heavy parallel linking/CPU contention was in flight. The concurrent first-boot convergence logic appears to have a race window (or insufficient retry/backoff) that surfaces under resource starvation.

## Expected vs Actual

- Expected: all 4 concurrent boots converge on one cert/key pair.
- Actual: under load, one thread returns `Err` (cert read/generate lost a race) → assertion panic.

## Reproduction

Not deterministic. Surfaces under `cargo test --workspace` with high parallelism. Does not reproduce in isolation.

## Notes

Discovered during bugfix #750 verification. The file is byte-identical to `main` (untouched by #750). Pre-existing flake, not caused by #750. Likely needs a bounded retry/backoff or file-lock hardening in `load_or_generate_cert` (or a serialized test).

Issue #756, opened 2026-06-15
[clippy] workspace-wide collapsible_if errors break cargo clippy -D warnings (let-chain lint)


dug-21:
## Summary

`cargo clippy --workspace -- -D warnings` fails with dozens of `error: this \`if\` statement can be collapsed` (`clippy::collapsible_if`) errors across the workspace. The lint now fires on nested `if let` / `if let` + `if` patterns (collapsible into a `let`-chain) — a behavior of a newer clippy toolchain.

## Affected (sample, NOT exhaustive)

- `crates/unimatrix-engine/src/auth.rs:113`
- `crates/unimatrix-engine/src/event_queue.rs:164`
- `crates/unimatrix-observe/src/metrics.rs` (multiple: 45, 48, 177, 178, 237-239, 248, 322-325)
- `crates/unimatrix-observe/src/source.rs:51-52`
- `crates/unimatrix-observe/src/synthesis.rs:183`
- `crates/unimatrix-observe/src/domain/evaluator.rs:155-156`
- `crates/unimatrix-observe/src/detection/session.rs:338`
- `crates/unimatrix-observe/src/extraction/{shadow.rs:64,dead_knowledge.rs:52}`

`unimatrix-observe` alone reports 52 such errors. The errors are present on `main` and in files unrelated to any current feature/bugfix work.

## Example

```
error: this `if` statement can be collapsed
   --> crates/unimatrix-engine/src/auth.rs:113:5
    |
113 | /     if let Some(pid) = creds.pid {
114 | |         if let Err(e) = verify_process_lineage(pid) {
```

## Why this is its own issue

Discovered during bugfix #750 verification. The #750 change set (session_metrics.rs, mcp/tools.rs, cycle_review_index.rs) is clippy-clean; these errors are pre-existing and workspace-wide. Fixing them inside #750 would be unrelated scope creep. This is a clean, mechanical sweep (`cargo clippy --fix` handles most) that deserves its own PR + review.

## Reproduction

```bash
cargo clippy --workspace -- -D warnings
```

## Notes

Discovered during infra-001 / bugfix #750 verification. Not caused by #750.

Issue #680, opened 2026-06-05
feat(vnc-027): TS UDS hook client + hook-set reduction (F4a)


dug-21:
## Summary

TS UDS hook client (local alternative to Rust `hook.rs`) + hook-set reduction + carry-items. **F4a** — the contractual-attribution half of the former F4 bundle is now **vnc-030 (F4b, #699)**.

Scope doc: `product/features/vnc-027/SCOPE.md` (approved 2026-06-08).
Implementation brief: `product/features/vnc-027/IMPLEMENTATION-BRIEF.md` (design session 2026-06-08).
Acceptance map: `product/features/vnc-027/ACCEPTANCE-MAP.md`.

## Goals

1. **`transport-uds.js`** — Node `net.connect` to `~/.unimatrix/{projectHash}/unimatrix.sock`, 4-byte BE u32 length-prefix framing byte-identical to `wire.rs` (1 MiB cap, zero-length reject). Transport selected by config: remote config → HTTP (unchanged); none → UDS. Event queue shared across transports.
2. **Round-trip parity with the Rust hook over UDS** — same HookRequest frames (extend F3 parity corpus/goldens), same sync semantics, same FNF write-then-disconnect, same fail-open guarantees.
3. **Hook-set reduction** — retire standalone PreToolUse observation (keep `cycle_start`/`cycle_stop` interception); SubagentStop optional (default-off, opt-in settings key). Keep sync-injection + PostToolUse/PostToolUseFailure + lifecycle events.
4. **Carry-items**: (a) C-04 size gate redefinition — comment-stripped ≤100 KB + raw ≤160 KB backstop, lands first; (b) FR-16 offset delete rekey (resolved age-prune-only, ADR-006).

## Non-Goals

- Contractual cycle attribution (stamp, precedence chain, heuristic demotion) → vnc-030 (#699)
- Rust `hook.rs` retirement or behavioral changes → F6 (#682)
- Init unification / installer flow → F5 (#681)
- Client-side PreCompact transcript prepend over UDS (prohibited)
- Windows local mode (UDS is Unix-only)

## Design Session Results (2026-06-08)

Artifacts: `architecture/ARCHITECTURE.md`, `specification/SPECIFICATION.md`, `RISK-TEST-STRATEGY.md`, `ALIGNMENT-REPORT.md` under `product/features/vnc-027/`.

### Resolved decisions (ADRs)

| Decision | ADR |
|----------|-----|
[...]

PR #758, opened 2026-06-15
fix: re-point per-session telemetry onto PostToolUse + guarded recompute (#750)

### LINKED ISSUE #750: bug(retrospective): TS hook client retired PreToolUse → context_cycle_review per-session Calls/Tools/context_reload collapse to 0
## Summary

`context_cycle_review` reports per-session **`Calls: 0`, `Tools: —`, and `0% context reload`** for every cycle delivered under the new **TS UDS hook client** (switched in via `f9397421` / #680, ADR-004 vnc-028 "F6 soak"), despite the cycle carrying full records and a populated cycle-level findings/hotspot section. Per-session telemetry is silently zeroed.

## Root cause

Event-type **contract drift** between the client (write side) and the retrospective aggregation (read side) — not data loss, not a field/casing mismatch (`session_id` is emitted correctly).

- **Write side (the change):** the TS client `buildCycleEventOrFallthrough` (`packages/unimatrix/lib/hook-client/build-request-tools.js:322-327`) returns a `null` no-send sentinel for any non-`context_cycle` `PreToolUse` — it **no longer emits a `PreToolUse` RecordEvent per tool**. ADR-004 (vnc-028) retired that observation deliberately. The old Rust client (`crates/unimatrix-server/src/uds/hook.rs:735,763`) emitted one `PreToolUse` row per tool.
- **Read side (unupdated):** per-session aggregation counts **only** `PreToolUse`:
  - `crates/unimatrix-observe/src/session_metrics.rs` `compute_session_summaries` (tool_distribution loop ~112-121; filters at 160/170/181) — `Calls` = sum of `PreToolUse` tool_distribution.
  - `compute_context_reload_pct` (same file) builds per-session file sets from `PreToolUse`-only records → 0%.
  - `crates/unimatrix-server/src/mcp/tools.rs:4580/4594/4609` — phase tool breakdown, same `PreToolUse` filter.
- Cycle-level findings/hotspots survive because they read **`PostToolUse`** (`post_tool_use_rework_candidate`, normalized server-side), which the TS client still emits (`build-request-tools.js:227-285`), persisted with a populated `tool` (`uds/listener.rs:3086-3090`).

## Discriminator (confirms the client switch is the cause)

| Cycle | Client | Per-session `Calls` | `context_reload` |
|-------|--------|---------------------|------------------|
| vnc-034 (pre-switch) | Rust `hook.rs` | 224 / 45 / 1056 | 26% |
| vnc-035 (post-switch) | TS UDS | 0 / 0 | 0% |

Same aggregation code, opposite result.

## Blast radius

All per-session columns (`Calls`, `Tools`, `Knowledge served/stored/curated`) and the per-session `context_reload`/cold-restart metric, for **every** cycle run under the TS client. Any automation reading `SessionSummary.tool_distribution` or `report.context_reload_pct`. Cycle-level findings/hotspots unaffected. **Gates the F6 soak (#682)**: that gate retires `hook.rs` on the basis of TS-client parity — telemetry must be trustworthy first.

## Fix options (decision needed — ADR-004 interaction)

- **Option A (recommended): re-point the aggregation onto `PostToolUse`.** Update `compute_session_summaries`, `compute_context_reload_pct`, and the `mcp/tools.rs` phase breakdown to count `PostToolUse` (the durable per-tool event under the new client). Keeps ADR-004's reduced observation volume.
  - **Caveat to verify first:** non-rework tools (Read, `context_*`) emit a generic `PostToolUse` only if `tool_name` is in `extra`. Confirm those rows carry a populated `tool` before relying on them, or `Tools`/knowledge columns may stay under-counted (Option A may restore `Calls` fully but `Tools`/knowledge only partially).
- **Option B: TS client re-emits non-cycle `PreToolUse`.** Revert the `null` sentinel. Restores telemetry wholesale but **contradicts ADR-004's retirement intent** and re-adds observation volume — should require a fresh ADR.

Do **not** "fix" by silently reverting the client switch — that reopens a deliberately-closed observation.

## Missing test

A **cross-client telemetry-parity** integration test: feed a representative **TS-client** hook event stream through ingestion, then assert `context_cycle_review` yields non-zero per-session `Calls`, a non-empty `Tools` breakdown, and non-zero `context_reload` on a multi-session re-read. A contract test asserting "every `event_type` the retrospective filters on is still emitted by the active client" would have caught this at #680 merge. Existing tests exercise the Rust client's `PreToolUse` stream only.

## Reproduction

With local hooks on the TS UDS client (post-`f9397421`), run any multi-tool session under a feature cycle, then `context_cycle_review(feature_cycle: "<cycle>", force: true)` → per-session `Calls: 0`, `context_reload: 0%`, while findings populate. Compare to a pre-switch cycle (vnc-034).

Knowledge: lesson #4998. Surfaced during the vnc-035 retrospective.

dug-21:
## Summary

Fixes #750 — under the TS UDS hook client (ADR-004 / vnc-028, #680), non-cycle `PreToolUse` RecordEvents are no longer emitted, but the per-session aggregation read-path still filtered on `event_type == "PreToolUse"`. Result: per-session `Calls`/`Tools`/`Knowledge` and `context_reload` folded to a believable `0` for every TS-client cycle (cycle-level hotspots survived because they already read `PostToolUse`).

## Fix (Option A — confirmed in the issue)

Re-point the per-session aggregation read-path from `PreToolUse` → `PostToolUse`:
- `unimatrix-observe/src/session_metrics.rs` — 6 filter sites (context_reload file-set, tool_distribution, file zones, knowledge served/stored/curated) + doc comments.
- `unimatrix-server/src/mcp/tools.rs` — phase tool/knowledge breakdown (3 sites).
- `SubagentStart` filters, listener ingestion, and cycle-level findings left intact (already correct).

The `context_reload` linchpin holds: Read/Edit/Write `PostToolUse` rows carry `tool` + `input.file_path` unconditionally (live DB: 100%).

## Cache-invalidation merge gate (human-directed)

`context_cycle_review` memoizes into `cycle_review_index`, so the fix alone would leave existing reviews serving the stale zero. `SUMMARY_SCHEMA_VERSION` bumped 3 → 4, **plus** a guarded recompute (amends ADR-002 / crt-033, head #5021):
- Stale `schema_version` → auto-recompute, implemented by clearing `memo_hit` and falling through the existing pipeline (no new write before the presence guards — empty-clobber structurally impossible).
- **Data-presence gate:** recompute only when source observations survive (`!attributed.is_empty()`, already loaded — zero added cost). After a review runs, source data is purgeable; recomputing a purged cycle would compute-from-nothing and overwrite a good summary with an empty one. When purged, the stored summary is **retained byte-identical** with a distinct "stale: source purged, cannot recompute" advisory.
- `force=true` clobber-guard preserved and pinned by regression test.

## Tests

New: believable-zero guard + cross-client parity assertion; guarded-recompute pins (data-present recomputes, data-purged retains untouched, force+purged no-clobber, current-schema serves stored, purged-advisory distinct); inverted `test_session_summaries_filters_posttooluse_only`; `schema_version_is_four`.

Results: session_metrics 46/46, cycle_review 41/41, cycle_review_index 29/29, integration smoke 23/23, cycle_review integration 36/36. Changed files clippy-clean.

## Expected behavior (not regressions)

- **MultiEdit N-row skew:** a single MultiEdit invocation emits N `PostToolUse` rows (one per file pair), so it counts as N in per-session `Calls`/write `tool_distribution`. This is an inherent, bounded property of counting `PostToolUse` and does not affect `context_reload`. Not a counting bug.
- **Pre-switch cycles legitimately read 0:** cycles recorded before the TS-client switch hold `PreToolUse`-only rows, so their per-session `Calls`/`Tools`/`context_reload` correctly read `0` under the new `PostToolUse` aggregation. This is accurate for those historical cycles, not a regression.

## Out of scope (pre-existing, filed)

- #756 — workspace clippy `collapsible_if` failures in files byte-identical to `main`.
- #757 — `cert_provisioner` concurrency flake (3/3 in isolation).

Gate: Bug Fix Validation — **PASS** (9/9).

### DIFF
--- a/crates/unimatrix-observe/src/session_metrics.rs
+++ b/crates/unimatrix-observe/src/session_metrics.rs
@@ -8,9 +8,13 @@ use crate::types::{ObservationRecord, SessionSummary};
 
 /// Compute per-session activity profiles from observation records.
 ///
-/// Groups records by `session_id`, computes tool distribution (PreToolUse only),
+/// Groups records by `session_id`, computes tool distribution (PostToolUse only),
 /// top file zones, agents spawned, and knowledge flow counts. Returns summaries
 /// sorted by `started_at` ascending with lexicographic `session_id` tiebreaker.
+///
+/// Per-tool aggregation reads `PostToolUse` events (the durable per-tool event
+/// under the TS UDS client, ADR-004 / vnc-028). The retired non-cycle `PreToolUse`
+/// event is no longer emitted by the active client (#750).
 pub fn compute_session_summaries(records: &[ObservationRecord]) -> Vec<SessionSummary> {
     // Group records by session_id
     let mut groups: HashMap<&str, Vec<&ObservationRecord>> = HashMap::new();
@@ -48,10 +52,11 @@ pub fn compute_context_reload_pct(
         return 0.0;
     }
 
-    // Build per-session file sets from observation records
+    // Build per-session file sets from observation records.
+    // PostToolUse is the surviving per-tool event under the TS UDS client (#750).
     let mut session_files: HashMap<String, HashSet<String>> = HashMap::new();
     for record in records {
-        if record.event_type != "PreToolUse" {
+        if record.event_type != "PostToolUse" {
             continue;
         }
         let path = record
@@ -109,10 +114,11 @@ fn build_session_summary(
     let max_ts = session_records.iter().map(|r| r.ts).max().unwrap_or(0);
     let duration_secs = (max_ts.saturating_sub(min_ts)) / 1000;
 
-    // Tool distribution: only PreToolUse events (FR-01.2)
+    // Tool distribution: only PostToolUse events (FR-01.2; #750 — PostToolUse is the
+    // durable per-tool event under the TS UDS client, ADR-004 / vnc-028).
     let mut tool_distribution: HashMap<String, u64> = HashMap::new();
     for record in session_records {
-        if record.event_type != "PreToolUse" {
+        if record.event_type != "PostToolUse" {
             continue;
         }
         let tool_name = record.tool.as_deref().unwrap_or("");
@@ -120,10 +126,10 @@ fn build_session_summary(
         *tool_distribution.entry(category.to_string()).or_default() += 1;
     }
 
-    // File zones: only PreToolUse events for file-touching tools
+    // File zones: only PostToolUse events for file-touching tools (#750)
     let mut file_counts: HashMap<String, u64> = HashMap::new();
     for record in session_records {
-        if record.event_type != "PreToolUse" {
+        if record.event_type != "PostToolUse" {
             continue;
         }
         let path = record
@@ -153,11 +159,11 @@ fn build_session_summary(
         }
     }
 
-    // Knowledge flow: PreToolUse events only
+    // Knowledge flow: PostToolUse events only (#750 — surviving per-tool event)
     let knowledge_served = session_records
         .iter()
         .filter(|r| {
-            r.event_type == "PreToolUse"
+            r.event_type == "PostToolUse"
                 && r.tool.as_deref().map(normalize_tool_name).is_some_and(|t| {
                     matches!(t, "context_search" | "context_lookup" | "context_get")
                 })
@@ -167,7 +173,7 @@ fn build_session_summary(
     let knowledge_stored = session_records
         .iter()
         .filter(|r| {
-            r.event_type == "PreToolUse"
+            r.event_type == "PostToolUse"
                 && r.tool
                     .as_deref()
                     .map(normalize_tool_name)
@@ -178,7 +184,7 @@ fn build_session_summary(
     let knowledge_curated = session_records
         .iter()
         .filter(|r| {
-            r.event_type == "PreToolUse"
+            r.event_type == "PostToolUse"
                 && r.tool.as_deref().map(normalize_tool_name).is_some_and(|t| {
                     matches!(
                         t,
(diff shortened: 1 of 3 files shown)


dug-21 (commented):
## Security & Risk Review — PR #758 (#750)

**Fresh-context review.** Read the full diff cold + traced the live handler control flow in `tools.rs`, the single writer, `build_cycle_review_record`, and the UDS listener ingestion arm. **Risk level: LOW. No blocking findings.**

### Data-destruction surface (the core concern) — guard holds

The stated invariant — *recompute-from-purged-data must never overwrite a good stored summary with an empty one* — is **structurally sound**, verified end-to-end:

1. **Single writer.** The only non-test `store_cycle_review()` call is at `tools.rs:2914`, inside the `if memo_hit.is_none()` block (:2269).
2. **Two independent empty-data gates before that writer:**
   - Memo site (:2060/:2080): `Stale { .. } if !attributed.is_empty()` recomputes (clears `memo_hit`); the purged arm (:2080) retains the stored report and serves via memo-hit return — **no writer reached**.
   - Pipeline re-check (:2271): even on fallthrough, `if attributed.is_empty()` routes to cached-MetricVector-or-error **before** the writer. A bug at the memo site alone cannot reach the writer with empty data — the second gate backstops it.
3. **`force=true` clobber-guard (:2134):** runs only `force && attributed.is_empty()`, serves the stored record verbatim and `return`s without writing (:2200). Force+purged cannot clobber.
4. **Recompute is idempotent / self-terminating.** `build_cycle_review_record` (:3803) always stamps `SUMMARY_SCHEMA_VERSION`, so a recompute persists at v4 → next read sees `Staleness::Current` → no recompute loop.

No bypass found across partial-data, race, or force paths. The "NEVER add a second writer here" comment is the load-bearing invariant; it is honored.

### OWASP / boundary checks
- **Injection / deserialization:** no new external input surface. `check_stored_review` deserializes stored `summary_json` and on error already (ADR-003) treats it as cache-miss → no panic, no clobber. Advisory strings are server-built constants with integer interpolation only — no user input reaches them.
- **Access control:** unchanged; the read-path audit event (:2158) is preserved on the purged path.
- **Secrets / deps:** none. `--stat` confirms 3 source files, **zero `Cargo.toml`/`Cargo.lock` changes** — no new dependencies.

### Filter swap (PreToolUse → PostToolUse) — correct & minimal
All 9 read-path sites (6 in `session_metrics.rs`, 3 in `compute_phase_stats`) repointed; `SubagentStart` and listener ingestion correctly left intact. Listener arm (`uds/listener.rs:3085-3097`) extracts `tool`/`input` unconditionally for `PostToolUse` and normalizes `post_tool_use_rework_candidate` → `PostToolUse` (:3138), so rework events fold in without double-counting. New `test_read_path_filter_matches_active_client_event` closes the ADR-004 audit gap that caused the bug.

### Regression risk: LOW–MEDIUM (informational, non-blocking)
- **Mixed/legacy cycles (old Rust hook emitted BOTH events):** switching to PostToolUse-only counts each tool once — correct, not a regression.
- **Edge worth naming:** an *ancient* cycle whose retained raw data is PreToolUse-only (pre-PostToolUse emission) would, on guarded recompute, yield zero PostToolUse rows → a believable-zero again, now persisted at v4 (so it would not re-flag). This is the inverse of the original bug for a bounded historical set. Acceptable given the issue's documented "dogfooding-internal" scope and the believable-zero advisory, but flagging it so it's a conscious acceptance, not a blind spot. **Non-blocking.**

### Verdict
Minimal, well-gated change. The guard genuinely closes the empty-clobber hole with defense-in-depth (two gates + idempotent writer). **Approve from a security/risk standpoint.**

Issue #744, opened 2026-06-12
graph integrity: inbound REDIRECT_CEILING=50 silently orphans referrers 51+ when correcting hot nodes


dug-21:
## Summary
`context_correct`'s incoming-edge redirect (`run_redirect_loop`, `tools.rs:4660`) caps at `REDIRECT_CEILING = 50` (`tools.rs:44`): it truncates-with-`warn!` past 50 and never aborts. Correcting an entry with **>50 inbound referrers** leaves referrers 51+ pointing at the now-**deprecated** entry — an inbound orphan, the mirror of the outbound edge-drop fixed in vnc-035 / #730.

## Why it matters
Inbound degree is corpus-coupled and heavy-tailed: foundational nodes (`vision_root` — every strategic goal `Advances→` it; core ADRs; widely-referenced patterns) accumulate inbound edges as the corpus grows. These hot nodes are exactly the ones most likely to cross 50 — and the worst to silently half-redirect. The `warn!` is observable in logs, but to graph-relational retrieval (architectural principle 4) the orphaned referrers degrade silently.

## Enterprise dimension
The ceiling is a fixed constant tuned for the current single-developer corpus scale. At enterprise / large-corpus scale, hot-node inbound degree grows substantially and a fixed 50 becomes a real correctness limit, not an edge case. The right behavior is a **scale decision that should be made deliberately**, not inherited from a single-user default.

## Options to weigh (design)
- Raise the ceiling / make it configurable (a `RetentionConfig` knob, like the transcript cap).
- Replace truncate-with-warn with a **complete** redirect via batched/paginated work — no silent incompleteness.
- Surface incomplete-redirect to the caller (ack/loud signal) so it is never silent.
- Accept the cap with an explicit, documented contract.

## Relation
- vnc-035 / #730 — fixes the **outbound** mirror of this defect (silent edge drop on correction).
- vnc-017 (#606) — established the incoming redirect, `REDIRECT_CEILING`, `Supersedes` exclusion, warn-and-continue.
- Code: `run_redirect_loop` `tools.rs:4660`; `REDIRECT_CEILING` `tools.rs:44`.

Issue #745, opened 2026-06-12
corpus repair: sweep already-orphaned outgoing edges from pre-vnc-035 corrections


dug-21:
## Context

Follow-up to #730 / vnc-035. That feature flips `context_correct` to **carry outgoing edges forward by default**, fixing the product going forward. It deliberately does **not** repair the existing corpus — that is a distinct concern (data, not product behavior).

## Problem

Entries corrected **before** vnc-035 ships silently dropped their outgoing edges (e.g. goals losing `Advances → vision_root`). Two known goal entries (`personal-cloud`, `proactive-delivery`) were hand-restored on 2026-06-10, but the historical correction chains were never swept systematically — other orphaned outgoing edges may exist.

## Scope to decide

- Identify entries whose pre-vnc-035 correction chains dropped outgoing edges.
- Decide whether a one-shot repair sweep reconstructs them (and from which prior link in the chain), vs. leaving them as accepted historical loss.
- Distinct from any agent-behavior change — vnc-035's response ack (`edges_carried`) already removes the need for agents to manually re-declare.

Blocked on / sequenced after vnc-035 (#730).

dug-21:
Consolidation note (from #743, now closed into this): the **sizing** sub-task — identify and count entries whose pre-vnc-035 correction chains dropped outgoing edges, broken down by category — is **not** blocked on vnc-035 (#730). It can run now, and the count is the cheap input that decides whether the repair sweep is worth building at all (no-action / targeted restore / one-shot sweep).

Only the **sweep itself** should sequence after vnc-035. Suggest treating this issue as two steps: (1) size now → (2) decide + maybe sweep after vnc-035 lands.

Issue #708, opened 2026-06-08
Research: context_get should surface an entry's known relationships (inbound + outbound edges)


dug-21:
## Observation
`context_get` returns an entry's fields (title, content, topic, tags, status, supersession, confidence) but **does not surface the entry's typed graph edges**. To see relationships today you must make a separate `context_graph` (neighbors) call. As a result, relationships are effectively invisible at the point of retrieval — an agent reading an ADR has no signal about what it depends on, conflicts with, or is supported by, and no signal when an entry has *no* edges.

This came out of reviewing crt-052 ADRs: the decisions cite each other and their predecessors richly in prose, but the typed-edge graph is sparse/incidental. Surfacing edges on every `context_get` would (a) make relationships usable where they're actually read, and (b) create the feedback loop that makes the new author-asserted-edge convention (uni-architect / uni-store-adr) self-correcting — a zero-edge entry becomes *visible* on read.

## Loose direction (needs research before scoping)
This is intentionally not specced. Open questions to flesh out:
- **Scope of surfacing** — direct neighbors only (depth 1), or some summarized view? `context_get` is a point lookup; multi-hop probably stays in `context_graph`.
- **What to include per edge** — edge_type, direction, target_id, and target_title? (title needs a join; weigh cost vs. usefulness.)
- **Supersession** — already in `supersedes`/`superseded_by`; avoid double-representing the `Supersedes` edge.
- **Inference vs. authored** — should inferred edges (CoAccess, similarity-derived) be visually distinguished from author-asserted ones? They have very different trust.
- **Response-size / high-degree nodes** — cap + total count? default-on vs. opt-in flag? (Default-on best serves the feedback-loop goal, but needs a bound.)
- **Format variants** — summary (counts), markdown (a 'Related' section), json (edges array).
- **Cross-tool consistency** — align shape with `context_graph` neighbors output so consumers don't learn two schemas.

## Why now / dependencies
Pairs with the author-asserted edge convention just added to `uni-architect` + `uni-store-adr` (Prerequisite/Contradicts/Supports, high bar, default-none). Asserting edges without surfacing them gives no feedback; surfacing without asserting shows empty boxes. Research should confirm the design, then route through the normal design → delivery protocol. Backfill of historical edges is deferred until both ends exist.

dug-21:
## Findings: ass-076 — Surfacing typed graph edges on `context_get`

[...]

PR #764, opened 2026-06-16
[vnc-037] Surface ranked typed graph edges on context_get

### LINKED ISSUE #754: feat(vnc-037): surface an entry's typed graph edges on context_get retrieval
**feat(vnc-037): surface an entry's typed graph edges on `context_get` retrieval — next-hop navigation affordance**

Edges on `context_get` are a **next-hop navigation affordance** — "having read this entry, here are a few genuinely related entries worth pulling next" — not a complete edge dump. Depth-1, both directions, default-on. **Read-path only — no new edge type, no schema migration, no multi-hop.** Ranking is the core: with only 3 display slots, *which 3* is the feature.

## Goals
1. On every `context_get`, surface the top depth-1 typed edges (both directions) by reusing `query_direct_neighbors`, reading live `graph_edges` (no snapshot staleness).
2. Distinguish **author-asserted** edges from **inferred** (co-access/cosine) edges.
3. Render across all three formats (summary/null, markdown, json) with one edge vocabulary aligned to `context_graph`'s `EdgeRecord`.
4. Keep the change surgical: read-path only, no migration; one serializer seam; list-view tools byte-identical.

## Key decisions (revised)
- **D-01 (query strategy):** rank-and-limit IN SQL — `ORDER BY authored DESC, target_confidence DESC LIMIT 3` + separate cheap split `COUNT(*)` for totals. Bounds hub-node fan-out (no pulling up to 1000 rows into memory). Default-on with `include_edges: Option<bool>` opt-out.
- **D-05 (cap):** display cap **3**. Totals stay exact, uncapped, split inbound/outbound, counting symmetric edges **once** (post-canonicalization). Cap is display-only.
- **D-09 (ranking — the core):** authored edges first (Prerequisite/Contradicts/Supports); inferred fills remaining slots only if authored < 3; inferred ranked by **target-entry confidence** (`entries.confidence`, Beta-Binomial composite — "Wilson" is naming shorthand), not edge weight (ass-079: weight is frozen/first-write-wins, a weak ranker).
- **D-10 (symmetric canonicalization — blocker):** `Contradicts`, `CoAccess`, `Informs` are stored two-row (A→B + B→A); collapse to one `↔` edge BEFORE ranking and counting. `→`/`←` direction is meaningful only for asymmetric `Prerequisite`/`Supports`.
- **D-02/03/04/06/07:** unchanged (D-02 minus the direction-semantics fix). Per-edge payload `{edge_type, direction, target_id, target_title, authored}`; `authored = source=='agent'`; Supersedes excluded; explicit zero-state; `None ⇒ key absent` serializer seam.
- **D-08 (renderings):** cap-3 + `↔`. summary digest shows true split (e.g. `edges: 5↑ 2↓ ↔3 (2 authored)`); markdown `### Related` shows the ranked 3 with `…and N more — use context_graph`.

## Acceptance Criteria
AC-01…AC-12 — see SCOPE.md. Adds AC-12 (per-call latency budget for the default-on path). AC-05/AC-08 updated for cap-3 + ranking + canonicalization. AC-10 tests: symmetric canonicalization, authored-priority-under-cap, inferred-fill-only-when-authored<3, ranking-by-target-confidence, opt-out, high-degree hits SQL LIMIT not memory, carried-forward/`context_edge` classify authored.

## Non-Goals
Multi-hop/traversal (stays in `context_graph`); new edge type or migration; re-representing supersession; backfilling historical edges (#708); changing edge-assertion conventions; edges on search/lookup/store/correct (byte-identical).

## Open Questions (design session)
- OQ-03 — should internal callers (hook path, briefing pipeline, by-ID loop fetches) default `include_edges:false`?
- AC-12 latency budget numbers (proposed ≤5ms p50 / ≤15ms p95) — confirm against a real edge-free baseline.

## Dependencies
ass-076 FINDINGS (#708); ass-079 (frozen-Informs-weight rationale, D-09); author-asserted-edge convention; vnc-035 carry-forward (#749); `context_graph` `EdgeRecord` (ADR #4478) + Supersedes exclusion (ADR #4461); relates to #744/#745.

Scope doc: `product/features/vnc-037/SCOPE.md`

dug-21:
## vnc-037 — Surface an entry's typed graph edges on `context_get` (next-hop navigation affordance)

Closes #754.

`context_get` now surfaces a **ranked, capped (≤3)** set of an entry's depth-1 typed graph edges as a *next-hop navigation affordance* — "having read this, here are a few genuinely related entries worth pulling next" — with honest **uncapped** direction-split totals. Read-path only: **no new edge type, no schema migration, no multi-hop.**

### What it does
- **Ranked next-hop list (D-09).** Authored edges (`Prerequisite`/`Contradicts`/`Supports`) fill slots first; inferred fills the remainder only if authored < 3, ranked by **target-entry `entries.confidence`** (not edge weight — frozen/non-discriminating per ass-079). Locked SQL: `ORDER BY (source='agent') DESC, t.confidence DESC NULLS LAST, target_id ASC LIMIT ?`.
- **Symmetric canonicalization (D-10).** `Contradicts`/`CoAccess`/`Informs` (stored as two reciprocal rows) collapse to one `↔` edge **in SQL, before ranking and counting** — on both the displayed set and the totals.
- **3-bucket totals.** `edge_totals {inbound, outbound, both}` — uncapped, symmetric counted once; a `↔` edge increments `both`, never `inbound` (keeps asymmetric-inbound degree honest for #744 observability).
- **Three renderings.** summary digest `… | edges: {out}↑ {in}↓ ↔{both} ({K} authored)`; markdown `### Related` flat ranked ≤3 with `…and N more — use context_graph`; json `edges[]` + 3-key `edge_totals`.
- **Default-on with opt-out.** `GetParams.include_edges: Option<bool>` — `None`/`Some(true)` surface, `Some(false)` skips all edge queries. Internal callers (hook/briefing/by-ID loops) bypass the handler entirely and are structurally edge-free.
- **Fail-loud (FR-19, OQ-A resolved).** A post-primary-read edge/count/title failure fails the whole `context_get` via the primary-read error mapping — never degrade, never silent-omit (a silent omit is indistinguishable from a true zero-edge entry and would poison the signal).
- **Cap is one named constant** `GET_EDGE_DISPLAY_LIMIT` (=3) — retuning is a one-line edit; tests reference the constant.

### Surgical blast radius
- `entry_to_json` / `format_entry_markdown_section` signatures unchanged; `None ⇒ edges key absent` → `search`/`lookup`/`store`/`correct` payloads byte-identical (verified via the real producer).
- `context_graph` neighbors contract unchanged — the additive `source` column is the only net-new read cost; existing neighbors suite passes unedited.
- New files: `graph_queries_ranked.rs` (store), `mcp/get_edges.rs`, `mcp/response/edges.rs` + `edges_render.rs` (server). No migration.

### Validation
- Gates **3a / 3b / 3c all PASS**.
- Unit: store 389/0, server 4184/0. Integration smoke 24/0; new `test_get_edges.py` 17/17; neighbors suite unedited.
- Critical risks proven: symmetric canonicalization (display + totals), discriminating ranking (#3886 proof-outside-cap + weight≠confidence), #744 `both`-never-`inbound`, FR-19 fail-loud RED tests, list-view byte-identity (#1268).

### Open / carry-forward
- **AC-12 latency** — provisional ≤5ms p50 / ≤15ms p95 **not number-locked**: no high-degree edge-free baseline measurable in this environment. The hub bound (rank-and-limit-in-SQL → 3 rows + scalar counts) is proven; only the measured-number lock is deferred (spec C-9 — human decision).
- ADR-005 `context_correct` re-sync deferred until the Unimatrix MCP reconnects (was disconnected during delivery).


dug-21 (commented):
## Security Review — vnc-037 (uni-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings.** Read the full diff cold against OWASP injection / access-control / input-validation / deserialization, blast radius, regression, and secrets.

### Injection — clean (the primary concern)
- **Ranked select + split count** (`graph_queries_ranked.rs`): one shared `CANON_CTE` const; anchor bound positionally (`?1`), cap bound positionally (`?2` ← `GET_EDGE_DISPLAY_LIMIT`, never a literal 3). The symmetric-type `IN ('Contradicts','CoAccess','Informs')` set, the `!= 'Supersedes'` filter, the canonicalization `CASE`, `ORDER BY`, and `LIMIT` are all static SQL — no input concatenation.
- **Batched title join** (`get_edges.rs::fetch_titles_batch`): `WHERE id IN (?, …)` placeholders generated from arity only (`repeat_n("?", ids.len())`); each id bound positionally via `query.bind(id as i64)`. No id is ever string-interpolated. Bounded to ≤cap ids, so no unbounded IN-list.
- **Neighbors SELECTs** (`graph_queries_neighbors.rs`): the added `source` column is a static column name; the `IN (…)` edge-type lists remain `?N` positional placeholders (unchanged pattern).

### Input validation & deserialization
- New param `include_edges: Option<bool>` with `#[serde(default)]` — three-state (None/Some(true)/Some(false)) resolved in the handler; no new free-text input crosses a boundary. `validate_get_params` unchanged (field is additive). Backward-compatible: omitted ⇒ None ⇒ default-on.
- `try_get` on `source` is safe — `graph_edges.source` is `NOT NULL DEFAULT ''`. `entries.confidence` handled as `Option<f64>` with `NULLS LAST` (LEFT-JOIN miss / dangling target ⇒ None, ranked last). No `.unwrap()`/`.expect()` on any read path; every sqlx/try_get error maps to `StoreError::Database`.

### Access control
- No new trust boundary. The store is single-project scoped (1-client:1-project); edge queries read only `graph_edges`/`entries` within the same store the primary `entry_store.get` already authorized. No cross-project surface introduced.

### Blast radius (worst case)
- Read-path only on the hottest read tool (`context_get`). Fail-loud (FR-19/C-13): a post-primary-read edge/count/title error is mapped with the **same** mapping as the primary read and returned — no degrade-with-note, no silent omission. Worst realistic case is a `context_get` returning an error instead of partial data — safe (no corruption, no disclosure, no DoS: queries are anchored + capped, totals are a single 4-scalar aggregate, no hub fan-out materialized).
- Two non-data-affecting fallbacks (`map_direction` → `both`, `render_json_edges` → `[]`) are defensive no-panic paths over a closed SQL-emitted direction set; benign.

### Regression risk — contained
- **Byte-identity invariant honored.** `format_single_entry` gained `edges: Option<&EdgesView>`; `None` is structural (key never inserted / section never appended / digest never emitted). The other production caller (lookup-by-id, tools.rs:647) passes `None` ⇒ unchanged output. Test `test_none_json_byte_identical_to_base_object` pins this through the real serializer path.
- **Neighbors contract intact.** `RawEdgeRow` gained two additive public fields; the plain path sets `target_confidence: None` and `query_ranked_neighbors` lives beside (never mutates) `query_direct_neighbors`. Only two construction sites exist, both in-crate and updated. context_graph consumers don't read the new fields.
- **#744 inbound-degree integrity** preserved — `↔` counted once in `both`, never folded into `inbound`.

### Dependencies & secrets
- No new dependencies (workspace-only: `sqlx`, `serde`, `std`). No hardcoded credentials/keys/tokens in the diff. Python harness change is additive opt-in plumbing.

**Recommendation: APPROVE for security.** No changes requested.

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. Low-risk, surgical, vision-advancing read-path addition; two clearly-bounded carry-forwards (AC-12 number-lock, ADR-005 re-sync) become follow-up issues, not merge blockers.

> Independence note: spawn prompt was clean (IDs + artifact paths only, no conclusions). Formed this from PRODUCT-VISION.md, the goal:self-learning issue landscape, the artifacts, and the diff directly. Unimatrix MCP was disconnected this session (per the PR and every agent report); orientation used gh + vision file. The goal-entry content lookup could not be run — flagging so the human weighs that the vision-fit read below is from the vision doc, not the live goal entries.

### Vision / roadmap fit
Strong fit, and notably *cheap* fit. vnc-037 is the **surface** half of the assert-and-surface loop under `goal:self-learning`: the authoring convention (assert edges) shipped earlier, and asserting edges nobody ever sees gives authors no feedback. Surfacing depth-1 typed edges on `context_get` — the hottest read — closes that loop exactly where knowledge is consumed. This is the right sequencing: surface follows assert, and backfill of historical edges (#708/#743/#745) stays correctly deferred until both ends exist.

It also lands cleanly against the architectural principles. Reading live `graph_edges` via SQL (not the in-memory snapshot) is the right call for a point read — no snapshot staleness, and it sidesteps the in-memory-hot-path principle without violating it (that principle governs analytics-derived search data, not a single anchored point lookup). The `↔`-counted-once-in-`both` totals deliberately keep asymmetric inbound degree honest for #744 observability — this feature makes the edge-loss cluster's effect *visible* rather than papering over it, which is the right relationship to in-flight integrity work.

On roadmap constraint: the design is additive and does not foreclose anything upcoming. Multi-hop stays in `context_graph`; no new edge type, no migration. #744 (REDIRECT_CEILING orphaning) and #745 (orphan sweep) remain independent and are *aided*, not blocked, by honest uncapped totals being surfaced.

One vision-lens caution worth the human's eye (not a blocker): the value of this feature is entirely a function of **ranking quality** with only 3 slots. The team correctly identified "which 3 is the whole feature" and grounded the ranker in ass-079 (edge weight is frozen/first-write-wins, a weak ranker → rank by target-entry confidence instead). That is a sound, evidence-backed call. But it means the affordance's real-world usefulness is unproven until there is a corpus with enough authored edges and confidence spread to discriminate. That is a *post-merge learning signal to watch*, not a reason to hold the PR.

### Approach commentary
Strengths:
- **Surgical blast radius is real, not asserted.** I confirmed in the diff that `format_single_entry` is the only signature change (`edges: Option<&EdgesView>`), the other production caller (lookup-by-id, tools.rs:644) passes `None`, and `None ⇒ key/section/digest absent` is structural (not a guarded omit). List-view byte-identity is pinned through the real producer (#1268). This is the right seam.
- **Fail-loud is correct and verified.** The handler maps a post-primary-read edge/count/title error with the *identical* `ServerError::Core(CoreError::Store(e))` mapping as the primary read and propagates via `?`. A silent edge omission would be indistinguishable from a true zero-edge entry and would poison the very feedback signal the feature exists to create — failing loud is the right product choice, not just the safe one.
- **Hub bound is structural, not hoped-for.** Rank-and-limit-in-SQL (`ORDER BY (source='agent') DESC, t.confidence DESC NULLS LAST, target_id ASC LIMIT ?`) plus a separate scalar count means a hub node never materializes its fan-out. Proven at the store boundary and the MCP boundary (50 edges → 3 displayed).
- **OQ-03 resolved by construction**, which is the elegant answer: internal callers (hook/briefing/by-ID loops) call `entry_store.get` directly and never reach the MCP handler, so they pay zero edge cost without needing an explicit `include_edges:false`. The agent-facing tool stays default-on with a single clean code path.

Concerns worth weighing (none blocking):
- **AC-12 latency is an honest OPEN, not a pass.** The ≤5ms p50 / ≤15ms p95 numbers are provisional — no representative high-degree edge-free baseline was measurable in this environment. The *mechanism* that makes the budget reachable is proven; only the measured-number lock is deferred. Validators recorded this transparently rather than rubber-stamping. The escalation path (relax budget / mandate opt-out / revisit default-on) is specified. Good discipline; the human just needs to own the deferral consciously.
- **ADR-005 `context_correct` re-sync is deferred** because the MCP was disconnected during delivery. This is the knowledge-stewardship debt, not a code defect. Should be tracked so the architectural decision record actually lands in the store once reconnected.
- **AC-13b** (runtime cap override) is a structural impossibility for a compile-time `const`, not a coverage gap — the load-bearing single-source half is fully proven. I agree with the validator's ruling.

The three SCOPE open questions were all resolved in design; recording my concurrence plus the two live carry-forwards.

- **OQ-01 (JSON totals shape)** — Resolved as a 3-key object `edge_totals {inbound, outbound, both}`. Concur. The object (vs scalar keys) is more extensible and the 3-bucket split is what keeps `↔`-once + #744 inbound honesty expressible. Right call.
- **OQ-02 (summary digest glyph form)** — Resolved to `… | edges: {out}↑ {in}↓ ↔{both} ({K} authored)` with authored counted over the full set. Concur — the digest tells the truth about the whole edge set, not the displayed 3, which is what an agent needs to decide whether to call `context_graph`.
- **OQ-03 (internal-caller default)** — Resolved structurally (internal callers bypass the handler). Concur; this is better than a default-`false` flag because it keeps one default-on path and removes the latency tax without a config surface to misset.
- **LIVE — AC-12 latency number-lock.** Recommended answer: **keep default-on, do not relax the budget pre-emptively.** Run the edge-free vs default-on baseline on a representative store with a real hub node post-merge; only if it breaches should the human pick from the C-9 menu. The hub bound is proven, so default-on is the right starting posture and reversing it later is a one-param change.
- **LIVE — ADR-005 re-sync.** Recommended answer: file a tracking item to run `context_correct` for ADR-005 once the MCP reconnects, so the decision record's provenance is intact (per the project's context_correct-for-updates rule).

### Recommended actions
1. **Merge PR #764 as-is.** Gates 3a/3b/3c PASS, security LOW with no blocking findings, CI green (the two pending Windows hook-client jobs are unrelated to this Rust read-path change). I independently confirmed the load-bearing seam, fail-loud mapping, and OQ-03-by-construction in the diff.
2. **File a follow-up issue: AC-12 latency baseline + number-lock** — run edge-free vs default-on on a high-degree store; decide per C-9 only if breached. Label `goal:self-learning`, link #754.
3. **File a follow-up issue (or checklist item): ADR-005 `context_correct` re-sync** when MCP reconnects — closes the stewardship gap noted in the PR.
4. **Watch post-merge as a learning signal (no issue needed yet):** does the ranked-3 affordance actually drive useful next-hop reads, given current authored-edge density and confidence spread? This is the real test of the feature's value and feeds the self-learning goal. Revisit at retro.
5. **Retro candidate (already flagged by the tester):** the integration-seeding gotcha — MCP semantic store-dedup collapses near-identical seeded entries, so seed distinct target rows via direct SQL. Promote to a stored pattern if a second feature hits it.

Issue #766, opened 2026-06-17
bug(vnc-034): bundle init posts Ping/observe to /v1/observe → 404; the "one bundle, one command" attach path is non-functional


dug-21:
## Summary
`npx @dug-21/unimatrix init --bundle <bundle>` — the headline personal-cloud attach flow — fails validation with `Remote validation failed: server returned HTTP 404`. The bundle decodes fine and the pinned TLS handshake succeeds; only the request *path* is wrong.

## Repro
1. Container up with `UNIMATRIX_PUBLIC_URL=https://unimatrix:8443`, default project (no slug registered).
2. `unimatrix client-bundle` → bundle.
3. `npx @dug-21/unimatrix init --bundle 'unimatrix-bundle:…'` → `HTTP 404`; "Configuration files were written; fix the URL/token and re-run init."

## Root cause
- Client resolves `remote = base_url + "/v1"` for the no-slug case (`packages/unimatrix/lib/init.js:307`).
- The Ping (and every hook event) posts to `{remote}/observe` (`packages/unimatrix/lib/hook-client/transport-http.js:84`) → **`/v1/observe`**.
- Server `/observe` is a **top-level** route, split off before slug routing (`crates/unimatrix-server/src/http/router.rs:188`; `http/router/seam.rs:177`). `/v1/observe` matches no route → `observe` parsed as a reserved slug → 404.

The bundle path conflates the **MCP base** (`/v1/{slug|tools}`) with the **observe base** (root `/observe`).

## Suspected wider blast radius
This likely breaks *runtime* remote telemetry too — every hook would POST to `/v1/observe` — which would defeat the goal's "full pipeline fidelity over HTTPS" claim. **Needs verification as part of scoping** (may warrant splitting into a separate confirmation task).

## Workaround (validated)
Legacy `--remote <base-url> --token <token>` (+ `NODE_TLS_REJECT_UNAUTHORIZED=0` for the self-signed cert) posts to top-level `/observe` and succeeds.

## Solution directions — APPROACH OPEN, choose by blast radius
Not prescriptive; each touches different wiring:
- **(a) Client-only:** derive the observe URL from `base_url` root (`{base_url}/observe`), independent of the `/v1` MCP remote. Lowest blast radius.
- **(b) Server-side:** also route `/v1/{slug}/observe` + `/v1/tools/observe` (per-slug observe). Larger; changes the routing contract but may be the more correct multi-project model.
[...]

Issue #771, opened 2026-06-17
[vnc-038] parity-layer2-concurrency.test.js: startRealServer times out — serve --foreground no longer exposes top-level HTTP /observe


dug-21:
## Discovered By

JS integration test (Node `node --test`, run via `npm test`):
`packages/unimatrix/test/hook-client/parity-layer2-concurrency.test.js`
(both `test_l2_concurrency_attribution` and `test_l2_raw_session_id_on_wire_server_mints_http_prefix`)

via the shared helper `packages/unimatrix/test/helpers/real-server.js` (`startRealServer`).

## Root Cause (vnc-038 deployment-model change)

`real-server.js` spawns `unimatrix serve --foreground` and then `waitForServer()` polls **`POST /observe`** (the top-level HTTP observe route) until it answers, to detect readiness.

vnc-038 (ADR-003/ADR-004):
1. **Deleted the top-level `/observe` route** — observe is now `POST /v1/{slug}/observe` only.
2. **Gated the HTTP/cloud listener on registered `[[projects]]`.** With no project registered, `serve --foreground` boots as the LOCAL path-hash daemon (UDS + MCP-UDS listeners only, "daemon ready") and binds **no HTTP TCP listener** at all (confirmed: boot log shows only UDS listeners, no HTTP bind).

So `waitForServer`'s TCP `POST /observe` poll never connects → the `before` hook hits its 60s `startTimeoutMs` → both tests fail.

## Expected vs Actual

- Expected: `startRealServer` brings up a server whose observe endpoint answers; the two Layer-2 byte-isolation tests run.
- Actual: `before` hook times out at 60s; both tests fail. Reproduces in isolation (`node --test test/hook-client/parity-layer2-concurrency.test.js`), not contention-related.

## Scope / Triage

- This is a vnc-038-CAUSED regression, but in **vnc-026/025/027 Layer-2 test infrastructure** that is OUTSIDE the vnc-038 modification scope (real-server.js / the Layer 2 suite are not in the vnc-038 brief's Files-to-Modify; not in the vnc-038 diff).
- The fix belongs to whoever owns the Layer-2 harness reconciliation: `real-server.js` must register a project (or run the cloud HTTP mode with a slug) and probe the new `/v1/{slug}/observe` readiness route — OR the suite must move to the local UDS transport it already uses elsewhere.
[...]

PR #772, opened 2026-06-17
[vnc-038] Mandatory project identity at the deployment entrypoint (closes #766, #735, #771)

dug-21:
## Summary
Revises the vnc-034 personal-cloud deployment contract so a project identity is **mandatory and uniform** at the entrypoint, and closes the #766 bug class by construction.

- **Removes the no-slug / default-project route** (hard cutover, cloud/container-HTTP only). A fresh deployment serves nothing until a project is registered — loud "register a project to begin". `DefaultResolver` + `/v1/tools→Default` + `_=>Default` deleted; single project = N=1 through the unified slug-keyed resolver.
- **Uniform one-command setup**: `unimatrix project register <slug>` now writes the `[[projects]]` routing intent atomically (temp+fsync+rename, re-attach-safe); restart applies it. Identical for project 1 and project N.
- **Server-composed `v:2` bundle**: carries fully-formed `mcp_url` + `observe_url` (not a bare slug). Atomic Rust-encoder / JS-decoder + regenerated parity corpus.
- **Dumb-client invariant**: the client composes no paths and derives no identity — it posts the server-authored URLs verbatim. This forecloses the #766 cause (client building `/v1/observe` from its own grammar).
- **Per-slug observe on the per-request funnel** (`/v1/{slug}/observe`), same resolver as MCP — no boot-bound single store.
- **Local STDIO/UDS unaffected**: keeps its direct path-hash binding, never routed through the resolver (verified guard).
- **Folded in #735**: token never to stdout (already-redacted, now regression-guarded), `router.rs` ≤500 (422), stale `dead_code` allow removed.

## Decisions
8 ADRs (#5080–#5088): dumb-client invariant, v:2 server-composed bundle, per-slug observe funnel, delete-default/unified resolver, reserved-slug re-derivation, local-UDS direct binding, register-writes-routing-intent, token-via-bundle-only.

## Validation — all gates PASS
- **Gate 3a/3b/3c: PASS.** Lib 4215 pass; Rust integration `project_routing_integration` 10/10 (three Default-arm tests **inverted to loud-404**, not deleted), `client_bundle_e2e` 4/4 (v:2); smoke 24; protocol+security 35; JS owned surface all green.
- **N=2 cross-store isolation proof (GATE-4): GREEN** for both MCP and observe (recording resolver, sequence `[alpha, beta]`; N=1 rejected per #4974).
- **#766 closure**: bundle → init → Ping/observe 200 over the real per-slug route; runtime hook posts `observe_url` verbatim.

## Closes
- #766 (the bundle-init `/v1/observe` 404)
- #735 (token-to-stdout + Wave-1 cleanups, folded in)
- #771 (Layer-2 harness readiness probe — reconciled to the per-slug model in-diff)

## Follow-up (out of scope, non-blocking)
Env-HTTPS spawned clients have no cert-pin plumbing in `config.js` (the #771 test fix used a test-only TLS override). Filed separately as a potential client gap if self-signed cloud cert verification via env config is desired.


dug-21 (commented):
## Security Review — vnc-038 (fresh-context)

Reviewed the full diff cold against OWASP concerns, blast radius, regression, dependencies, and secrets. The security-critical surfaces are well-constructed:

**Strengths (verified):**
- **Slug input (injection / path traversal):** `ProjectSlug::try_from` enforces `^[a-z0-9][a-z0-9-]{0,62}$` at the parse edge before any filesystem or TOML use. `..`, `/`, `\`, `%2f`, `%2e`, whitespace, uppercase cannot pass the charset — traversal and TOML injection are *unrepresentable*, not runtime-rejected. Validation is centralized in one `TryFrom`-only newtype.
- **Cross-project store isolation (the integrity stake):** observe now resolves per-request through the **same** `resolve_store(parse_project_key(path))` funnel as MCP (`handlers.rs::route_observe`) — the boot-bound observe store handle is deleted. `ProjectKey` is now a single-variant `Slug(_)` enum; the `Default` variant + resolver `default` field + every `Default` arm are gone, so a default-store fall-through can't be constructed. Identity is transport-derived (URL path only), never from a payload. A request bound to B cannot reach A's store.
- **`v:2` bundle decode (deserialization trust boundary):** Rust + JS decoders enforce identical guard ordering — 4 KB raw length cap FIRST (DoS pre-filter), scheme, base64url round-trip, JSON, then strict schema (exactly 5 keys, `v==2`, both URLs `https://`-only, token/fp regex). `v:1` fails closed with an actionable re-issue message; no compat arm; token never in any error message; JS decoder zero-dependency. `https://`-only blocks SSRF/downgrade.
- **Config write (TOML injection / atomicity):** `config_write.rs` does read-modify-write of the parsed TOML doc serialized via the `toml` lib (no string concat → no injection), then temp + `fsync` + atomic `rename`. Idempotent; malformed config is a loud error, never clobbered. Genesis-clobber prevented (State B opens, State C genesis only when `!data_exists`).
- **First-boot token (secret exposure):** `render_first_boot_notice` is a static string with no token and no token-file path; `client-bundle` keeps the token only inside the base64url blob. Golden fixtures use placeholder tokens.
- **Test-only TLS override:** `NODE_TLS_REJECT_UNAUTHORIZED=0` appears only in two `test/` harness files (gated on HTTPS-origin spawns), never under `lib/`. No production TLS relaxation introduced.
- **Dependencies:** no manifest changes — `Cargo.toml`/`Cargo.lock`/`package.json`/`package-lock.json` unchanged. No new transitive advisories.

**Regression:** hard cutover is bounded to the cloud HTTP surface (zero current deployments); local STDIO/UDS boot binding is untouched (the only main.rs reference is a comment). `v:1` holders fail closed with a re-issue message. Low, contained.

**Non-blocking notes:**
- Legacy `{remote, token}` init path still composes `/observe` client-side and accepts `http:` — pre-existing, isolated to the legacy branch, not the #766 surface. Track for eventual retirement.
- Config-write temp file is PID-only named; atomic rename is the correctness mechanism so this is fine.

Full report: `product/features/vnc-038/agents/vnc-038-security-reviewer-report.md`

PR #778, opened 2026-06-18
fix(http): wire rmcp allowed_hosts from UNIMATRIX_PUBLIC_URL (#774)

### LINKED ISSUE #774: bug(personal-cloud): rmcp allowed_hosts never wired from UNIMATRIX_PUBLIC_URL — direct HTTPS by public hostname 403s at the host gate
## Summary
A remote client connecting to the cloud over direct container HTTPS by its configured public hostname is rejected with **`Forbidden: Host header is not allowed`** *before* auth or MCP run. This is rmcp's Streamable-HTTP host-allowlist (CVE-2026-42559 defense), which defaults to **localhost only**. The operator's `UNIMATRIX_PUBLIC_URL` host is **never wired** into that allowlist, so the only OSS-supported remote path (direct container HTTPS — proxy/K8s TLS termination is explicitly enterprise/out-of-scope) is structurally broken for any host other than `localhost`.

## Reproduction (confirmed)
Deployment with `UNIMATRIX_PUBLIC_URL=https://unimatrix:8443`, slug `arch-research`, bundle-issued token.

Real hostname → 403:
```
curl -sk -X POST https://unimatrix:8443/v1/arch-research -H "Authorization: Bearer <token>" -d '{...initialize...}'
HTTP/1.1 403 Forbidden
Forbidden: Host header is not allowed
```
Same request with `-H "Host: localhost"` → **200 + valid MCP `initialize` result.** TLS, routing, auth, and MCP all work — the host gate is the sole blocker.

## Root cause
`public_url.rs` derives a single `PublicUrl` from `UNIMATRIX_PUBLIC_URL` and documents **three** consumers (module doc lines 6-9, field doc line 77): cert SAN, bundle `base_url`, and the `allowed_hosts` default. Only the first two are wired:
- Cert SAN ✅ (TLS validates — failure was 403, never a cert error)
- Bundle `base_url` ✅ (`mcp_url` correctly carries the public host)
- `allowed_hosts` ❌ **phantom consumer** — `McpAdapter::new` (`crates/unimatrix-server/src/http/router.rs:326-336`) constructs `StreamableHttpServerConfig::default()`, sets only `allowed_origins`, and leaves `allowed_hosts` at rmcp's localhost-only default. The comment at router.rs:324-325 states it explicitly.

Confirmed: **zero** non-test assignments to `allowed_hosts` exist anywhere in `crates/unimatrix-server/src`. The R-09 "single derivation, three consumers" invariant holds for only 2 of 3.

## Impact
- Breaks the documented personal-cloud happy path ("one container, one command", direct HTTPS by the configured public host). Every remote `context_*` call 403s.
- This is the *immediate* blocker; it sits in front of a separate, larger gap (remote `init` wires no MCP server / no JS TLS-owning bridge — to be filed as the design follow-up). Even a correct bridge 403s until this is fixed.

## Proposed fix
Wire `PublicUrl.sans` (already includes the local SANs + the configured public host) into `StreamableHttpServerConfig.allowed_hosts` at `McpAdapter::new`, alongside the existing `allowed_origins`. Preserves the CVE-2026-42559 allowlist posture; adds only the legitimately-configured operator host. No new config surface — derives from the `UNIMATRIX_PUBLIC_URL` already set in compose.

## Open questions (triage)
1. **Port matching** — rmcp's host comparison vs `Host: unimatrix:8443` (with port). Confirm whether allowlist entries need port-stripping/normalization so `unimatrix` matches `unimatrix:8443`.
2. **Docker service name vs public domain** — on a compose network clients connect by the **service name** (`unimatrix`); if an operator sets `UNIMATRIX_PUBLIC_URL` to an external domain but connects internally by service name, the Host still won't match. Should `allowed_hosts` accept multiple names (service name + public domain), or is single-host the contract?
3. Should this also cover the legacy env-HTTPS path, or is the bundle the only supported cloud-attach (cf. #773)?

## Refs
- `crates/unimatrix-server/src/http/router.rs:324-336`
- `crates/unimatrix-server/src/http/public_url.rs:1-17,77`
- Related: #773 (env-HTTPS cert-pin), #768 (stale remote-mode docs)

dug-21:
## Summary
Wires rmcp's Streamable-HTTP `allowed_hosts` host-allowlist from `UNIMATRIX_PUBLIC_URL`, fixing `403 Forbidden: Host header is not allowed` for remote clients connecting over direct container HTTPS by their configured public hostname.

Fixes #774.

## Root cause
`McpAdapter::new` set `config.allowed_origins` but left `config.allowed_hosts` at rmcp's localhost-only default. `PublicUrl.sans` — which already contains the configured public host plus the local SANs — was available at the wiring site but dropped, because the constructor chain (`from_servers` → `from_server` → `McpAdapter::new`) only threaded `allowed_origins`. This left the R-09 "single derivation, three consumers" invariant satisfied for only 2 of 3 consumers (cert SAN ✅, bundle base_url ✅, allowed_hosts ❌).

## Fix
- Thread `allowed_hosts: Vec<String>` (= `public_url.sans`) alongside `allowed_origins` through the constructor chain, mirroring the vnc-023/ADR-002 additive-param pattern; set `config.allowed_hosts` in `McpAdapter::new`.
- Add `debug_assert!(!allowed_hosts.is_empty())` at the `main.rs` wiring site — rmcp treats an empty `allowed_hosts` as allow-all (fail-open), and `allowed_hosts`/`allowed_origins` share constructor params with opposite empty-semantics, so the guard prevents a future refactor silently failing open.
- Rewrite the now-stale `router.rs` comment to document the wiring and the empty=fail-open hazard.

`.sans` is structurally non-empty (≥3 local SANs even when `UNIMATRIX_PUBLIC_URL` is unset), so the unset case stays localhost-only and the CVE-2026-42559 posture is preserved. Single-host contract, consistent with cert SAN + bundle base_url.

## Tests
- `test_adapter_allows_configured_public_host_rejects_others` — host-gate regression: configured public host clears the gate, unrelated host 403s.
- `test_unset_public_url_yields_non_empty_allowed_hosts` — fail-open guard: unset still yields a restrictive localhost-only allowlist.
- `test_portless_allowed_host_matches_host_with_port` — rmcp port-less-matches-any-port semantic.

## Validation
- Workspace: 6422 passed / 0 failed. Integration smoke: 24/24. Bug-relevant suites: 207 passed / 1 pre-existing xfail (GH#405).
- `unimatrix-server` clippy clean. (Pre-existing workspace clippy debt in `unimatrix-observe`/`unimatrix-engine` is tracked by GH#756/#688 and untouched here.)
- Bugfix gate: PASS (7/7).

### DIFF
--- a/crates/unimatrix-server/src/http/public_url.rs
+++ b/crates/unimatrix-server/src/http/public_url.rs
@@ -24,8 +24,9 @@ const PUBLIC_URL_VAR: &str = "UNIMATRIX_PUBLIC_URL";
 /// (FR-A5b pairing) before distributing the bundle.
 const PLACEHOLDER: &str = "https://<EDIT-ME>:8443";
 
-/// Sentinel host emitted when the knob is unset. Triggers permissive-with-warning
-/// `allowed_hosts` posture downstream; never appended to the SAN list.
+/// Sentinel host emitted when the knob is unset. Yields a localhost-restrictive
+/// `allowed_hosts` posture downstream (only the local SANs gate-pass; no public
+/// host is admitted) plus a startup WARN; never appended to the SAN list.
 const PLACEHOLDER_HOST: &str = "<EDIT-ME>";
 
 /// Local SANs always present regardless of the public URL (loopback + wildcard
@@ -120,7 +121,8 @@ pub fn derive_public_url(env: &Env) -> PublicUrl {
 fn placeholder() -> PublicUrl {
     tracing::warn!(
         "UNIMATRIX_PUBLIC_URL unset — emitting placeholder base-url ({PLACEHOLDER}) and \
-         permissive-with-warning allowed_hosts. Set it before distributing the bundle."
+         localhost-restrictive allowed_hosts (no public host admitted). Set it before \
+         distributing the bundle."
     );
     PublicUrl {
         base_url: PLACEHOLDER.to_string(),
@@ -310,9 +312,9 @@ mod tests {
     }
 
     #[test]
-    fn test_unset_public_url_allowed_hosts_permissive_with_warning() {
-        // Permissive-with-warning posture is signalled by the sentinel host;
-        // the real host is NOT appended to the SAN list.
+    fn test_unset_public_url_allowed_hosts_localhost_restrictive() {
+        // Localhost-restrictive posture: the sentinel host is NOT appended to the
+        // SAN list, so allowed_hosts carries only the local SANs (no public host).
         let pu = derive(None);
         assert_eq!(pu.host, "<EDIT-ME>");
         assert_eq!(
--- a/crates/unimatrix-server/src/http/router.rs
+++ b/crates/unimatrix-server/src/http/router.rs
@@ -320,13 +320,27 @@ impl McpAdapter {
     /// Create a new McpAdapter wrapping a `StreamableHttpService`.
     ///
     /// `allowed_origins` configures Origin header validation (ADR-002).
-    /// Empty vec = no origin restriction (backward-compatible default).
-    /// `allowed_hosts` is NOT modified — rmcp defaults it to localhost,
-    /// which is the CVE-2026-42559 fix.
-    fn new(server: UnimatrixServer, max_body_bytes: usize, allowed_origins: Vec<String>) -> Self {
+    /// `allowed_hosts` configures Host header validation (rmcp CVE-2026-42559),
+    /// wired from `PublicUrl.sans` (bug #774 — was previously left at rmcp's
+    /// localhost-only default, 403ing every non-localhost public host).
+    ///
+    /// OPPOSITE EMPTY-SEMANTICS — read before touching either field:
+    /// - empty `allowed_origins` = NO origin restriction (safe default, vnc-023).
+    /// - empty `allowed_hosts`   = ALLOW-ALL (FAIL-OPEN, defeats CVE-2026-42559).
+    ///
+    /// Callers MUST pass a NON-EMPTY `allowed_hosts` (`PublicUrl.sans` is
+    /// structurally non-empty — ≥3 local SANs even when the public URL is unset).
+    /// The wiring site guards this with a `debug_assert!`.
+    fn new(
+        server: UnimatrixServer,
+        max_body_bytes: usize,
+        allowed_origins: Vec<String>,
+        allowed_hosts: Vec<String>,
+    ) -> Self {
         let session_manager = Arc::new(LocalSessionManager::default());
         let mut config = StreamableHttpServerConfig::default();
         config.allowed_origins = allowed_origins;
+        config.allowed_hosts = allowed_hosts;
 
         // Capture the adapter's store BEFORE `server` is moved into the rmcp
         // closure (vnc-034 Wave 2 — resolve/dispatch agreement, OQ-PR-4).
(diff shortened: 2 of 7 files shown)


dug-21 (commented):
## Security Review — bug #774 (allowed_hosts wiring) — Fresh Context

**Risk level: LOW. No blocking findings.** (Posted as comment — cannot self-approve.)

Reviewed the full diff cold against rmcp 1.7.0 source (`=1.7.0` pin) and the `PublicUrl` construction paths. The fix correctly closes the DNS-rebinding gap (CVE-2026-42559) without widening the trust boundary.

### Verified correct

- **Empty=fail-open trap is guarded.** rmcp `host_is_allowed` (tower.rs:252) returns `true` (allow-all) on an empty `allowed_hosts`. The fix sources `allowed_hosts` from `PublicUrl.sans`, which is **structurally non-empty on every construction path**: `placeholder()` (unset / empty / un-parseable URL) always returns the 3 `LOCAL_SANS`; the parse-success path starts from `LOCAL_SANS` then appends. There is no path yielding an empty `.sans`. The `debug_assert!` is a correct defense-in-depth tripwire for future refactors — not the primary guarantee.
- **Release-build fail-open is NOT reachable.** `debug_assert!` compiles out in release, but `.sans` non-emptiness is a structural invariant of `PublicUrl`, independent of the assert. Acceptable.
- **Unset PublicUrl stays restrictive (CVE posture preserved).** When `UNIMATRIX_PUBLIC_URL` is unset, `allowed_hosts = [localhost, 127.0.0.1, 0.0.0.0]` — localhost-only, real host deliberately omitted. The gate does not widen on the unconfigured path. Covered by `test_unset_public_url_yields_non_empty_allowed_hosts`.
- **Port-less match semantic is real.** Confirmed against rmcp tower.rs:261-264 — a bare allowlist host (`port: None`) matches any incoming port, so `unimatrix` accepts `Host: unimatrix:8443`. Test T-774-3 pins this.
- **No over-permissive widening.** Only the operator-configured public host + loopback SANs clear the gate; an unrelated Host is 403'd (T-774-1 asserts both directions). Case-insensitivity holds (rmcp lowercases both sides).
- **All call sites wired.** Every `McpAdapter::new` / `from_server` / `from_servers` real call site threads `allowed_hosts`. The `from_servers` string in `local_binding_guard_tests.rs` is a meta-guard literal, not a call site.
- **Build + 28 targeted tests pass. No new dependencies. No secrets in diff.**

### Non-blocking nits

1. **Stale comment (LOW, cosmetic).** `public_url.rs` `placeholder()` says "permissive-with-warning allowed_hosts" but the wired behavior on unset is *localhost-restrictive*, not permissive. The code is the safe one; the wording predates the wiring and now misleads. Suggest aligning the comment to "localhost-only restrictive allowed_hosts" to avoid a future reader loosening it to match the stale doc.
2. **`0.0.0.0` in the allowlist (INFORMATIONAL).** A client sending `Host: 0.0.0.0` clears the gate. Not a meaningful rebinding vector (rebinding targets a name, not the literal wildcard-bind address) and it mirrors rmcp's own default + the cert SAN set. No action needed.

### Blast radius
Worst case of a subtle bug here is an over-restrictive 403 (legit host rejected → availability, the original #774 symptom) rather than an over-permissive gate (the fix sources only the configured host + loopback). The failure mode is fail-closed/safe, not silent exposure.

### Regression risk
Low. Localhost / HTTP-only / bundle-attach paths all retain `localhost`/`127.0.0.1` in `.sans`, so they continue to clear the gate. Test fixtures correctly pass non-empty `["localhost"]` rather than `vec![]`, avoiding baking the fail-open shape into the suite.

Issue #783, opened 2026-06-18
fix: provisioned slug store empty — knowledge routes to hash-dir store instead of per-slug store (vnc-034)


dug-21:
## Bug Report

**Environment:** Cloud / HTTP multi-project deployment (Docker, `serve --project-dir /data`).

**Symptom:** A project slug (`arch-research`) was provisioned via `unimatrix project register` — the `[[projects]]` stanza is present in `config.toml` and the slug data dir was created. However, knowledge stored 'into the project' did **not** land in the per-slug store at `/data/.unimatrix/{slug}/unimatrix.db`:

- The slug dir `/data/.unimatrix/arch-research/` has **no `unimatrix.db` and no vector data**.
- The path-hash data dir `/data/.unimatrix/{hash}/` has a **populated `unimatrix.db`**.

So writes appear to route to the hash-dir / default store rather than the registered per-slug store, despite multi-project routing (ADR-004: no default store).

**Secondary observation:** the production `config.toml` contains only the `[[projects]]` stanza + slug (no annotated default template). The full `DEFAULT_CONFIG_TOML` is only written by `handle_version` (`init`/`version`), never by the container `serve` path. Possibly in scope as a UX gap.

**Affected area (hint):** vnc-034 multi-project HTTP provisioning + store routing — `http_provision.rs` (build_project_server), `main.rs` listener wiring (MultiProjectRouter / StoreResolver), `projects.rs` (register), `projects/config_write.rs`.

**Expected:** knowledge stored against a registered slug persists in that slug's per-slug store.

**Open question for investigation:** how can a successfully-provisioned slug end up with an empty data dir while stored knowledge appears in the hash-dir store? Trace the write/store-resolution path end-to-end.

dug-21:
# Bug Investigation Report: 783-investigator

## Bug Summary
In a cloud/HTTP multi-project Docker deployment (`serve --project-dir /data`), a slug (`arch-research`) was registered (stanza present, slug dir created) but stored knowledge landed in the **path-hash store** `/data/.unimatrix/{hash}/unimatrix.db` instead of the per-slug store `/data/.unimatrix/arch-research/unimatrix.db`, which is empty.

## Root Cause Analysis

**The daemon was almost certainly NOT serving HTTP multi-project — it was serving the local UDS/STDIO transports bound directly to the path-hash store, and the client's writes went there.** The HTTP per-slug resolver was never reached.

[...]

PR #784, opened 2026-06-18
fix: downgrade benign pre-initialize disconnect to debug (#769)

### LINKED ISSUE #769: bug(personal-cloud): healthcheck logs "MCP session setup failed" at ERROR every 30s — false-alarm noise
## Summary
The container healthcheck makes operators think the server is broken (it did, during the dogfood).

## Evidence
```
ERROR unimatrix_server::uds::mcp_listener: MCP session setup failed error=connection closed: initialize request
```
…repeating exactly every 30s = `HEALTHCHECK --interval=30s`. The `health` probe connects to the MCP UDS socket and disconnects without an MCP `initialize`, which the listener logs at ERROR.

## Solution directions — APPROACH OPEN, choose by blast radius
- **(a) Log-level only:** downgrade this specific disconnect-before-initialize to `debug`. Tiny; leaves the probe behavior as-is.
- **(b) Clean probe handshake:** have `health` complete a minimal handshake or close in a way the listener treats as expected. Medium; touches the probe + listener contract.
- **(c) Dedicated liveness path** that doesn't traverse the MCP initialize path at all. Larger; cleanest separation of liveness vs MCP.

Server is healthy — this is pure log noise on the hot path. Pick per appetite.

## Vision alignment
`goal:personal-cloud` operator experience — ERROR-level false alarms erode trust in a self-run cloud.

dug-21:
## Summary
Fixes #769 — the container healthcheck logged `ERROR unimatrix_server::uds::mcp_listener: MCP session setup failed error=connection closed: initialize request` every 30s, making operators think a healthy server was broken (personal-cloud false alarm).

## Root cause
`run_session` (`crates/unimatrix-server/src/uds/mcp_listener.rs`) logged **every** `server.serve(...)` failure at `error!`. The ADR-003 health probe connects to the MCP UDS socket and disconnects before sending an MCP `initialize` (by design), which rmcp maps to the benign `ServerInitializeError::ConnectionClosed("initialize request")`. A liveness probe therefore emitted an ERROR line on every 30s `HEALTHCHECK` tick.

## Fix (Option a — variant-discriminated log level)
Match the `ServerInitializeError` variant in the `Err` arm:
- `ConnectionClosed(_)` → `debug!` (benign probe / client early disconnect)
- **all other variants + the `#[non_exhaustive]` catch-all `_`** → `error!` (default-to-loud)

Real handshake failures (`ExpectedInitializeRequest`, `InitializeFailed`, `UnsupportedProtocolVersion`, …) stay at ERROR. Pure log-level reclassification — no DB/locks/I/O, no behavior change to the probe or HTTP path. Diff confined to one file (+116/-1).

## Tests
- `test_run_session_client_closes_before_initialize_logs_debug_not_error` — connect+close with no bytes asserts **no** ERROR (DEBUG present).
- `test_run_session_malformed_message_still_logs_error` — non-initialize frame asserts ERROR **still** fires.

Locks in the benign-vs-real boundary. `uds::mcp_listener` 15/15 pass; workspace 6554/0; integration smoke 24/0; protocol suite 13/0. Changed file is clippy-clean (the workspace clippy red is pre-existing, tracked by #756).

## Gate
Bug Fix Validation — **PASS** (7/7).

### DIFF
--- a/crates/unimatrix-server/src/uds/mcp_listener.rs
+++ b/crates/unimatrix-server/src/uds/mcp_listener.rs
@@ -258,7 +258,20 @@ async fn run_session(
     let running = match server.serve((read_half, write_half)).await {
         Ok(r) => r,
         Err(e) => {
-            tracing::error!(error = %e, "MCP session setup failed");
+            // ConnectionClosed is benign: the health probe (ADR-003) and any client
+            // that disconnects before sending the MCP `initialize` request land here.
+            // Downgrade ONLY this variant to debug to stop ERROR-level log noise on the
+            // operator hot path. Every other variant — including the #[non_exhaustive]
+            // catch-all for future/unknown variants — stays at error! (default-to-loud)
+            // so real initialize failures remain diagnosable.
+            match e {
+                rmcp::service::ServerInitializeError::ConnectionClosed(_) => {
+                    tracing::debug!(error = %e, "MCP session setup closed before initialize");
+                }
+                _ => {
+                    tracing::error!(error = %e, "MCP session setup failed");
+                }
+            }
             return;
         }
     };
@@ -715,6 +728,108 @@ mod tests {
                 panic!("server.serve() hung for more than 5 seconds with closed client");
             }
         }
+    }
+
+    // -----------------------------------------------------------------------
+    // GH#769: health-probe / early-disconnect log level — benign vs. real failure
+    // -----------------------------------------------------------------------
+    //
+    // The container healthcheck (ADR-003) connects to the MCP UDS socket and
+    // disconnects before sending an MCP `initialize` request. rmcp maps that EOF
+    // to ServerInitializeError::ConnectionClosed. The fix routes ConnectionClosed
+    // to debug! while keeping every other variant (and the #[non_exhaustive]
+    // catch-all) at error!. These two cases lock in that benign-vs-real boundary.
+
+    /// Drive a single `run_session` against a server-side `UnixStream` whose peer
+    /// performs `client_action` (e.g., close immediately, or send a non-initialize
+    /// message). Returns once `run_session` has handled the failed `serve(...)` arm.
+    async fn drive_run_session_with_client<F, Fut>(dir: &TempDir, client_action: F)
+    where
+        F: FnOnce(tokio::net::UnixStream) -> Fut + Send + 'static,
+        Fut: std::future::Future<Output = ()> + Send,
+    {
+        use crate::infra::shutdown::new_daemon_token;
+        use tokio::net::{UnixListener, UnixStream};
+
+        let sock = dir.path().join("gh769-test.sock");
+        let listener = UnixListener::bind(&sock).unwrap();
+
+        let sock_path = sock.clone();
+        let client_task = tokio::spawn(async move {
+            let stream = UnixStream::connect(&sock_path).await.unwrap();
+            client_action(stream).await;
+        });
+
+        let (server_stream, _) = listener.accept().await.unwrap();
+
+        let server = build_test_server(dir);
+        let token = new_daemon_token();
+
+        // run_session must return promptly once the client's behaviour yields a
+        // failed serve(...) — bound it so a regression that hangs is caught.
+        let _ = tokio::time::timeout(
+            Duration::from_secs(5),
+            run_session(server_stream, server, token.child_token()),
+        )
+        .await
+        .expect("run_session must return after serve(...) fails");
+
+        let _ = client_task.await;
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[tracing_test::traced_test]
+    async fn test_run_session_client_closes_before_initialize_logs_debug_not_error() {
+        let dir = TempDir::new().unwrap();
+
+        // Client connects and immediately closes without writing any bytes —
+        // exactly what the container healthcheck does (ADR-003).
+        drive_run_session_with_client(&dir, |stream| async move {
+            // Drop the stream — sends EOF to the server before any initialize.
+            drop(stream);
+        })
+        .await;
+
+        // Benign path: a DEBUG event is present and NO ERROR event was emitted.
+        assert!(
+            logs_contain("MCP session setup closed before initialize"),
+            "ConnectionClosed must produce the debug-level setup-closed event"
+        );
+        assert!(
+            !logs_contain("MCP session setup failed"),
+            "ConnectionClosed must NOT emit the error-level setup-failed event (GH#769 noise)"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[tracing_test::traced_test]
+    async fn test_run_session_malformed_message_still_logs_error() {
+        use tokio::io::AsyncWriteExt;
+
+        let dir = TempDir::new().unwrap();
+
+        // Client sends a well-formed JSON-RPC message that is NOT an `initialize`
+        // request, then closes. rmcp yields a non-ConnectionClosed variant
+        // (ExpectedInitializeRequest), which must remain at ERROR level.
+        drive_run_session_with_client(&dir, |mut stream| async move {
+            let not_initialize =
+                b"{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"ping\",\"params\":{}}\n";
+            let _ = stream.write_all(not_initialize).await;
+            let _ = stream.flush().await;
+            // Close so the server stops waiting for more frames.
+            drop(stream);
+        })
+        .await;
+
+        // Real-failure path: the error-level setup-failed event MUST still fire.
+        assert!(
+            logs_contain("MCP session setup failed"),
+            "a non-initialize message must keep the error-level setup-failed event"
+        );
+        assert!(
+            !logs_contain("MCP session setup closed before initialize"),
+            "a real failure must NOT be downgraded to the benign debug event"
+        );
     }
 
     // -----------------------------------------------------------------------


dug-21 (commented):
## Security Review — PR #784 (GH#769 healthcheck log noise)

**Risk level: LOW. No blocking findings.** Fresh-context review of `git diff main...HEAD` (confined to `crates/unimatrix-server/src/uds/mcp_listener.rs`, +116/-1).

### What I verified
- **Variant match is correct against the pinned rmcp.** Lockfile resolves `rmcp 1.7.0`. In `rmcp-1.7.0/src/service/server.rs`, `ServerInitializeError::ConnectionClosed(String)` is constructed **only** by `expect_next_message` when `transport.receive()` returns `None` (EOF before a complete frame). A complete-but-non-initialize frame yields `ExpectedInitializeRequest`; a pre-init `ping` is answered and the loop continues. So `ConnectionClosed → debug!` matches exactly the benign probe / early-disconnect case, and every real handshake failure (`ExpectedInitializeRequest`, `InitializeFailed`, `UnsupportedProtocolVersion`, `TransportError`, `UnexpectedInitializeResponse`, `Cancelled`) falls to `_ → error!`.
- **`#[non_exhaustive]` is honored.** The enum is `#[non_exhaustive]`; the `_` arm stays loud (error!), so future/unknown variants default to ERROR. Correct default-to-loud posture.
- **No new dependencies.** `tracing-test` is a pre-existing dev-dependency; `build_test_server` / `new_daemon_token` / `TempDir` / `traced_test` are all already used elsewhere in this test module. Tests extend the existing module rather than scaffolding new infra.
- **Boundary tests pass locally:** `test_run_session_client_closes_before_initialize_logs_debug_not_error` and `test_run_session_malformed_message_still_logs_error` — 2/2 green.

### OWASP / boundary checks
- **Injection / log-injection:** None. The change adds no parsing and no new untrusted input. The `ConnectionClosed` payload is rmcp's hardcoded context literal `"initialize request"`, not attacker-controlled. `error = %e` Display formatting is unchanged from prior behavior.
- **Access control / trust boundary:** Unchanged. No auth/privilege logic touched. The UDS handshake contract is untouched — only the log macro selected on the failure path changes.
- **Deserialization / input validation:** No new deserialization; rmcp still owns frame parsing. No validation was removed.
- **Secrets:** None introduced. No credentials, tokens, or keys in the diff; no new sensitive data is logged (the downgraded line carries strictly less to the operator log, not more).

### Log-suppression / flood angle (explicitly assessed)
- **Suppression:** An attacker who can reach the UDS socket could connect-then-close to keep landing in the now-`debug!` arm. This is **not materially worse than today** — pre-fix, that same action emitted ERROR noise (the very bug). It does not suppress *other* sessions' logs; each session logs independently. Real handshake failures from that same attacker still log at ERROR.
- **Flood:** Pre-init disconnects no longer flood ERROR; they flood DEBUG only when debug logging is enabled. Net reduction in default-level log pressure. UDS reachability is already a local trust boundary (filesystem-permissioned socket), so this is not a new exposure.

### Blast radius (worst case if the match were subtly wrong)
Bounded to log level only — no DB/lock/I/O/control-flow change; `return` behavior is identical in both arms. Worst case of a *misclassification* would be a real initialize failure logged at `debug!` instead of `error!` (reduced observability, not a security or correctness defect). I confirmed against rmcp 1.7.0 source that this cannot happen for the enumerated real-failure variants.

### Regression risk (informational, non-blocking)
A client whose connection drops **mid-handshake** due to a genuine fault (crash/network) also maps to `ConnectionClosed` and will now log at `debug!`. At this layer it is indistinguishable from a benign probe disconnect, and the session simply ends with no side effects — an acceptable, intentional trade-off consistent with the issue's chosen Option (a). No action required.

**Findings: 0 blocking, 1 informational. Approving from a security standpoint.**

dug-21 (commented):

**Gate**: pr-review | **Stance**: Merge as-is. Clean, minimal hygiene fix directly serving `goal:personal-cloud` operator trust; security review is LOW with zero blocking findings; all gates and CI green. No follow-up blocking.

### Vision / roadmap fit
This advances `goal:personal-cloud` (#4946). A single-tenant, self-operated deployment has exactly one operator-trust signal — its logs — and no SRE team to triage false alarms. An ERROR firing every 30s from the server's own ADR-003 healthcheck is the precise failure mode that makes a healthy server read as a broken one. Removing it keeps the nan-016 (#681) UDS dogfood proving ground credible on its way to F6 Rust-hook retirement, which is the named delivery track for this goal.

It also reinforces the "quiet, cheap hot path" architectural posture: the 30s probe is on the accept path, and the fix is strictly cheaper than today (DEBUG filtered at default vs. ERROR formatted+emitted). Nothing in-flight is affected in either direction — the vnc-034 Wave 1 HTTPS serving arc (#726/#725) uses an independent streamable-http serve flow and is untouched. Pure hygiene that removes a trust-eroding signal; no roadmap item is constrained or enabled.

### Approach commentary
The delivery as a whole is coherent and the convergence across investigation, design review, security review, and gate is unusually tight — and grounded, not improvised. The fix applies a *codified subsystem convention* (#3448 col-006 "classify errors at the point of origin, never a single loud catch-all"; #3455 bugfix-66 "intercept expected fire-and-forget EOF, downgrade to DEBUG, keep genuine violations loud") to rmcp's typed `#[non_exhaustive]` enum instead of an `io::ErrorKind`. The health probe's connect-then-close-without-writing is the MCP-transport analogue of that exact UDS case, so this makes `run_session` consistent with `handle_connection` rather than inventing a new rule.

Strengths the human can rely on:
- **Boundary is structurally enforced, not just asserted.** Security review independently re-verified against pinned rmcp 1.7.0 source that `ConnectionClosed(String)` is constructed *only* on EOF-before-first-frame. A real client sending malformed bytes yields `ExpectedInitializeRequest`/`InitializeFailed`/`UnsupportedProtocolVersion` — all of which fall to `_ => error!`. Misclassification of a real failure as benign is not reachable for the enumerated variants.
- **Default-to-loud catch-all is honored.** The `#[non_exhaustive]` catch-all stays at `error!`, so a future rmcp variant surfaces loudly rather than being silently swallowed. This was the single blocking design constraint and it is satisfied (gate spot-checked source lines 267-274).
- **The durable value is the companion test.** `test_run_session_malformed_message_still_logs_error` is what prevents a future change from silently re-muting real handshake failures — that regression guard outlasts the bug itself.

Security review: LOW, 0 blocking / 1 informational. I concur with assessing its findings as non-blocking. The log-suppression/flood angle is correctly judged "not materially worse than today" (the pre-fix behavior emitted *louder* ERROR noise for the same connect-close, and UDS reachability is already a filesystem-permissioned local trust boundary). The one informational item — a genuine mid-handshake fault also maps to `ConnectionClosed` and now logs at DEBUG — is an accepted, intentional trade-off of Option (a) and indistinguishable at this layer; no action needed.

No open questions remain at this gate — all four fix-approach questions (solution direction, catch-all behavior, `Cancelled` scope, match width) were resolved at the earlier human gate and the implementation matches those resolutions.

One closed decision worth surfacing for the human's confirmation, not reopening:
- **`Cancelled` was intentionally left at ERROR (not downgraded).** The fix-approach review (and my own recommendation there) suggested optionally downgrading `Cancelled` with a distinct message. The implementer scoped it out and matched `ConnectionClosed(_)` only. **Recommended answer: accept as-is.** `Cancelled` is rare, orthogonal to this bug, and keeping the diff to the one proven-noisy variant is the right minimality call. If daemon-shutdown-during-setup ERROR noise is observed in the dogfood logs later, file it as its own one-line follow-up — do not hold this PR for it.

### Recommended actions
1. **Merge PR #784 as-is.** Root-cause coverage, minimality (+116/-1, one file), test completeness (silence guard + real-failure guard), smoke (24/0) and protocol suite (13/0) all PASS; security LOW with no blocking findings; all CI checks green.
2. **No blocking follow-up.** Optional, non-blocking: if `Cancelled`-during-shutdown ERROR noise appears in dogfood logs, open a trivial follow-up to give it the same DEBUG treatment with a distinct message — not required now.
3. **Knowledge thread already closed well.** The implementer stored #5131 ("rmcp serve() EOF before initialize is benign — match the variant, default-to-loud"). Recommend (non-blocking) that it be linked as an edge/extension to #3448 so the "classify-at-origin / downgrade-benign / default-to-loud" convention stays one traceable thread rather than a disconnected entry — both prior reviewers flagged this and it preserves the provenance that makes the convention reusable.

dug-21:
## uni-zero product review (advisory — human makes the merge call)

**Stance: approve.** Correct fix at the right altitude. One known-benign variant to decide consciously before merge.

### Endorse
Exactly the recommended Option (a): `ServerInitializeError::ConnectionClosed(_)` → `debug!`, catch-all `_` → `error!` (default-to-loud on the `#[non_exhaustive]` enum), pure log reclassification, one file, no probe/contract change. The tests are the right shape — `traced_test` + `logs_contain` assert **both** directions (benign disconnect → debug present + no error; non-initialize frame → error present + no debug), locking the benign-vs-real boundary behaviorally, not just at compile time. Low risk, high operator-trust payoff (this noise was hit live during the arch-research dogfood).

### One thing to decide consciously: `Cancelled` still logs at ERROR
The investigator flagged `Cancelled` as "arguably also downgrade-worthy"; the implementation folds it into the `_ => error!` catch-all. The default-to-loud bias is correct for *unknown/future* variants — but `Cancelled` is a **known-benign** one (graceful daemon shutdown cancelling in-flight sessions, not a fault), so it's a different case from the catch-all.

Why it matters here specifically: the **#783 happy path is `register` → restart**. Any in-flight MCP session cancelled on that restart emits the exact class of ERROR false-alarm this PR exists to kill — lower frequency than the 30s healthcheck, but right inside a common personal-cloud operator flow. Recommend either including `Cancelled` in the `debug!` arm, or documenting why it stays loud. Treating a *known* benign variant identically to *unknown* ones conflates two cases.

### Optional, not for this PR
Longer term, a dedicated liveness path (issue option c) that never traverses the MCP `initialize` path removes the "probe deliberately trips a benign error, discriminate by log level" fragility (a future rmcp reclassification could re-surface noise or downgrade a real error). Over-build for a log-level issue today; (a) is proportionate. Note only if the pattern recurs.

**Net: approve; the single author decision is the `Cancelled`-on-restart case — pick it deliberately rather than by catch-all.**

Issue #785, opened 2026-06-18
feat(personal-cloud): per-slug config overlay resolution — C6 / Feature A (vnc-034 deferred)


dug-21:
## vnc-040 — Per-Slug Configuration Overlay Resolution (C6 / Feature A)

> Scope approved 2026-06-19. Source of truth: `product/features/vnc-040/SCOPE.md`.

Delivers **C6 — Per-slug configuration** (Unimatrix #5148). Prerequisite C5 (#5190) proved on the crt-056 / #789 merge (2026-06-19) → unblocked. **Feature A** of the two-part split; **B (seeding) is OUT of scope** here (tracked follow-up).

### Problem
Configuration is daemon-global: `load_config` resolves one `UnimatrixConfig` (defaults → global → per-project → env) that governs **every** routed slug. Each slug owns an isolated store (vnc-034) but **no config file**. `ProjectConfigEntry` is slug-only by design; crt-056 ADR-002 reserved this exact seam (*"#785 later adds overrides; this ADR must not introduce that seam"*).

### Approach (locked)
At the per-slug loop (`main.rs:1089-1110`): probe `{base_dir}/{slug}/config.toml`; if present, `load_single_config` + `validate_config` then `merge_configs(global_resolved, per_slug_file)` (per-key merge, **reuse** existing infra — no new merge). Derive per-slug `Arc`s for overlayable fields; always `Arc::clone` global model handles + pool. No file → byte-for-byte global fallthrough. Resolution is the third precedence layer atop dsn-001 #2286's field-level replace (the issue's "ADR-003" → resolved as #2286, no conflict).

### Overlay-vs-global verdict — centerpiece deliverable (10 call-site inputs)
| Input | Treatment |
|---|---|
| `embed_handle` + entire `[embedding]` section (model/dims/sha256) | **GLOBAL — locked, hard invariant** |
| `nli_handle`, `rayon_pool` | **GLOBAL — locked** |
| `nli_top_k`, `nli_enabled` | overlayable (not model-coupled) |
| `inference_config` | overlayable EXCEPT `*_sha256` pins global-wins (#4655/#4649) |
| `confidence_params`, `categories`, `observation_registry`, `boosted_categories` | **per-slug** |

[...]

PR #786, opened 2026-06-18
fix: enable HTTP multi-project posture in cloud Docker image (#783)

### LINKED ISSUE #783: fix: provisioned slug store empty — knowledge routes to hash-dir store instead of per-slug store (vnc-034)

dug-21:
## Summary
Fixes #783. The cloud serving env `UNIMATRIX_HTTP_ENABLED=true` (vnc-034 ADR-007) lived only in `docker-compose.yml`, which the release pipeline does not ship. A clean `docker run` of the GHCR image booted with `http.enabled=false` → `[[projects]]` routing inert → only the local UDS transport live → writes against a registered slug silently misrouted to the path-hash store (`/data/.unimatrix/{hash}/`) instead of the per-slug store (`/data/.unimatrix/{slug}/`).

## Changes
- **Dockerfile**: bake `ENV UNIMATRIX_HTTP_ENABLED=true` into the **runtime** stage (not builder); operator escape hatch `-e UNIMATRIX_HTTP_ENABLED=false` preserved (env override precedence unchanged). Stale 'set in compose.yaml' comment corrected.
- **main.rs**: pure fail-fast guard `require_http_for_projects(http_enabled, slug_count)` called after config load — aborts boot with an actionable error on HTTP-off + `[[projects]]` present. Defense-in-depth; provably does NOT trip the legitimate local single-project install (empty `[[projects]]` + HTTP off).

## Tests
- 6 guard unit tests (all HTTP×slugs quadrants + 2 explicit-override-regression cases)
- 2 Dockerfile-content assertions (runtime enables, builder does not)
- Real `docker build`+boot smoke (`product/test/infra-001/scripts/docker-http-posture-smoke.sh`): clean `docker run` → register slug → restart → HTTPS write → asserts entry lands in the per-slug DB, not the hash store. **Ran here (Docker available): 3/3 PASS.**
- Full workspace: **6560 passed / 0 failed**; integration smoke 24/24.

## Notes
- Boot-safety verified: HTTP-on + zero projects boots cleanly; `UNIMATRIX_PUBLIC_URL` unset is non-fatal; TLS self-provisions first boot.
- Pre-existing workspace clippy drift (clippy 0.1.95 in `unimatrix-engine`/`unimatrix-observe`, untouched crates) is out of scope — none in the changed files.
- Follow-ups: #785 (per-slug config overlay — design track; also notes the `serve` config-seeding devex gap). Wire the docker smoke into a Docker-capable CI lane (self-skips when Docker absent).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,10 +119,17 @@ COPY --from=builder --chown=65532:65532 /data /data
 COPY --from=builder --chown=65532:65532 /shared /shared
 
 # Environment (ADR-005).
+# UNIMATRIX_HTTP_ENABLED=true bakes the cloud serving posture into the shipped
+# image (vnc-034 ADR-007 #4953). A clean `docker run` of this image must boot
+# HTTP-serving so registered `[[projects]]` slugs are routable; without it the
+# binary default `http.enabled=false` leaves slugs inert and misroutes writes to
+# the path-hash store (#783). Override for a UDS-only container with
+# `-e UNIMATRIX_HTTP_ENABLED=false`. RUNTIME stage only — never the builder.
 ENV HOME=/data \
     LD_LIBRARY_PATH=/usr/local/lib \
     UNIMATRIX_LOG=info \
-    UNIMATRIX_MODEL_CACHE=/shared/models
+    UNIMATRIX_MODEL_CACHE=/shared/models \
+    UNIMATRIX_HTTP_ENABLED=true
 
 # Volume mount point for persistent data.
 #
@@ -141,9 +148,11 @@ HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
 # TLS serving port (vnc-034 / ADR-007, FR-A8, AC-W1-S2, R-12).
 # ONLY the TLS port is exposed — there is NO plaintext port. The container serves
 # pinned-TLS HTTPS only; OSS posture has no plaintext-to-client mode.
-# The HTTP listener is gated on UNIMATRIX_HTTP_ENABLED=true (set in compose.yaml,
-# ADR-007). The global binary default http.enabled=false is unchanged — flipping
-# serving on is a container-scoped env concern, not a code-default change.
+# The HTTP listener is enabled by `ENV UNIMATRIX_HTTP_ENABLED=true` baked into the
+# runtime stage above (cloud posture, vnc-034 ADR-007). Override with
+# `-e UNIMATRIX_HTTP_ENABLED=false` for a UDS-only container. The global binary
+# default http.enabled=false is unchanged — flipping serving on is a
+# container-scoped env concern, not a code-default change.
 # TLS auto-enables from the first-boot-provisioned cert (provisioned in the Rust
 # binary, since distroless has no shell). The token/cert stay 0600 files on the
 # /data volume and are NEVER baked into an image layer (NFR-06).
--- a/crates/unimatrix-server/src/main.rs
+++ b/crates/unimatrix-server/src/main.rs
@@ -103,6 +103,36 @@ fn assert_wave_b_precondition(
     Ok(())
 }
 
+/// #783 defense-in-depth: refuse to boot when `[[projects]]` declares routable
+/// slugs but the HTTP transport is disabled.
+///
+/// Registered projects are ONLY reachable over the HTTP per-slug resolver
+/// (vnc-038 ADR-006: local STDIO/UDS keep a direct path-hash store binding and
+/// never enter the resolver). With HTTP off, declared slugs are inert and client
+/// writes silently land in the path-hash store instead of the per-slug store — a
+/// catastrophic, unrollbackable misroute (vnc-034 ADR-004 SR-06). Failing loud at
+/// boot makes that misconfiguration impossible to hit silently.
+///
+/// Pure predicate (takes the resolved `http_enabled` flag and slug count, not the
+/// env var) so it is unit-testable without booting — mirrors the config.rs
+/// pure-function idiom. The env override precedence is ALREADY applied inside
+/// `load_config` (config.rs step 3c); callers pass the resolved `config.http.enabled`.
+///
+/// Returns `Err` ONLY for the (HTTP-off + slugs-present) quadrant. The legitimate
+/// local single-project install (empty `[[projects]]` + HTTP off) returns `Ok`.
+fn require_http_for_projects(http_enabled: bool, slug_count: usize) -> Result<(), String> {
+    if !http_enabled && slug_count > 0 {
+        return Err(format!(
+            "{slug_count} project slug(s) are declared in `[[projects]]` but the HTTP \
+             transport is disabled. Registered projects are reachable ONLY over HTTP; with \
+             HTTP off they are inert and writes misroute to the path-hash store. Remedy: \
+             enable HTTP (set `[http] enabled = true` in config.toml, or \
+             `UNIMATRIX_HTTP_ENABLED=true`), or de-register the projects."
+        ));
+    }
+    Ok(())
+}
+
 /// Unimatrix knowledge engine.
 #[derive(Parser)]
 #[command(name = "unimatrix", about = "Unimatrix knowledge engine")]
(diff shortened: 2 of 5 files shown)


dug-21 (commented):

**Risk level: LOW. No blocking findings. Approve from a security standpoint.**

Reviewed the full diff cold (Dockerfile, main.rs guard, main_tests.rs, dockerfile_http_posture.rs, docker-http-posture-smoke.sh) and verified the load-bearing security claims against source rather than the narrative.

### The central question: does HTTP-on-by-default expose an unauthenticated surface?
**No.** Verified in `main.rs` (HTTP startup block, ~1006-1150):
- The listener mandatorily calls `load_or_generate_token` and wraps **every** served route in `StaticTokenAuthLayer` (main.rs:1137-1138). There is no unauthenticated arm — a clean image is NOT exposed without a token. Token is auto-generated `O_CREAT|O_EXCL` mode 0600 (`http/token.rs`).
- TLS is self-provisioned first-boot (`provision_tls`), key written 0600 (`cert_provisioner.rs`); no plaintext port — only 8443/TLS is `EXPOSE`d.
- rmcp Host-header allowlist is fed from `public_url.sans` (non-empty by construction) with a `debug_assert!` fail-open guard against CVE-2026-42559. Origin allowlist wired from config.
- Zero-projects clean boot serves nothing routable — every `/v1/...` is a uniform 404 (main.rs:1061-1076). So enabling HTTP by default adds no servable surface until an operator explicitly registers a slug.

This is the intended cloud posture (vnc-034 ADR-007) and it is safe out-of-the-box: auth + TLS are non-optional on the served port.

### Escape hatch / env precedence
`-e UNIMATRIX_HTTP_ENABLED=false` is honored: `resolve_http_enabled_override` (config.rs) maps `false`/`0` → `Some(false)`, applied at step 3c AFTER file/global merge so env wins (config.rs:3300-3305). The boot guard reads the **resolved** `config.http.enabled` field — it does not re-read the env var, so there is no precedence drift or bypass. No injection in the override parse (total function, case-insensitive, trimmed, never panics).

### Boot guard (require_http_for_projects) — OWASP / access control
Pure predicate; fires only in the (HTTP-off + slugs-present) quadrant. Error string is operator-facing (daemon stderr/log) only — no client channel, no remote disclosure, leaks only the operator's own slug count back to the operator. No injection, no input from untrusted source. Correct.

### Blast radius / regression
- Dockerfile ENV affects the **image only**. The local UDS single-project install runs the binary directly and never inherits image ENV — compiled default `http.enabled=false` provably intact (config.rs default + ADR-006 direct path-hash binding). No routing/store-binding code changed.
- Worst-case if the guard were subtly wrong: it could refuse a legitimate boot. Verified it CANNOT fire for the local case (empty `[[projects]]` → slug_count 0 → Ok), and the clean image ships HTTP-on so it never trips by default. Failure mode is safe (loud refusal), never silent corruption — which is the exact integrity property this fix restores.

### Smoke shell script
`set -euo pipefail`; `mktemp -d` (no predictable temp path); busybox sidecar mounts the volume **read-only** for inspection and never `docker exec`s into distroless; `$$`-suffixed unique container/volume names; cleanup trap on EXIT; `curl --cacert` (cert-pinned, NOT `-k`/insecure); bearer token read from the volume at runtime (never baked). No `eval`, no unquoted command injection, no privileged/`--cap-add` flags. Clean.

### Secrets
No hardcoded credentials/tokens/keys in the diff. The baked ENV carries only the non-sensitive `UNIMATRIX_HTTP_ENABLED=true` boolean; token/cert remain 0600 files on the /data volume, never an image layer (NFR-06). The smoke's token comes from the running volume.

### Minor observations (non-blocking, no action required)
- `docker-compose.yml` still sets `UNIMATRIX_HTTP_ENABLED: "true"` — now redundant with the baked ENV but harmless and documents intent.
- dockerfile_http_posture.rs correctly scopes the assertion to the runtime stage and asserts the builder stage does NOT carry the env — good defense against accidental relocation.

**Verdict: LOW risk, no blocking findings.**

dug-21:
## uni-zero product review (advisory — human makes the merge call)

**Stance: endorse.** The release-layer fix is the correct one, the guard is clean, and the regression test is the right shape. One follow-up below must not stay a PR note.

### Fix is correct
- **Dockerfile bake** is exactly right: `UNIMATRIX_HTTP_ENABLED=true` in the **runtime** stage only, escape hatch `-e UNIMATRIX_HTTP_ENABLED=false` preserved, and the stale "set in compose.yaml" comment corrected. This makes the *shipped artifact* self-correct rather than depending on a compose file the release never ships — the actual root cause.
- **Guard** (`require_http_for_projects`) is a pure predicate reading the **resolved** `config.http.enabled` (not re-reading the env), fires only in the (HTTP-off + slugs-present) quadrant, and provably returns `Ok` for the local single-project install. Correct defense-in-depth that does not violate ADR-006.
- **No new unauthenticated surface** (the one real risk of HTTP-on-by-default): verified — auth (`StaticTokenAuthLayer`) + self-provisioned TLS are mandatory on the served port, and a zero-project clean boot serves a uniform 404. Enabling HTTP adds no servable surface until an operator registers a slug.

### The test is the highest-value part — and the reason this whole class existed
The real `docker build` → boot → register → restart → per-slug HTTPS route smoke (`docker-http-posture-smoke.sh`, 3/3 PASS) is precisely the test that would have caught the shipped bug. #783, #774, vnc-039's init wiring, and #769 were all the **same blind spot: the shipped image's first-run path was never validated against the actual artifact.** This smoke is the fix for the *class*, not just this bug.

**Which is why the deferred step is load-bearing, not a footnote:** the smoke currently self-skips when Docker is absent and ran only locally this time. Until it is wired as a **standing gate on the release/image-build pipeline** (where the image is built anyway), the protection is "someone remembered to run it." I'd track that wiring as its own issue against the release workflow and treat it as the real close-out of this class — not a "wire it later" note. Rust/container validation has no home in JS CI by design; the release workflow is the correct home.

### Optional hardening (non-blocking)
Gate 3 asserts the per-slug DB *exists*; gate 2's 204 proves the slug route is *live* (the structural fix, since ADR-004 has no fallback arm). To guard the **exact** original symptom ("slug dir empty, hash dir populated") against a *different-mechanism* future regression, consider also asserting the per-slug store actually **grew** from the write and the hash store did **not**. The 204 already covers the routing; this just pins the data-landing assertion to the literal bug. Incremental, not a blocker.

### Scope
`#785` (per-slug config overlay + the `serve` config-seeding devex gap) correctly split to a separate track — agreed; that's feature work, not this bugfix.

**Net: approve. Merge gate from my lens = wire the docker smoke into the release pipeline as a standing gate (tracked), so this class can't reship silently.**

PR #796, opened 2026-06-19
fix(personal-cloud): release gate pulls prebuilt IMAGE in smoke; gate npm/binary builds off dispatch (#795)

### LINKED ISSUE #795: bug(personal-cloud): nan-019 release gate blocks every release — smoke jobs never `docker pull` the image (caught by dispatch dry-run)
**Found by the nan-019 `workflow_dispatch` dry-run** ([run 27834008646](https://github.com/dug-21/unimatrix/actions/runs/27834008646)) — the first real cross-platform exercise of the new release gate, exactly the post-merge round-trip the design budgeted for (#4796 / AC-07). The gate caught a defect in its **own wiring** before any `v*` tag was cut.

Follow-up to nan-019 (#788, PR #792). Guards **N5** (the gate must itself be deployable/working) — without this fix the gate blocks every release.

## 🔴 Finding 1 (release-blocking) — smoke jobs never `docker pull` the resolved image

Both `smoke-amd64` and `smoke-arm64` failed **identically**:

```
smoke target IMAGE=ghcr.io/dug-21/unimatrix:latest-amd64
[783-smoke] using prebuilt image: ghcr.io/dug-21/unimatrix:latest-amd64
Error response from daemon: No such image: ghcr.io/dug-21/unimatrix:latest-amd64
[783-smoke] FAIL: image ENV is missing UNIMATRIX_HTTP_ENABLED=true
##[error]smoke FAILED (exit 1): shipped image first-run path is broken.
```

**Root cause:** the tag resolution is correct (`:latest-<arch>`), but the smoke job does `docker login` and then runs the smoke with `IMAGE=` set. With `IMAGE=` set, `docker-http-posture-smoke.sh` **skips its build branch and assumes the image is present in the local daemon** — but `smoke-amd64`/`smoke-arm64` run on a *different* runner than `build-container-x64`/`build-container-arm64`, so the GHCR image is not in the smoke runner's local cache. The job never `docker pull`s it → `docker image inspect` → `No such image` → the ENV check false-fails.

**Not dispatch-specific or arch-specific** — this is the same fresh-runner, no-local-image path a `v*` push hits. On a real release: `build-container-*` pushes `:v<version>-<arch>`, then `smoke-*` (separate runner) resolves the right tag but can't inspect/run it → smoke fails → manifest (which `needs:` the smokes on push) is blocked → **every `v*` release blocked.**

**Why it escaped the gates:**
- Local validation (impl + Stage 3c) ran the smoke with `IMAGE=` where the image had just been **built locally** (present in cache) — the pull was never needed.
- The pre-merge gate-logic test **stubs** the smoke; the tag-parity test checks only the tag *string*. Nothing exercised "pull a remote image that is not in the local cache, then smoke it."

**Fix:** add an explicit `docker pull "$IMAGE"` in each smoke job after `docker login` and before `run_smoke_gate` (or pull inside the gate lib / the smoke's `IMAGE=` branch — design choice for the bugfix). The pull must hard-fail the gate if it 404s (a genuinely missing pushed tag IS a real release defect — keep that signal; just distinguish "couldn't pull" from "image first-run broken" in the diagnostic).

**Regression coverage (the actual gap):** add a test that exercises the image-not-present-locally path — e.g. resolve + pull a tag, or assert the smoke job step issues a `docker pull` before invoking the smoke. The current suite cannot catch a re-introduction of this bug.

## 🟠 Finding 2 (minor) — `package-npm` / `create-release` not gated off on `workflow_dispatch`

nan-019 added `workflow_dispatch` at the workflow level and gated only `create-container-manifest` off for dispatch. `package-npm` and `create-release` were not gated, so they also run on a dispatch.

**No harm occurred (verified):** `package-npm` failed at its *Validate versions* step (`TAG_VERSION=${GITHUB_REF_NAME#v}` = `main` ≠ Cargo version) **before** reaching any `npm publish`; `create-release` (`needs: package-npm`) was skipped. No package published, no release cut.

**But** it makes the sanctioned dry-run show an overall **FAILURE** even when only the smoke signal matters — muddying exactly the signal the manifest gating was added to keep clean.

**Fix:** add `if: github.event_name != 'workflow_dispatch'` to `package-npm` (and `create-release` then skips via `needs:`), so a dispatch dry-run reduces cleanly to the two `smoke-*` statuses.

## Validation for the bugfix
Re-run the `workflow_dispatch` dry-run after the fix: expect `smoke-amd64` + `smoke-arm64` **green** (the first true both-arch hosted proof, AC-07 dispatch leg — watch arm64 cold-boot vs the 90s deadline, #767 / R-05), `create-container-manifest` **skipped**, and `package-npm`/`create-release` **skipped** → overall green dry-run.

## Refs
- Feature: nan-019 (#788), PR #792 (merged `28de4553`)
- Failing run: https://github.com/dug-21/unimatrix/actions/runs/27834008646
- Files: `.github/workflows/release.yml` (smoke jobs + `package-npm`/`create-release` gating), `product/test/infra-001/scripts/release-gate-lib.sh`, `product/test/infra-001/scripts/docker-http-posture-smoke.sh` (`IMAGE=` branch)

dug-21:
## Summary

Fixes #795 — the nan-019 release gate blocked **every** `v*` release because the `smoke-amd64`/`smoke-arm64` jobs never `docker pull`ed the resolved image on their fresh runner.

**Root cause:** smoke jobs run on a runner separate from `build-container-*`. They set `IMAGE=ghcr.io/.../:latest-<arch>`, so `docker-http-posture-smoke.sh` takes its prebuilt branch and skips the build — but the first docker op was `docker image inspect` (no implicit pull) → `No such image` → false "image first-run path is broken." Caught by the nan-019 `workflow_dispatch` dry-run ([run 27834008646](https://github.com/dug-21/unimatrix/actions/runs/27834008646)) — the post-merge round-trip the design budgeted (#4796 / AC-07).

## Changes

- **F1 — conditional pull (not unconditional).** `docker-http-posture-smoke.sh` IMAGE= branch now does `docker pull "$IMAGE" || docker image inspect "$IMAGE" || exit 4` **before** inspect. On an ephemeral hosted runner this always really pulls (preserves ADR-002 "test the pushed bytes"); on a dev box with a locally-built tag it falls back to the present local image; only a genuinely unavailable image hits the dedicated **exit 4**. The local-build path is untouched.
- **F1 (lib) — dedicated exit-4 arm** in `run_smoke_gate` (`release-gate-lib.sh`) with a distinct `could not pull` diagnostic, so a never-pushed tag / network failure stays a real, *distinguishable* release defect rather than being masked as "first-run broken."
- **F1 (test) — exit-4 truth-table row** + `test_gate_pull_failed_exit4` in `release-gate-logic-test.sh` (stub-driven, no Docker) — the plumbing regression guard.
- **F2/F3 — dispatch noise gating.** `if: github.event_name != 'workflow_dispatch'` on `package-npm`, `build-linux-x64`, `build-linux-arm64`; `create-release` skips via `needs: package-npm`. A dispatch dry-run now reduces cleanly to `build-container-*` → `smoke-*`.

## Testing

- `release-gate-logic-test.sh` 14/14 · `release-tag-parity-test.sh` 13/13 · `bash -n` clean
- Integration smoke 24/24 · no Rust regression (change touches no Rust)
- **Deferred acceptance gate:** the genuine fresh-runner remote-pull proof is the post-PR `workflow_dispatch` dry-run — expect `smoke-amd64`/`smoke-arm64` green (first true both-arch hosted proof, AC-07; watch arm64 cold-boot vs the 90s deadline), `create-container-manifest` + `package-npm`/`create-release`/`build-linux-*` skipped → overall green.

Diagnosis, design review, product review, and gate report are on #795.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   build-linux-x64:
+    if: github.event_name != 'workflow_dispatch'
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
@@ -110,6 +111,7 @@ jobs:
           retention-days: 1
 
   build-linux-arm64:
+    if: github.event_name != 'workflow_dispatch'
     runs-on: ubuntu-22.04-arm
     steps:
       - uses: actions/checkout@v4
@@ -202,6 +204,7 @@ jobs:
 
   package-npm:
     needs: [build-linux-x64, build-linux-arm64]
+    if: github.event_name != 'workflow_dispatch'
     runs-on: ubuntu-latest
     permissions:
       id-token: write
--- a/product/test/infra-001/scripts/docker-http-posture-smoke.sh
+++ b/product/test/infra-001/scripts/docker-http-posture-smoke.sh
@@ -19,6 +19,8 @@
 #
 # Requires Docker. Exits 3 (with a clear SKIP reason) when Docker is
 # unavailable so CI flags it as a deferred step rather than false-green.
+# In IMAGE= mode, exits 4 when the prebuilt tag can neither be pulled nor found
+# locally (tag not pushed / network unhealthy) — distinct from fail()'s exit 1.
 set -euo pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
@@ -59,8 +61,21 @@ if ! command -v docker >/dev/null 2>&1 || ! docker info >/dev/null 2>&1; then
 fi
 
 # -- Build (or reuse) the production image ---------------------------------
+# Precondition for the IMAGE= mode: the caller must set/activate IMAGE; in that
+# mode this script PULLS the tag (or falls back to a present local image) rather
+# than building — the first docker op below is `inspect`, which never pulls, so a
+# cross-runner cache miss would otherwise false-fail the gate (#795).
 if [ -n "${IMAGE:-}" ]; then
   log "using prebuilt image: $IMAGE"
+  # Pull-preferring with local fallback: on a fresh hosted runner this really
+  # pulls the pushed bytes (ADR-002); on a dev box with a locally-built tag (not
+  # in any registry) it falls back to the present image; only when genuinely
+  # unavailable do we exit 4 (distinct from fail()'s exit 1). The `|| ... || {}`
+  # chain is intentionally not routed through fail() and must run all arms before
+  # set -e/pipefail trips.
+  docker pull "$IMAGE" \
+    || docker image inspect "$IMAGE" >/dev/null 2>&1 \
+    || { printf '[783-smoke] FAIL: could not pull %s (confirm the tag was pushed and network is healthy)\n' "$IMAGE" >&2; exit 4; }
 else
   IMAGE="unimatrix:783-smoke"
   log "building image $IMAGE from $REPO_ROOT/Dockerfile (this is slow) ..."
--- a/product/test/infra-001/scripts/release-gate-lib.sh
+++ b/product/test/infra-001/scripts/release-gate-lib.sh
@@ -8,6 +8,7 @@
 #
 # Contracts (VERBATIM — IMPLEMENTATION-BRIEF Data Structures / ADR-003 / ADR-002 / ADR-004):
 #   smoke exit contract : 0 = ran+passed · 1 = ran+failed (fail()) · 3 = self-skipped (Docker absent)
+#                         · 4 = IMAGE= prebuilt tag could not be pulled / not present locally (#795)
 #   positive run-marker : terminal line  [783-smoke] ALL GATES PASSED  (printed only after all gates)
 #   tag resolution      : push    -> TAG="${GITHUB_REF_NAME}"  (UN-stripped, keeps the v) => :v<version>-<arch>
 #                         dispatch -> TAG="latest"                                          => :latest-<arch>
@@ -49,6 +50,7 @@ run_smoke_gate() {
   case "${rc}" in
     0) : ;;
     3) echo "::error::smoke SKIPPED (exit 3): Docker-capable lane mis-provisioned — HARD failure (SR-01)."; return 1 ;;
+    4) echo "::error::smoke FAILED (exit 4): could not pull prebuilt IMAGE — confirm the tag was pushed / network healthy (#795)."; return 1 ;;
     1) echo "::error::smoke FAILED (exit 1): shipped image first-run path is broken."; return 1 ;;
     *) echo "::error::smoke exited unexpectedly (exit ${rc})."; return 1 ;;
   esac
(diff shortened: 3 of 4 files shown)


dug-21:
## Security Review — PR #796 (bugfix-795 release-gate docker pull)

**Risk level: LOW. No blocking findings.** CI/release-gate-only change (shell + Actions YAML); no Rust/JS, no runtime/product surface.

### Scope verified
- Full `main...HEAD` diff read cold; all four changed files read in full; gate-logic truth-table test executed locally (14/14 pass, incl. the new exit-4 row); `bash -n` clean on all three scripts.

### Command injection / unsafe expansion — PASS
- `$IMAGE` is double-quoted in the new `docker pull "$IMAGE"` and in the pre-existing `docker image inspect "$IMAGE"`. No word-splitting/glob exposure.
- `IMAGE` provenance is trusted: it is built by `resolve_image()` from GitHub-controlled context (`GITHUB_REPOSITORY_OWNER`, `GITHUB_EVENT_NAME`, `GITHUB_REF_NAME`) plus a literal `arch` — never from PR/fork-controlled input. This is the **Release** workflow, triggered only by `push: tags: v*` and `workflow_dispatch`, so `GITHUB_REF_NAME` is a repo-owned tag/branch, not attacker-supplied. No injection path.

### `set -euo pipefail` behavior of the new `|| ... || { …; exit 4; }` chain — CORRECT
- The left operands of `||` are not subject to `errexit`, so a failed `docker pull` does not abort; control falls through to the local `docker image inspect`, then to the explicit `exit 4` group. The chain runs all arms as intended; exit 4 is now a *known* code (own `case` arm in `run_smoke_gate` and its own truth-table row) — no longer swallowed by the `*)` unexpected arm.

### Supply-chain / TOCTOU — no new exploitable boundary
- After a successful pull, line 86 `docker image inspect "$IMAGE"` reads the **local** daemon image (pull already fetched the bytes); it does not re-hit the registry, so there is no pull→inspect TOCTOU window inside the gate.
- The local-inspect fallback fires only when `docker pull` *fails*. On a fresh hosted CI runner (no local image) a failed pull → exit 4 (hard gate fail), exactly the #795 signal. The fallback can only "mask" staleness on a dev box that already holds a locally-built tag — the documented, intended dev-loop behavior, not a CI trust hole. GHCR auth (`docker/login-action`) is unchanged.

### Workflow gating as a security property — PASS
- `build-linux-x64`, `build-linux-arm64`, `package-npm` each gain `if: github.event_name != 'workflow_dispatch'`. This only suppresses the **dispatch dry-run** legs; the `v*` push release path is unaffected (`github.event_name == 'push'`).
- `create-release` is skipped on dispatch purely via `needs: package-npm` (a skipped dependency) — confirmed it is **not** a new always-skip `if:`. The real publish chain on a `v*` push is intact; no build that a publish step depends on is dropped on the release path.

### Secrets — PASS
- No hardcoded credentials/tokens added. GHCR/npm auth flows unchanged.

### Blast radius / regression risk
- Worst case if the `||` chain misbehaved: a false exit 4 → gate goes RED → release **blocked** (fail-closed, safe). It cannot false-GREEN a release: exit 4 maps to `return 1` and the terminal `ALL GATES PASSED` marker assertion still gates the success path. The change strictly *removes* a false-RED (the #795 missing-pull false-fail) and adds a fail-closed code; no path becomes more permissive.
- Diff is minimal and confined to the bug (pull-with-fallback + exit-4 plumbing + dispatch gating). No unrelated changes.

LGTM from a security/blast-radius standpoint.

PR #799, opened 2026-06-19
[vnc-040] Per-slug configuration overlay resolution (C6 / Feature A)

dug-21:
## vnc-040 — Per-Slug Configuration Overlay Resolution (C6 / Feature A)

Delivers **C6 — Per-slug configuration** (Unimatrix #5148), Feature A of #785. Each registered slug can now resolve its own `{base_dir}/{slug}/config.toml` overlaid onto the daemon's global config; when no per-slug file exists, behavior is byte-for-byte identical to the current single-project path.

Design: #785 (SCOPE / ARCHITECTURE / SPEC / RISK strategy + ADRs in `product/features/vnc-040/`). All Session-1 gates + Session-2 gates passed.

### What changed
- **`resolve_slug_config` helper** (`http_provision.rs`) — probe → `Cow::Borrowed(global)` on no-file (no merge) → else `load_single_config` → per-file `validate_config` → `merge_configs(global.clone(), slug_file)` → **post-merge `validate_config(&merged)`** (#3905 third-layer fix) → `Cow::Owned`. Reuses existing config fns unchanged (visibility-only `pub` bump).
- **Per-slug loop** (`main.rs`) — derives the overlayable values (categories, boosted_categories, domain packs, confidence, overlayable inference, `nli_top_k/enabled`, **`server.instructions`**) from the resolved config; `Arc::clone`s the 3 global handles (embed/nli/pool) + passes `permissive` **unconditionally outside any overlay branch**. Relocated the `instructions` source from the pre-loop hoist into the loop (FR-14, #785's explicit per-slug knob).
- **ADR-004 canonical classification** (`infra/config.rs`) — a data-only `PER_SLUG_CONFIG_CLASSIFICATION` registry + `is_per_slug_overlayable`; the single owner of the per-slug-vs-global split. `merge_configs` is **not** rewritten.

### Invariants (by construction)
- Exactly one NLI model + one embedding model + one rayon pool at any N slugs (crt-056 AC-2) — handles sourced only from global `Arc`s, never from a merged config. `Arc::ptr_eq` machine-checked on the no-file arm.
- `permissive` / transport / `[embedding]` descriptor stay global. `*_sha256` pins stay global-wins.
- **AC-11 drift-guard test** pins `merge_configs`' real behavior to the registry — a field merged but not classified (or mis-classified) breaks the build.

### Tests
- 34 component tests (registry 11 / helper 13 / loop 10), all green. lib+bin suite 4250+98, 0 failed.
- Integration: smoke **24/24**, regression (protocol/tools/lifecycle/confidence) **0 failed** (pre-existing xfail only, none added here).
- All 14 risks / 32 scenarios mapped to passing tests; all 11 ACs PASS. Gates 3a/3b/3c PASS.

### Scope notes
- **Feature A only** — operator hand-places the per-slug file. Feature B (seeding) is a tracked follow-up; B will *render* its seed annotations from this registry (one-way A→B contract).
- **Accepted residual (R-13):** a per-slug file setting a global-locked section is silently ignored at the seam (only `*_sha256` divergence warns); a seam-level warn is an explicit future enhancement.
- **Follow-up:** infra-001 is single-project; per-slug coverage is the in-crate N=2 model-free harness + unit/construction tests. Multi-slug harness uplift tracked separately.


dug-21 (commented):
## Security Review (uni-security-reviewer, fresh context) — Risk: LOW, non-blocking

Reviewed the Rust diff cold against ARCHITECTURE.md and RISK-TEST-STRATEGY.md, verifying claims against the actual code.

### Verified controls (held)
- **Path traversal** — CLOSED by construction. `ProjectSlug::try_from` enforces `^[a-z0-9][a-z0-9-]{0,62}$`; `.`/`/`/`\`/`%` cannot pass, so `base_dir.join(slug.as_str())` cannot escape `{base_dir}/{slug}/`.
- **Hash-pin global-wins** (`embedding_model_sha256`, `nli_model_sha256`) — HELD. `merge_configs` ignores the per-slug pin when a global pin exists and `tracing::warn`s on mismatch. Model-substitution attack defused.
- **`permissive` global-locked** — HELD by construction. Computed from global `config.agents.default_trust` (main.rs:688), passed unchanged; the loop never reads `resolved.agents.default_trust`. A per-slug `default_trust = "permissive"` cannot escalate the daemon at this seam.
- **Transport/TLS/http global-locked** — HELD by construction. The loop reads ONLY instructions/inference/knowledge/observation from `resolved`; never transport fields.
- **Single model in memory** — HELD. The 3 handles are `Arc::clone`d unconditionally outside the overlay branch, never sourced from `resolved`.
- **Post-merge cross-field re-validation (#3905)** — HELD. `validate_config(&merged, &path)` runs after `merge_configs`, before return; sum-of-six + per-field range checks present.
- **64 KiB cap + TOML deser + fail-loud errors + no `.unwrap()`** — all present via reused `load_single_config`; slug-named `ServerError::Config` at startup.
- **No-file fallthrough** — preserved (`Cow::Borrowed(global)`, byte-for-byte); `merge_configs` logic unchanged; `pub` change is visibility-only.
- **Dependencies** — no Cargo.toml/lock change; no new deps. **Secrets** — none.

### Findings (both non-blocking)

**F1 (low) — `knowledge.adaptive_categories` is a seam input but absent from the canonical classification.** The loop derives `slug_categories` from both `r.knowledge.categories` AND `r.knowledge.adaptive_categories`, and `merge_configs` merges `adaptive_categories` project-wins — yet it appears in neither `PER_SLUG_CONFIG_CLASSIFICATION` nor the `EXPECTED_CLASSIFIED_KEYS` exhaustiveness list. The drift-guard test checks the registry against a hand-maintained expected list, not against the live fields the loop reads from `resolved`, so the "closed set" guard cannot catch this. No security impact (adaptive_categories is the same trust class as categories; overlayable is correct), but it is the R-07/R-14 "seam input absent from the checklist" pattern recurring. Recommend adding the key to both lists and, longer-term, deriving the exhaustiveness baseline from live seam reads.

**F2 (info, pre-existing) — `check_permissions` hard-rejects world-writable (`0o002`) but only WARNS on group-writable (`0o020`),** whereas the docs frame the control as `mode() & 0o022` rejection. Reused unchanged from the global/project config path; consistent daemon-wide; not a regression in this PR.

### Blast radius
Confined to one slug's ServiceLayer at startup. A bad per-slug file fails the daemon loud (no partial serve) — safe failure mode, never silent corruption. Cannot load a 2nd model, alter another slug/the global config, escalate permission posture, change transport/TLS, or bypass a hash pin.

Approving from a security standpoint — no changes required to merge.

Issue #788, opened 2026-06-19
feat(personal-cloud): standing release gate — run the docker build→boot→register→per-slug-write smoke on every release


dug-21:
**nan-019** — wire the existing `docker-http-posture-smoke.sh` (#786) into `.github/workflows/release.yml` as a standing, **verify-by-name** release gate. **NFR-maintenance feature**: advances **N5** ("the shipped artifact is always deployable as released"), guards **N3** ("writes are integrity-protected — never mis-routed across projects"). Scope doc: `product/features/nan-019/SCOPE.md`.

## Problem
The release ships the GHCR multi-arch image, but nothing exercises the **actual shipped image** end-to-end the way an operator would. Unit/integration tests check code in isolation; no step builds the released image, boots it, registers a project, writes over the served HTTPS route, and confirms the write lands in the **per-slug** store. An image can pass every test and still be broken on first run — the root of #774 (host-allowlist never wired) and #783 (booted HTTP-off, silently mis-routed writes to the hash store). Both shipped green because no test ran the real artifact's first-run path. The fix for the class exists (`docker-http-posture-smoke.sh`); the gap is it isn't wired as a standing gate.

A second failure mode: the smoke **self-skips** (`exit 3`) when Docker is absent. Wired naively, a skip passes green — re-creating the #4796/#4970 false-green class. The gate must verify the smoke **actually ran**, not merely that the job exited 0.

## Goals
1. Shipped image exercised end-to-end on **every release, on BOTH arches** (amd64 + arm64) against the **pushed** bytes: boot-HTTP-on → register → restart → per-slug HTTPS write → assert-landed → assert per-slug store grew & hash store did not.
2. A smoke failure **blocks the release**.
3. **Verify-by-name + positive run-marker**: a self-skip (`exit 3`) is a hard failure; the gate asserts the terminal `ALL GATES PASSED` line so an early-exit-0 cannot masquerade as success.
4. **ADR-004 container-lane independence** preserved — smoke/ARM64 flakiness cannot block (or be blocked by) the binary/npm branch.
5. **No silent retry** — a flaky deployability gate is itself a signal.

## Non-Goals
- No new deploy behavior or container changes — the gate only *verifies* what ships.
- Not the functional per-slug analytics work (crt-056 / #787); orthogonal.
- Not a new/rewritten smoke — wires the existing one (only change: AC-05 grew-assertion).
[...]

PR #810, opened 2026-06-20
fix: /shared read-write default + first-boot embed-readiness release gate (#767)

### LINKED ISSUE #767: bug(personal-cloud): container first boot can't load embedding model — no model baked + compose ships /shared:ro
## Summary
Fresh `docker compose up` cannot load the embedding model → `context_store`/search unusable. Three artifacts disagree about how models arrive.

## Evidence (first-boot log)
```
ERROR unimatrix_server::infra::embed_handle: embedding model failed to load error=io error: Read-only file system (os error 30) attempt=1
```

## Root cause (three-way inconsistency)
- `Dockerfile` bakes **no** model — line 99 only `mkdir -p /shared/models`; no `model-download`/`.onnx` COPY. (Only the ONNX *runtime lib* is fetched, line 67.)
- `README.md:62` claims *"Both ONNX models … baked into the image — no internet access required after pull."* — false.
- `docker-compose.yml` mounts `unimatrix-shared:/shared:ro`, so the runtime download into `/shared/models` hits a read-only FS.

## Solution directions — APPROACH OPEN, choose by blast radius
Product owner leaned toward download-on-first-use this session, but the call is not locked — weigh image size vs offline guarantee:
- **(a) Compose RW + doc fix:** ship `/shared` read-write by default; correct `README.md:62` to "downloaded on first boot (internet required); pre-populate the volume for air-gap." Smallest change; loses offline-after-pull.
- **(b) Bake models into the image:** add a `model-download` step in the Dockerfile builder before the `/shared` COPY. Restores offline + air-gap and makes `:ro` honest, at the cost of image size. The vision-aligned option if air-gap matters.
- **(c) Entrypoint pre-populate:** baked cache → copy into the volume on first boot, then run RO. Middle ground; adds boot logic.

## Vision alignment
`goal:personal-cloud` success criterion *"docker compose up deploys a working instance with ONNX models auto-populated."* Goal entry stays accurate under any option (air-gap remains reachable via pre-population). Found during first dogfood.

dug-21:
## Summary
Fixes #767 — fresh `docker compose up` could not load the embedding model, breaking `context_store`/`context_search`.

**Root cause:** drift/regression from ADR-003 nan-015, which decided `/shared` defaults to **read-write** (`:ro` is optional hardening *after* population). The shipped `docker-compose.yml` had drifted to `/shared:ro`; with no model baked into the image (deliberate, nan-015), first boot had nowhere to write the model download → `fs::create_dir_all` (EROFS) → embed `EmbedFailed` after retries. Same class as bugfix-783 (lesson #5130): deploy posture in compose whose precondition the shipped image never satisfies.

## Changes
- **`docker-compose.yml`** — `/shared:ro` → `:rw` (restores ADR-003 nan-015; the throwaway one-liner).
- **`README.md`** — remove false "models baked into the image / no internet after pull" claim; document air-gap as a **pre-populate-then-`:ro` procedure** (no new false absolute; UID-65532 writable caveat).
- **`product/test/infra-001/scripts/docker-embed-readiness-smoke.sh`** *(new — the substantive deliverable)* — first-boot gate asserting a **real `context_store` → `context_search` embed round trip** (not `/health` liveness, which passes on the broken state). Mounts `/shared` in the **compose-declared mode** (a `:ro`-without-prepopulate drift fails the gate), waits past the 10s/20s/40s retry window (180s), and **SKIPs loudly** (`exit 3`, never a silent pass).
- **`.github/workflows/release.yml`** — wire `embed-amd64`/`embed-arm64` gate jobs; `create-container-manifest` now depends on them. On the **release gate (#788), not per-PR CI**.

## Verification
- Embed-readiness smoke **ran for real** (Docker + HF reachable): full build + model download + live store→search round trip → **PASS**.
- Gate-logic self-test **15/15**; pytest smoke **24/24**; compose-mode regression detection verified (`:ro` drift → gate red); diff confirmed non-Rust (clippy/cargo skip authorized).
- Gate: Bug Fix Validation — **PASS** (13/13).

## GH Issue
Closes #767


dug-21 (commented):
## Security Review — #810 (uni-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings.** Reviewed the full diff cold against the source, the sibling http-posture smoke, release.yml, and ADR-003 nan-015 #4652.

### Security boundary: `/shared` `:ro` -> `:rw`
Not a new weakening. This restores ADR-003 nan-015 #4652 (default `:rw`; `:ro` is optional post-populate hardening), which the shipped compose had drifted away from. `:rw` is strictly more permissive than `:ro`, but the *only* residual exposure — the embedding model has no load-time hash check (#651) — was explicitly accepted in #4652's Consequences. SR-02 (supply-chain) is **not** lowered below the ADR-sanctioned baseline. NLI verify-then-load (`nli_model_sha256`) is untouched. The air-gap `pre-populate -> :ro` path is preserved and now correctly documented in both compose comments and README (replacing the false "baked into the image" claim). Net: drift-correction, not a posture change.

### Bash smoke script (`docker-embed-readiness-smoke.sh`) — clean
- `set -euo pipefail`; no `eval`, no `curl | sh`, no `docker exec` into the distroless image (busybox sidecars only, mounted `:ro`).
- Compose-mode parse is fixed-pattern `grep`/`printf`, no eval of file content. All shell expansions quoted. PID-namespaced volume/container names. `trap cleanup EXIT` removes containers, volumes, and the `mktemp -d` temp dir.
- Token/cert pulled from the data volume into a `mktemp -d`; token is `tr`-stripped and **never echoed** to logs (only `SESSION_ID` is logged, which is non-secret). No model-download tokens are used (anonymous HuggingFace pull). No hardcoded credentials anywhere.
- LOUD skip (exit 3) on Docker/network absent — preserves the deferred-step contract; never a silent green.

### release.yml `embed-amd64` / `embed-arm64` — clean
Byte-for-byte structural clones of the existing `smoke-amd64`/`smoke-arm64` jobs: same `actions/checkout@v4` + `docker/login-action@v3` pins (no new action sources), same `secrets.GITHUB_TOKEN`, only repo-controlled `GITHUB_*` context flows into `run:` steps (no attacker-supplied input). No per-job `permissions:` elevation — they inherit the workflow-level `contents/id-token/packages: write`. Gating `create-container-manifest` on the embed jobs is correct.

### Blast radius / regression
- Worst case of a subtle bug in the compose flip: near-nil — a single mount-mode flag, and compose is not a shipped release artifact (lesson #5130); the bare `docker run` path was never `:ro`.
- New gate downloads ~166MB from HuggingFace and waits past the 10s/20s/40s embed backoff (~70s+) — correctly release-only (not per-PR) and backoff-aware, so no per-PR cost and low flake risk. Budget a post-PR CI/network round-trip for the new lane.
- No new runtime dependencies; no Rust changes.

### Non-blocking hardening notes
1. `READY_TIMEOUT_SECS` env is interpolated into shell arithmetic (`deadline=$(( ... + READY_TIMEOUT_SECS ))`); a non-numeric value would error under `set -e` (loud, not exploitable). CI sets it, so informational only.
2. `mktemp -d` uses the default `$TMPDIR`; on a shared runner a world-readable temp dir would briefly hold the bearer token+cert. On GitHub hosted runners this is single-tenant — acceptable. If ever run on multi-tenant infra, `chmod 700` the temp dir.

Verified: full diff read, root-cause/design/zero comments cross-checked against source, OWASP (injection/path/secrets/supply-chain) evaluated, blast radius named. Approving — advisory; human makes the call.

Issue #815, opened 2026-06-20
[infra-001] test_no_new_smoke_script: stale R-15 invariant — #810 legitimately added a 2nd smoke script


dug-21:
## Discovered By

Static gate test: `product/test/infra-001/scripts/release-gate-bundle-static-test.sh::test_no_new_smoke_script` (R-15, nan-020).

Discovered during #812 verification (uni-tester). **Not caused by #812** — #812 is a shell-only diagnostic-capture change to `docker-http-posture-smoke.sh`. The failure reproduces identically with the #812 change stashed (11 passed / 1 failed at HEAD).

## Expected Behavior

`ls scripts/*smoke*.sh | grep -v stub-smoke.sh | wc -l` == 1 (only `docker-http-posture-smoke.sh`). The R-15 invariant asserts the nan-020 bundle round-trip lives inside the single existing smoke, with no second smoke script.

## Actual Behavior

Found 2 smoke scripts:
- `docker-http-posture-smoke.sh` (the nan-019/nan-020 posture smoke)
- `docker-embed-readiness-smoke.sh` — added in commit **e7bab4fb (#810)** "first-boot embed-readiness release gate (#767)", a separate already-merged feature.

The R-15 invariant predates #810 and was not updated when #810 legitimately introduced a distinct embed-readiness smoke. The assertion is now stale, not the codebase.

```
== R-15 no new smoke script; round-trip lives in the extended smoke (static) ==
  FAIL test_no_new_smoke_script
       expected exactly 1 smoke script, found 2
```

## Reproduction

```bash
cd /workspaces/unimatrix
bash product/test/infra-001/scripts/release-gate-bundle-static-test.sh
# -> 11 passed, 1 failed (test_no_new_smoke_script)
```

## Suggested Fix (separate PR)

Update `test_no_new_smoke_script` to allowlist `docker-embed-readiness-smoke.sh` (the #810 gate) alongside `stub-smoke.sh`, OR re-scope R-15 to "no second *posture/bundle* smoke." The intent — that nan-020's bundle round-trip stays inside `docker-http-posture-smoke.sh` and is not split into a parallel script — remains valid; the glob is simply too broad now that an unrelated embed-readiness smoke exists.

## Notes

Pre-existing, introduced by #810. Not in scope for the #812 bugfix PR. The #812 diff itself is shellcheck-clean and passes all 19 bundle gate-logic assertions.

dug-21:
[...]

Issue #814, opened 2026-06-20
Add real-server `init --bundle` round-trip to the delivery gate (prevention for #812)


dug-21:
## Why
#812 (release smoke Gate 6: `init --bundle` rc=1) was a release-only, first-run-on-tag failure. The bundle → transport → cert-pin attach path has **no end-to-end test against a real booted server** in the delivery gate — pin/transport bugs hide behind synthetic certs (the vnc-034 F1 lesson, #4965). This is the second release-only gate in two turns to carry unverified risk to the tag.

## What
Add a real-server `init --bundle` round-trip test to the **delivery gate** (not a follow-up nice-to-have): boot a real server with HTTP enabled, emit a v:2 bundle, run the shipped `init --bundle` against it, assert the pinned Ping succeeds — exercising the bundle/transport/cert-pin path **pre-tag**. Build on the #5098 real-server harness.

## Acceptance
- Delivery gate runs a real-server `init --bundle` round-trip (no synthetic certs masking pin/transport behavior).
- Future changes to the bundle/transport/cert-pin path are validated before tag, not first-run-on-release.

## Refs
- #812 (release smoke Gate 6 failure — the bug this prevents)
- #4965 (vnc-034 F1 lesson: synthetic certs hide pin bugs)
- #5098 (real-server harness to build on)
- nan-019 (#788 smoke gate), nan-020 (pinned-Node init --bundle leg)

_Filed by the #812 bugfix session per human direction — prevention is the durable fix; scope/deliver separately._

dug-21:
Capability stakes (uni-zero): this is the prevention leg for #812 — moving the real-server `init --bundle` round-trip into the **delivery** gate (not just release smoke) advances **N5 — the shipped artifact is always deployable as released** (#5254, currently `partial`). Without it, a C9 bundle-attach regression can reach a tag again, exactly as #812 did.

Issue #798, opened 2026-06-19
chore(clippy): zero out workspace clippy warnings + lock the baseline


dug-21:
## Problem

`cargo clippy --workspace --all-targets` emits **~541 warnings** (with target duplicates; a few hundred unique). The delivery/bugfix gate runs `cargo clippy --workspace -- -D warnings` (uni-delivery-protocol.md:570), so every feature swarm wades through this pre-existing noise, flags it, and skips it as out-of-scope. The signal-to-noise on real, feature-introduced lints is gone.

Goal: drive the workspace to a **clean clippy baseline** and **lock it** so future warnings are actually about the feature in flight.

There is currently **no `[workspace.lints]` config** and no `-D warnings` enforcement outside the protocol gate — nothing prevents re-accumulation.

## Top categories (by raw count)

| count | lint |
|------:|------|
| 116 | collapsible `if` |
| 50 | assertion with constant value |
| 42 | field assignment outside `Default::default()` initializer |
| 25 | manual `RangeInclusive::contains` |
| 20 | non-binding `let` on a future |
| 17 | doc list item without indentation |
| 14 | unneeded `return` |
| 10 | `assert_eq!` with literal bool |
| 9 ea | `map_or` simplify · auto-deref · clamp pattern · too-many-arguments |
| — | useless `vec!`, `sort_by_key`, unused imports, redundant guards, … |

Per-crate concentration: `unimatrix-server` (~176 lib + ~333 lib-test), `unimatrix-observe` (~53), `unimatrix-engine`, plus scattered `unimatrix-store` migration tests.

## Scope — mechanical only, in phases

**Phase A — autofix.** Run clippy's own fixes, then review the diff.
```
cargo clippy --workspace --all-targets --fix --allow-dirty
```
Covers collapsible-if, needless-return, redundant-closure, unused-imports, useless-vec, `map_or`, `RangeInclusive::contains`, etc. (~200+ suggestions are auto-applicable.)

**Phase B — manual mechanical.** Hand-resolve the remaining trivially-correct lints autofix won't touch (constant-value assertions, `assert_eq!(x, true)` → `assert!(x)`, `Default::default()` field-init reshaping, non-binding `let _ = fut;`, doc-list indentation).

**Phase C — structural lints: DO NOT force-refactor.** These change signatures/types and may be genuinely feature-relevant — they are **not** "just a clippy run":
- `too_many_arguments` (8–12 args)
- `type_complexity` ("very complex type")
[...]

PR #816, opened 2026-06-20
fix: capture init --bundle stderr in release smoke Gate 6 (#812)

dug-21:
## Summary
Release smoke **Gate 6** (`product/test/infra-001/scripts/docker-http-posture-smoke.sh`) failed on arm64 with `init --bundle` rc=1, but `consume_bundle()` routed the child's stderr to `/dev/null` — the failure was **undiagnosable by construction**. This is **Step 1 (diagnostic capture) only**.

## Change
- `consume_bundle()`: capture the `init --bundle` child's stderr+stdout to `$SANDBOX/init.out` (was `>/dev/null 2>&1`) — a file under the hermetic sandbox reaped by the existing `cleanup()` trap; redirect to a file (not a pipe) so `rc` capture and `set -e`/`pipefail` are unaffected. Both real-node and `SMOKE_INIT_CMD` stub branches capture identically.
- Gate 6: on `init_rc != 0`, dump a tail-bounded (last 50 lines) excerpt to the smoke log under the `[783-smoke] init:` prefix, then the existing `fail()` (exit 1). Exit semantics unchanged.
- `emit_bundle`'s token-bearing stderr (L78) left **untouched** — no token can reach the logs.

## Why Step 1 alone
The client `init --bundle` path reproduced clean (rc=0) on Node 24 — the bug is environmental to the CI arm64 published-port topology. The real corrective (Step 2) is **conditional** on the mechanism the captured stderr reveals on the next tagged run:
- connect/timeout via dual-stack `localhost`→`::1` + dropped v6 SYN + Happy-Eyeballs > 750ms → **a same-host operator hits the identical failure**. Open product question: do we support client-and-server-same-box `localhost` attach? If yes, the durable fix is **client-side loopback address-family handling** (hot-path-safe), not a harness IPv4-pin and **not** a `connectMs`/`syncMs` widen (those fire fail-open on every production hook event).
- fp-mismatch / bad-Pong → real client/server defect at the cert-pin/transport site.

See #812 human-checkpoint comment for the full decision tree.

## Validation
- Stub-seam dual-branch (failure dump surfaced under `[783-smoke] init:` rc=1; success quiet, no dump) — PASS
- Integration smoke suite 24/24 PASS; gate-logic stub-driver 19/19 PASS
- shellcheck: new lines clean (3 findings proven pre-existing); `bash -n` OK
- Gate 3 Bugfix Validation: **PASS** (9/9)
- Token-leak check: capture scoped to init child only, dumped on failure only — PASS

## Notes
- #812 alone does **not** unblock v0.8.3 — the x64 anndists `simdeez_f` compile break (via #798) is a separate failure that must also land before a clean re-tag.
- Pre-existing static-gate failure `test_no_new_smoke_script` (caused by #810, not this change) filed as #815.
- Prevention follow-up: real-server `init --bundle` delivery-gate round-trip filed as #814.

## GH Issue
Refs #812

### DIFF
--- a/product/test/infra-001/scripts/docker-http-posture-smoke.sh
+++ b/product/test/infra-001/scripts/docker-http-posture-smoke.sh
@@ -80,15 +80,24 @@ emit_bundle() {
 }
 
 # consume_bundle <blob> -> rc only (hermetic; HOME on the CHILD, never in-process).
+# #812: the init child's stderr+stdout are NO LONGER discarded — they are captured
+# to $SANDBOX/init.out (reaped by the cleanup() trap, hermetic). The file is the
+# diagnostic surface the caller tail-dumps ON FAILURE ONLY (Gate 6 was undiagnosable
+# by construction while this went to /dev/null). Safe to capture: init's stderr on a
+# Ping failure is a connect/URL/fingerprint string and NEVER the token (NFR-06; the
+# bearer is flushed to the request body only after the pin verifies, never to a
+# stream). This captures ONLY the init child — emit_bundle's blob-bearing stderr
+# (L78) stays dropped. Redirect to a FILE, never a pipe, so set -e/pipefail and the
+# rc capture (#4873 class) are unaffected.
 consume_bundle() {
   local blob="$1"
   if [ -n "${SMOKE_INIT_CMD:-}" ]; then
     # shellcheck disable=SC2086
-    HOME="$SANDBOX/home" $SMOKE_INIT_CMD "$blob" "$SANDBOX/proj" >/dev/null 2>&1
+    HOME="$SANDBOX/home" $SMOKE_INIT_CMD "$blob" "$SANDBOX/proj" >"$SANDBOX/init.out" 2>&1
   else
     HOME="$SANDBOX/home" \
       node "$REPO_ROOT/packages/unimatrix/bin/unimatrix.js" \
-        init --bundle "$blob" --project-dir "$SANDBOX/proj" >/dev/null 2>&1
+        init --bundle "$blob" --project-dir "$SANDBOX/proj" >"$SANDBOX/init.out" 2>&1
   fi
 }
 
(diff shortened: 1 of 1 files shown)


dug-21:
## Security Review — PR #816 (fresh-context) — Risk: LOW · No blocking findings

Reviewed the full diff cold and independently verified the central token-leak claim against the shipped client source (not just the diff narrative). The Step-1 diagnostic capture is token-safe end-to-end.

### Token-leak verification (the central risk) — CLEAR
Traced the only stream the new code can dump (the `init --bundle` child's stderr/stdout):
- `bin/unimatrix.js:30` writes `"unimatrix init failed: " + error.message` on the catch path.
- The Ping-failure throw (`init.js:533`) is `"Remote validation failed: " + res.message + "\n…"`.
- `res.message` is composed by `transport-http.js actionable()` — host-only strings via `safeHost()` (strips to `host`, never the full URL, never the token): `"token rejected (HTTP N) — check --token"`, `"cannot reach <host> — check --remote URL"`, or `"server returned HTTP N"`. The token is never interpolated into any message.
- The token-bearing request body is flushed only via `req.end(body)` AFTER fingerprint verification (`transport-http.js:171`); no failure path writes it to any stream (NFR-06 holds).
- `emit_bundle`'s blob-bearing stderr — the one stream that DOES carry the token — is dropped at L78 and is **untouched** by this diff. Confirmed the new capture is scoped to the `init` child only.

### Shell hardening — CLEAR
- **Hermeticity:** capture target `"$SANDBOX/init.out"` lives under `mktemp -d` (L160) and is reaped by the `cleanup()` EXIT trap (`rm -rf "$SANDBOX"`, L41). No leftover artifacts; not `/tmp`-root or `$REPO_ROOT`.
- **rc/`set -e` safety:** redirect is to a **file, not a pipe**, so `init_rc=$?` and `set -e`/`pipefail` are unaffected (avoids the #4873 pipe-swallow class). Both the real-node and `SMOKE_INIT_CMD` stub branches capture identically.
- **Dump is failure-path only** (guarded by `if [ "$init_rc" -ne 0 ]`); happy path stays quiet — no stdout pollution, no incidental logging.
- **No format-string injection:** `log() { printf '[783-smoke] %s\n' "$*"; }` passes `$line` as a printf **argument**, so init output containing `%s`/`%n` cannot be interpreted. `read -r` prevents backslash mangling. `tail -n 50` bounds CI-log volume. `[ -s ... ]` guards the empty-capture case.
- **No injection/eval/unbounded dump**, no new dependencies, no hardcoded credentials, minimal diff (one file, +27/-4).

### Blast radius / regression — LOW
Harness-only. Cannot reach the shipped client or the Rust workspace. Worst case of a subtle bug = a broken redirect tripping the gate's exit-code contract (ADR-001 numerics 0/1/3/4 + terminal marker) — but the file-not-pipe redirect preserves `init_rc`, exit semantics are unchanged, and the dump is additive on the already-failing path. The PR notes the unrelated pre-existing `test_no_new_smoke_script` failure (#810 → #815) and the x64 anndists break — neither caused by this change.

### Note (non-blocking, informational)
`tail | while read` runs the loop in a subshell; harmless here since nothing inside mutates outer state. No action needed.

**Verdict:** Step-1 capture is correctly scoped, hermetic, and token-safe. Approve from a security standpoint.

PR #817, opened 2026-06-21
fix: set UNIMATRIX_PUBLIC_URL on smoke emit container + guard placeholder bundle (#812)

dug-21:
## Summary
**Step 2 of #812** — the actual corrective. Release smoke **Gate 6** (`init --bundle`) failed on **both arches** with `Remote validation failed: cannot reach (invalid URL)`. Step 1 (#816) made it diagnosable; the next tag's captured stderr revealed the real cause.

## Root cause (HIGH confidence, fully reproduced — harness bug)
The throwaway `client-bundle` **emit container** (`docker-http-posture-smoke.sh` ~L78) ran **without** `-e UNIMATRIX_PUBLIC_URL`. Env vars don't cross containers, so the server's `derive_public_url()` fell back to its `<EDIT-ME>` placeholder and baked `observe_url = https://<EDIT-ME>:8443/...` into the v:2 bundle. The host client's `new URL()` then threw `ERR_INVALID_URL` → rc=1. The boot container (L290) and Gate-2 curl (L345) already set the value; the emit container was the one step in the two-runtime split that dropped it.

Disproves the earlier IPv6/connectMs hypothesis (the client never reaches a connect attempt — the URL is structurally invalid) and is arch-independent (amd64 reached Gate 6 once #798's compile fix landed and failed identically).

## Changes (harness-only, one file)
1. **Fix:** emit container now sets `UNIMATRIX_PUBLIC_URL="https://localhost:${PORT}"`, matching the boot container exactly.
2. **Gate-5 emit-time guard:** decode the emitted bundle; `fail()` (exit 1, **no new exit code** — append-only per nan-019 ADR-001) if `observe_url` carries `<EDIT-ME>` or doesn't parse. Converts this whole bug class into a precise emit-time failure (`bundle carries placeholder URL — emit container missing UNIMATRIX_PUBLIC_URL`) instead of an opaque Gate-6 `(invalid URL)`.

No `packages/unimatrix/**`, no `crates/**`, no timeout/IPv4 change.

## Validation
- Gate-5 guard: fires on `<EDIT-ME>` (rc=1, exact message) / passes on valid URL / parse-branch on malformed — verified
- gate-logic stub test 19/19; integration smoke 24/24
- shellcheck: **0 new** findings (3 pre-existing baseline); `bash -n` clean
- Token-safe (guard inspects `observe_url` only); append-only exit contract intact
- Gate 3 Bugfix Validation: **PASS** (9/9)

## Notes
- Should green Gate 6 on both arches and let `create-container-manifest` publish.
- Pre-existing static-gate failure `test_no_new_smoke_script` (caused by #810) tracked in #815 — not touched here.
- Durable prevention: real-server `init --bundle` delivery-gate round-trip tracked in #814 (this guard complements it).

## GH Issue
Refs #812

### DIFF
--- a/product/test/infra-001/scripts/docker-http-posture-smoke.sh
+++ b/product/test/infra-001/scripts/docker-http-posture-smoke.sh
@@ -75,7 +75,14 @@ emit_bundle() {
     # shellcheck disable=SC2086
     $SMOKE_EMIT_CMD 2>/dev/null
   else
-    docker run --rm -v "$VOL:/data" "$IMAGE" --project-dir /data client-bundle "$SLUG" 2>/dev/null
+    # #812: the emit container MUST receive UNIMATRIX_PUBLIC_URL — env does not cross
+    # containers, so without it the server bakes the `<EDIT-ME>` placeholder into the
+    # v:2 bundle's observe_url and the host `init --bundle` rejects it as an invalid
+    # URL. Same scheme/host/port as the boot container (L290) and Gate 2 curl: one
+    # consistent source of truth across boot / curl / emit (nan-020 ADR-002 #5256).
+    docker run --rm -v "$VOL:/data" \
+      -e UNIMATRIX_PUBLIC_URL="https://localhost:${PORT}" \
+      "$IMAGE" --project-dir /data client-bundle "$SLUG" 2>/dev/null
   fi
 }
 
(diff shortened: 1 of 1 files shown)


dug-21:
## Security Review — #812 Step 2 (uni-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings.** Harness-only change to `docker-http-posture-smoke.sh` (+31/-1). Independently verified `bash -n` clean and shellcheck = 0 new findings (3 pre-existing on L209/L275/L341, all outside the diff hunks).

### Central risk — token/secret leakage: CLEAR
The bundle schema (`lib/hook-client/bundle.js`) is `{v, mcp_url, observe_url, token, fp}` — the bearer `token` is a **distinct field** from `observe_url`. The new Gate-5 guard:
- Feeds `$BUNDLE` to node over **stdin** (`printf '%s' "$BUNDLE" | node -e '...'`), extracts and writes **only** `String(j.observe_url||"")` to stdout. The token field is never read.
- On any decode error: `catch{process.exit(0)}` writes nothing — a malformed/token-bearing blob never echoes the JSON.
- node stderr is `2>/dev/null`, so a runtime error pre-`catch` cannot leak either.
- Both `fail()` messages and the PASS `log` are **static strings** — no variable interpolation, so neither the blob, the JSON, nor the token can surface in failure/log output.
- `emit_bundle`'s token-bearing stderr stays `2>/dev/null` (L85), unchanged. The Gate-6 init dump (L226-234) tails only `init.out`, never the blob.

### Injection / safe decode: CLEAR
- Decode (L194-195): bundle content arrives via stdin; the node source is a single-quoted literal — no shell expansion of attacker-influenceable content into the script. No `eval`.
- URL parse (L199): `node -e 'new URL(process.argv[1])' "$OBSERVE_URL"` passes the value as **argv** (`process.argv[1]`), not interpolated into the `-e` source — no JS injection from URL content. `$OBSERVE_URL` is only the `observe_url` field.
- `$BUNDLE` is quoted at every use (L168/176/194/218).

### Fail-closed semantics: GOOD
Empty/failed decode → `OBSERVE_URL=""` → `new URL("")` throws → `fail()` (RED). A broken decode cannot silent-green the gate. The guard correctly runs on the real emit path only (`[ -z "${SMOKE_EMIT_CMD:-}" ]`), so the stub seam's synthetic non-base64url blobs aren't spuriously rejected.

### Hermeticity / artifacts: CLEAR
The new block creates no files or temp dirs (pure pipe to node); nothing to leak or clean up. Existing `cleanup()` trap unaffected.

### Blast radius / regression: LOW
- Cannot affect the shipped client or the Rust workspace — no `packages/unimatrix/**`, no `crates/**`. Test-harness only.
- Folds into the existing `fail()`/exit-1 contract — no new exit code (append-only, nan-019 ADR-001 intact).
- Worst case if subtly wrong: the smoke gate false-fails (blocks a release tag) — a **safe** failure mode (loud, fail-closed). It cannot false-green or corrupt data.
- The fix itself (`-e UNIMATRIX_PUBLIC_URL`) mirrors the boot container (L320) and Gate-2 (L375) — one consistent source of truth; no divergent value introduced.

### Hardcoded credentials: NONE
`UNIMATRIX_PUBLIC_URL="https://localhost:${PORT}"` is a non-secret loopback URL identical to existing lines. No keys/tokens added.

**Verdict: approve from a security standpoint.** No changes required.

Issue #819, opened 2026-06-21
Observability: observe that Acks 204 but persists nothing is silent


dug-21:
## Summary
A behavioral gap surfaced while fixing #818: an `/observe` POST can return **HTTP 204 (Ack)** while persisting **nothing** to the knowledge store, with no trace or counter. This is **correct idempotent behavior** (e.g. SessionClose for an unregistered session — `infra/session.rs:946` `sessions.remove()` → None, write skipped at `uds/listener.rs:2184`, still 204 at `http/router/observe.rs:65-72`), but it is **unobservable**.

## Why it matters
This silent Ack-but-no-write is precisely what hid the Gate 7 bug (#818) for the entire lifetime of that gate — the smoke saw 204 and assumed the observe landed. An operator (or a future test) has no signal distinguishing "accepted and persisted" from "accepted and dropped."

## Proposed (lightweight, server-side, separate from #818)
- Emit a debug-level trace and/or increment a counter when an observe Acks (204) but produces zero store writes (e.g. close/drain on an unregistered/already-drained session).
- Goal: make the no-op observable in logs/metrics without changing the idempotent 204 contract.

## Scope / priority
- Lower priority; **not** folded into #818 (which is the harness-only Gate 7 fix).
- No change to the 204 idempotency semantics — observability only.

## Refs
- #818 (Gate 7 harness fix that exposed this)
- `crates/unimatrix-server/src/infra/session.rs:946`, `uds/listener.rs:2184`, `http/router/observe.rs:65-72`
- Related drain mechanics: per-turn Stop→SessionClose drain (build-request.js:59-66 → listener.rs drain path)

dug-21:
Capability stakes (uni-zero): **silent data loss** — a 204 ack with zero persistence is a *false success*, the nastier class: the operator sees success and loses the signal. Touches **C11 — remote observe over HTTPS** (#5153) and inverts the spirit of **N4 — a healthy deployment emits no false-alarm signals** (#5162): here the deployment emits a false *success*, which is worse than a false alarm. Labeled goal:personal-cloud.

dug-21:
[...]

PR #820, opened 2026-06-21
fix: smoke Gate 7 fires SessionStart (persisting) + bounded poll (#818)

### LINKED ISSUE #818: Release smoke Gate 7 fails: bundle-path observe did not land in per-slug store
## Summary
With #812 fixed (Gate 6 `init --bundle` now attaches), the release smoke (`product/test/infra-001/scripts/docker-http-posture-smoke.sh`) now fails at the **next** gate on both arches:

```
[783-smoke] init --bundle attached against the booted image (hermetic HOME). PASS gate 6
[783-smoke] FAIL: bundle-path observe did not land in per-slug store
##[error]smoke FAILED (exit 1): shipped image first-run path is broken.
```

Observed in run https://github.com/dug-21/unimatrix/actions/runs/27891209558 (smoke-arm64 + smoke-amd64), commit b29a3cc6.

## Context
- Gates 1–6 + the Gate-5 guard all PASS: HTTP-on boot, per-slug observe 204, per-slug store present + grew (via direct curl, Gate 4), client-bundle blob emitted, observe_url well-formed, `init --bundle` attaches.
- **Gate 7** is the end-to-end assertion: after the hermetic client is configured from the bundle, an observe sent **through the bundle-configured client path** should land in the per-slug store. The store delta is 0 → FAIL.
- Like Gate 6, this gate has **never been reached on a tag before** (Gate 6 always failed first, and before that the x64 compile broke) — no green baseline. It is freshly exercised for the first time.

## Investigation needed
1. What exactly does Gate 7 do after Gate 6 (the observe it fires through the bundle-configured client, and how it checks the per-slug store delta)?
2. Why doesn't the bundle-path observe land — real client/server bug (hook fire fail-open / async timing / wrong slug or path baked into bundle config), or a harness assertion bug (checks the wrong store, races the async write, wrong slug)?
3. Does the bundle-configured client target the same per-slug funnel that Gate 2/4 proved works via direct curl?

## Acceptance
- Release smoke Gate 7 passes on both arches (bundle-path observe lands in the per-slug store), or the gate/harness is corrected if the failure is harness-side.
- Root cause documented; verified by a green release smoke.

## Refs
- #812 (Gate 6 init --bundle — fixed; PRs #816/#817)
- #814 (real-server init --bundle delivery-gate prevention test — would also cover this round-trip)
- nan-019 (#788 smoke gate), nan-020 (init --bundle leg)

_Note: a clean release also requires a version bump — v0.8.3 npm packages are already partially published (`package-npm` fails: "cannot publish over previously published versions: 0.8.3"), so v0.8.3 is unreleaseable regardless of this fix._

dug-21:
## Summary
With #812 fixed (Gate 6 `init --bundle` attaches), the release smoke failed at the next-ever-reached gate: **Gate 7** — `bundle-path observe did not land in per-slug store` — on both arches.

## Root cause (HIGH confidence, reproduced — HARNESS bug)
Gate 7 fired a `Stop` hook through the bundle-configured client. `Stop` maps to a **SessionClose** for `s-783-bundle` — a session **never registered** (Gate 2 registered `s-783` via direct curl). The server no-ops SessionClose for an unknown session but still returns **204**, so the observe was accepted yet persisted nothing → store delta 0 → deterministic FAIL. The bundle client path itself is healthy: `SessionStart`→`SessionRegister` through the same client grows the store (verified 376→392 KB + WAL).

## Changes (harness-only, one file)
1. **Persisting probe:** `Stop`/`s-783-bundle` → `SessionStart`/`s-818-bundle`, in **both** the argv event name (dispatch keys on `argv[2]`) and `HOOK_EVENT_JSON`. SessionStart→SessionRegister is the real first-run frame — identical to what Gate 2 fires / Gate 4 asserts.
2. **Bounded deadline-poll** (~10s cap, `sleep 1`, re-sample whole slug dir incl `-wal`) replaces the immediate delta check; existing `fail()`/exit-1 on timeout (append-only per nan-019 ADR-001, no new exit code). Writes are `tokio::spawn` fire-and-forget + WAL, so sampling must retry — **not** a flat sleep.
3. Stale L261/L141 comments updated.
4. **Scope boundary named in-code:** Gate 7 proves the **lifecycle-register leg** (transport / bundle config / project-hash / cert-pin / persistence of a SessionRegister via `/observe`). It does **not** exercise the **behavioral-observe payload** (PostToolUse / transcript deltas) that self-learning (C11/C0) depends on — transcript is never-persist by design, so it produces no knowledge-store `du` delta. Behavioral-observe coverage → #814 (real-server round-trip) + the server-logged PostToolUse path (vnc-039).

No `packages/unimatrix/**`, no `crates/**`, no timeout/client change.

## Validation
- Gate-logic test 19/19; integration smoke 24/24
- Bounded poll: store grows → early-break PASS; never grows → `fail()` rc=1 at the deadline (~11s) — both demonstrated
- shellcheck **0 new** findings (3 pre-existing baseline); `bash -n` clean
- Append-only exit contract intact; harness-only blast radius
- Gate 3 Bugfix Validation: **PASS** (7/7)

## Notes / follow-ups (NOT folded in)
- **#819** (filed): server observability — an observe that Acks 204 but persists nothing is silent (the thing that hid this bug). Correct idempotency; trace/counter only.
- **#814**: real-server `init --bundle` round-trip — the durable behavioral-observe prevention.
- **#815**: pre-existing `test_no_new_smoke_script` failure (from #810) — not touched.
- **Release reality:** a clean tag also needs #798 (x64 anndists compile break) resolved and a **version bump v0.8.3 → v0.8.4** (v0.8.3 npm packages already partially published — `package-npm` fails on republish).

## GH Issue
Closes #818

### DIFF
--- a/product/test/infra-001/scripts/docker-http-posture-smoke.sh
+++ b/product/test/infra-001/scripts/docker-http-posture-smoke.sh
@@ -121,7 +121,7 @@ fire_hook() {
     # is the load-bearing assertion. We emit nothing => observe_code stays "".
     printf '%s' "$HOOK_EVENT_JSON" \
       | HOME="$SANDBOX/home" \
-        node "$REPO_ROOT/packages/unimatrix/lib/hook-client/index.js" Stop \
+        node "$REPO_ROOT/packages/unimatrix/lib/hook-client/index.js" SessionStart \
         >/dev/null 2>&1 || true
     printf ''
   fi
@@ -138,8 +138,9 @@ gate7_store_size() {
   fi
 }
 
-# A minimal hook stdin event for ONE observable record (Stop is a benign,
-# always-emitted lifecycle event). Overridable by the stub harness. cwd is
+# A minimal hook stdin event for ONE observable record (#818: SessionStart ->
+# SessionRegister, the first event a freshly-bundled client fires, proven to
+# persist on the per-slug route). Overridable by the stub harness. cwd is
 # stamped at fire time (Gate 7) once $SANDBOX/proj exists.
 HOOK_EVENT_JSON="${HOOK_EVENT_JSON:-}"
 
@@ -242,8 +243,15 @@ bundle_attach_gates() {
   BUNDLE_BEFORE="$(gate7_store_size "$SLUG_DIR")"
 
   # Stamp the event cwd into the isolated proj tree at fire time (once $SANDBOX exists).
+  # #818: probe is SessionStart (-> SessionRegister, build-request.js:50-57), NOT Stop.
+  # SessionStart is the FIRST event a freshly-bundled client fires and is proven to
+  # persist on this exact per-slug route (Gate 2 fires the identical SessionRegister
+  # frame; Gate 4 asserts its store delta). A Stop -> SessionClose for a never-
+  # registered id is a server no-op (no write) => the old probe could never grow the
+  # store. Use a FRESH, unique id (s-818-bundle, never s-783) so this delta is solely
+  # attributable to THIS fire (R-07 sc.4 fresh-write evidence).
   [ -n "$HOOK_EVENT_JSON" ] \
-    || HOOK_EVENT_JSON="{\"hook_event_name\":\"Stop\",\"session_id\":\"s-783-bundle\",\"cwd\":\"$SANDBOX/proj\"}"
+    || HOOK_EVENT_JSON="{\"hook_event_name\":\"SessionStart\",\"session_id\":\"s-818-bundle\",\"cwd\":\"$SANDBOX/proj\"}"
 
   # Fire ONE hook event through the SAME isolated HOME so the hook client reads THIS run's
   # credstore ($SANDBOX/home/.unimatrix/<hash>/remote.json), never the runner's real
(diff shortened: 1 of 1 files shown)


dug-21:
## Security Review — #818 Gate 7 (818-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings.** Verified the full diff cold against the source file, the hook client, and the in-file idioms. Approving from a security/risk standpoint.

### Verified
- **No secret/token leakage.** The new `HOOK_EVENT_JSON` carries only `hook_event_name` / `session_id` / `cwd` — no credential. The bearer lives in the per-run credstore the hook client reads internally; the new `log` line prints only two integers (`BUNDLE_BEFORE -> BUNDLE_AFTER`). `git diff | grep -i token|bearer|secret|key|credential` on added lines: zero hits.
- **Probe change is load-bearing and self-consistent.** `index.js:343` dispatches on `process.argv[2]`; the argv (`SessionStart`, L124) AND the JSON `hook_event_name` (L254) both changed. `build-request.js:50-57` maps `SessionStart -> SessionRegister` — a persisting write, the identical frame Gate 2 fires (L409) and Gate 4 asserts. The old `Stop -> SessionClose` for the never-registered `s-783-bundle` is a confirmed server no-op. Fresh id `s-818-bundle` isolates the delta. Root-cause claim confirmed at source.
- **Safe shell / no injection.** Bounded poll uses only integer `date` arithmetic and quoted expansions; no `eval`, no command-substitution on influenceable content. `$SLUG_DIR` is a literal constructed from the constant `$SLUG`.
- **Loop terminates — cannot hang the gate.** Fixed `deadline = now + 10`; `date +%s` advances monotonically past it; `sleep 1` each pass. On no growth the `fail()` branch fires within ~10-11s, well under the in-file 90s boot deadlines. No unbounded loop.
- **Fail-closed confirmed (cannot false-green).** Reproduced locally: a false `[ AFTER -gt BEFORE ] && break` does NOT abort under `set -e` (left side of `&&` is exempt) → loop proceeds to the deadline check → `fail()` exit 1, same message. Even the degenerate "du emits empty" path stays fail-closed (lands in `fail()`), not silent-pass.
- **Append-only exit contract intact (nan-019 ADR-001).** Reuses the existing `fail()` exit 1 with the same message; no new exit code; Gates 1-4 / stub seam / terminal marker untouched.
- **Hermeticity unchanged.** Sandbox + cleanup trap untouched; the poll adds no artifacts.
- **shellcheck: 0 new.** 3 findings (L210/L303/L369) are all pre-existing, none in the changed region. `bash -n` clean.

### Blast radius / regression
Harness-only — one CI smoke script. Cannot reach the shipped client or the Rust workspace. Worst case if subtly wrong is a flaky/failing **release gate** (loud, fail-closed), never a shipped defect or silent pass.

### Non-blocking nit (pre-existing, not introduced here)
When `du` yields a non-integer (e.g. slug dir vanished), `[ -gt ]` prints a `integer expression expected` stderr line each iteration before fail-closing. Same pattern exists in the original check and Gate 4 (L429-432), so not regressed by this PR. Optional future hardening: default the sampler to `0` on empty output. No action required for merge.

PR #822, opened 2026-06-21
fix(release): make npm publish steps idempotent on re-run (#821)

### LINKED ISSUE #821: Release `package-npm` not idempotent — re-run on already-published version fails with "cannot publish over previously published versions"
## Symptom

The v0.8.3 Release workflow fails at `package-npm` → **Publish `@dug-21/unimatrix-linux-arm64`** (run [27902516992](../../actions/runs/27902516992)):

```
npm error You cannot publish over the previously published versions: 0.8.3.
```

`create-release` (`needs: package-npm`) is skipped as a result, so no GitHub Release is cut.

## Root cause

All three npm packages are **already published at 0.8.3** — the publish succeeded in an earlier run of the same tag. The smoke-gate fixes (#812, #818) each re-ran the **entire** Release workflow on the unchanged `v0.8.3` tag. The publish steps run `npm publish` unconditionally:

```yaml
# .github/workflows/release.yml:266-276
- name: Publish @dug-21/unimatrix-linux-arm64
  run: npm publish --access public --provenance
- name: Publish @dug-21/unimatrix-linux-x64
  run: npm publish --access public --provenance
- name: Publish @dug-21/unimatrix
  run: npm publish --access public --provenance
```

npm forbids republishing an existing version, so any re-run after the first successful publish fails. The job is **not idempotent across re-runs** — a recurring trap because our release recovery model is "fix the smoke gate, re-run the tag."

## Fix

Make each publish step idempotent: query `npm view <pkg>@<version> version` and skip the publish if it already exists; only publish if absent. Apply to all three packages. This lets a re-run drive a partially-published release to completion and lets `create-release` finally succeed.

## Acceptance

- Re-running the Release workflow on a tag whose npm packages are already published succeeds (publishes skipped, not errored).
- A partially-published release (e.g. arm64 published, x64 not) completes on re-run.
- `create-release` runs after `package-npm` on such a re-run.

## Notes / blast radius

- `release.yml` only; no Rust/client code.
- v0.8.3 npm packages and container images are already published; this only unblocks the re-run and the GitHub Release.

dug-21:
Fixes #821.

## Problem
The last v0.8.3 Release run failed at `package-npm` → Publish `@dug-21/unimatrix-linux-arm64`:
```
npm error You cannot publish over the previously published versions: 0.8.3.
```
All three `@dug-21` packages were already published at 0.8.3 (they succeeded in an earlier run). The smoke-gate fixes (#812, #818) re-ran the whole Release workflow on the unchanged `v0.8.3` tag, but the publish steps ran `npm publish` unconditionally. npm versions are immutable, so every re-run after the first success died at the first publish — and `create-release` (`needs: package-npm`) was skipped, so no GitHub Release was cut.

## Fix
Guard each of the three publish steps: derive name+version from the package's own `package.json`, query the registry, and publish (`--access public --provenance`) only when the version is absent.

**Load-bearing detail:** gate on **non-empty `npm view` stdout**, never exit code. `npm view pkg@ver version` exits 0 with empty stdout when the version is missing but the package name exists, and only E404s when the name was never published — gating on `$?` would silently skip a new package's first publish.

`create-release` verified re-run-safe (`softprops/action-gh-release@v2` updates an existing release); added invariant comments documenting the workflow-wide rule.

## Scope & verification
- `.github/workflows/release.yml` only (+43/-3). No Rust/JS/source changes.
- YAML parses; guard logic verified by a 3-branch shell simulation under `set -euo pipefail` (skip-on-exists / publish-on-missing-version / publish-on-E404-without-abort) plus a negative control confirming `set -e` is active.
- Bugfix Gate: PASS (7/7).

## Effect on v0.8.3
v0.8.3's npm packages are already fully published; once merged, re-running the v0.8.3 tag will skip the publishes and finally cut the GitHub Release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -263,17 +263,53 @@ jobs:
             fi
           done
 
+      # INVARIANT: every publish/push step in this workflow MUST be safe to
+      # re-run on an already-completed version. Re-running the same tag is the
+      # standard release-recovery path (see #812, #818) — a non-idempotent step
+      # breaks recovery and cascades (e.g. skips create-release via `needs:`).
+      # npm versions are immutable, so each publish below is guarded: gate on
+      # NON-EMPTY `npm view` stdout, never exit code — `npm view pkg@ver version`
+      # exits 0 with empty stdout when the version is missing but the package
+      # name exists, and only exits non-zero (E404) when the name was never
+      # published; gating on $? would silently skip a new package's first publish.
       - name: Publish @dug-21/unimatrix-linux-arm64
         working-directory: packages/unimatrix-linux-arm64
-        run: npm publish --access public --provenance
+        run: |
+          set -euo pipefail
+          PKG=$(node -p "require('./package.json').name")
+          VER=$(node -p "require('./package.json').version")
+          if [ -n "$(npm view "${PKG}@${VER}" version 2>/dev/null)" ]; then
+            echo "skip: ${PKG}@${VER} already published"
+          else
+            echo "publish: ${PKG}@${VER}"
+            npm publish --access public --provenance
+          fi
 
       - name: Publish @dug-21/unimatrix-linux-x64
         working-directory: packages/unimatrix-linux-x64
-        run: npm publish --access public --provenance
+        run: |
+          set -euo pipefail
+          PKG=$(node -p "require('./package.json').name")
+          VER=$(node -p "require('./package.json').version")
+          if [ -n "$(npm view "${PKG}@${VER}" version 2>/dev/null)" ]; then
+            echo "skip: ${PKG}@${VER} already published"
+          else
+            echo "publish: ${PKG}@${VER}"
+            npm publish --access public --provenance
+          fi
 
       - name: Publish @dug-21/unimatrix
         working-directory: packages/unimatrix
-        run: npm publish --access public --provenance
+        run: |
+          set -euo pipefail
+          PKG=$(node -p "require('./package.json').name")
+          VER=$(node -p "require('./package.json').version")
+          if [ -n "$(npm view "${PKG}@${VER}" version 2>/dev/null)" ]; then
+            echo "skip: ${PKG}@${VER} already published"
+          else
+            echo "publish: ${PKG}@${VER}"
+            npm publish --access public --provenance
+          fi
 
   create-release:
     needs: package-npm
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — PR #822 (uni-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings. Approve to merge.**

Reviewed `git diff main...bugfix/821-npm-publish-idempotent` cold (release.yml only, +43/-3) plus the full job context and #821.

### Command/shell injection — no surface
`PKG`/`VER` come from `node -p require('./package.json').{name,version}` and are used only as `npm view "${PKG}@${VER}" version`. Three reasons this is safe:
- package.json files are repo-controlled, not external input.
- The earlier **Validate versions** step already string-compares every package's version to Cargo.toml and the tag — a malformed/unexpected version fails the job before any publish step runs.
- Vars are **double-quoted** and passed via command substitution, which is not re-evaluated by the shell. A hostile value like `foo; rm -rf /` would reach `npm view` as one literal argument, never executed.

`npm publish` reads package.json directly and takes no interpolated var.

### Guard correctness (load-bearing) — correct
Gating on **non-empty stdout** rather than exit code is right: `npm view pkg@ver version` exits 0 with empty stdout when the version is absent-but-name-exists, and only E404s when the name was never published. Gating on `$?` would silently skip a brand-new package's first publish. `2>/dev/null` + the `if [ -n "$(...)" ]` test context means `set -euo pipefail` does not abort on the E404 path (command substitution in a test condition is exempt from `set -e`). Verified.

### Supply-chain / provenance — not weakened
- A skip can only occur for the exact `name@version` already on the registry, and npm versions are immutable — so a skip never substitutes a wrong artifact; the correct bytes are already published.
- `--provenance` and `id-token: write` preserved unchanged on the publish path. OIDC/provenance intact.
- Note (non-blocking, inherent to npm): the guard masks any local artifact drift vs. an already-published version. This is unavoidable under npm immutability (you can't republish regardless) and is the intended recovery semantics, not a new risk.

### Secrets — clean
No hardcoded credentials/tokens. Auth via OIDC + setup-node (`node-auth-token: ''` unchanged). Diff adds zero secrets.

### Blast radius if the guard misfires — safe failure modes
- False "exists" → a publish is skipped; resolved by a later fixed re-run. npm immutability makes overwriting/corrupting an artifact impossible.
- False "missing" → `npm publish` re-errors exactly as today ("cannot publish over previously published versions"). Degrades to current behavior, not worse.
- No path to silent data corruption, privilege change, or provenance bypass.

### Regression risk — low
`set -euo pipefail` is stricter than the prior unconditional one-liners (improvement). `create-release` / `needs:` chain untouched; the `softprops/action-gh-release@v2` re-run-safety note is accurate (it updates an existing release). Remaining changes are documentation comments.

Self-check: full diff read; #821 root cause verified against code; full release.yml read in context; OWASP injection/secrets/supply-chain evaluated; worst case named.

Issue #824, opened 2026-06-21
follow-up: per-slug vector index heal-on-boot + non-atomic dump load-fallback (N2 from #823)


dug-21:
Tracked follow-up from the #823 design review (architect N2 + product review).

## Context
#823 fixes the regression where per-slug HTTP vector indices were never persisted (dump wiring). That fix restores the dump→load round-trip but does **not** add resilience to a torn or deleted per-slug `{slug}/vector/` dir.

## Problem (N2)
`VectorIndex::dump` is non-atomic (writes graph/data then meta, no temp+rename). A crash mid-dump leaves a `.meta` pointing at a torn graph. On next boot `VectorIndex::load` errors and `build_project_server` propagates it as a **hard boot failure for that slug** — tensions with Architectural Principle 5 (graceful degradation). Same property is shared by the daemon dump.

## Proposed scope
1. On `load` failure, fall back to `VectorIndex::new` (degrade + heal) rather than abort boot.
2. Boot-time heal-from-store: rebuild the per-slug index from the store when the vector dir is missing/torn.
3. Consider atomic dump (temp + rename) for both per-slug and daemon dumps.

## Out of scope of #823 deliberately
Per human approval on #823: land minimal dump wiring + round-trip test now; this resilience work is a strictly larger, separate capability.

dug-21:
# Bug Investigation Report: 824-investigator

## Bug Summary
#823 wired the per-slug HTTP `VectorIndex` dump→load round-trip but added no resilience to a torn or deleted per-slug `{slug}/vector/` dir. A crash mid-`dump` (non-atomic) leaves a `.meta` whose `point_count > 0` while the `.hnsw.graph`/`.data` are missing or torn. On next boot `VectorIndex::load` returns `Err(Persistence)` and `build_project_server` propagates it as `ServerError`, aborting boot for that slug — a hard failure that tensions with Architectural Principle 5 (graceful degradation).

## Root Cause Analysis

### Code Path Trace
- `infra/shutdown.rs::graceful_shutdown` (Step 1 / Step 1-slug) → `VectorIndex::dump(dir)` — NON-ATOMIC
- `persistence.rs::dump` writes `{basename}.hnsw.graph` + `.hnsw.data` via `hnsw.file_dump(...)`, THEN `std::fs::write(meta)` last. No temp + rename. A crash between the graph write and the meta write — or a partial graph write — leaves an inconsistent on-disk set: meta claiming N points over a torn/missing graph.
[...]

PR #825, opened 2026-06-22
fix: persist per-slug HTTP vector indices on shutdown (#823)

### LINKED ISSUE #823: fix: per-slug HTTP vector indices never persisted to disk (multi-project mode)
## Summary
In multi-project HTTP mode, per-slug vector (HNSW) indices are **never persisted to disk**. Confirmed in a live deployment: after writes to `/v1/{slug}/...`, `{slug}/vector/` stays empty; only the daemon's hash-rooted index is ever dumped.

## Symptom
- Knowledge writes store entries correctly in `{slug}/unimatrix.db`.
- The per-slug HNSW index serves search in-session (correct per-slug data, single correct directory).
- But `{slug}/vector/` is never written — the index is in-memory only and lost on restart, with no per-slug load/rebuild path, so semantic search silently degrades after a reboot.

## Root cause (file:line)
- `crates/unimatrix-server/src/http_provision.rs:165,185-199` — `build_project_server` builds the per-slug `VectorIndex` over `vector_dir = {base}/{slug}/vector`, but `vector_dir` is a local var that is dropped.
- `crates/unimatrix-server/src/http/router/project_resolver.rs:126` — `ProjectServerInput` carries only `{slug, store, server}`: no `vector_dir`, no index handle.
- `crates/unimatrix-server/src/infra/shutdown.rs:173` — the only production dump site, `handles.vector_index.dump(&handles.vector_dir)`, is wired (`main.rs:1302-1304`) to the **daemon's hash-rooted** index/dir only. Per-slug indices are never registered for dump. No `.dump(` exists in `background/`.

## Impact
- Regresses capability C3 (per-slug isolated stores) for the vector backend.
- Semantic search degrades after any restart in multi-project deployments.

## Proposed fix
1. Add `vector_dir: PathBuf` (and expose the per-slug `Arc<VectorIndex>`) on `ProjectServerInput`; populate from `build_project_server`.
2. Carry `Vec<(Arc<VectorIndex>, PathBuf)>` of per-slug (index, dir) into `LifecycleHandles`; in `graceful_shutdown`, dump each to its own `{slug}/vector/` (warn-and-continue per entry, mirroring the daemon dump).
3. On startup, load/rebuild each per-slug index from `{slug}/vector/` (or heal-pass from the store) so search survives restart.

## Missing test
Multi-project HTTP integration test: register a slug, write via `/v1/{slug}/...`, graceful shutdown, then assert `{slug}/vector/unimatrix-vector.meta` exists AND the path-hash `{hash}/vector/` did not gain the slug's vectors. (N>=1 persistence analog of the #4974 funnel risk.)

## Repro
HTTP-only deployment, one registered slug. Write knowledge over HTTP -> `{slug}/vector/` stays empty. Restart -> semantic search degraded/empty.

_Root-caused during a deployment debug session; scope to be confirmed by a bugfix investigation before implementation._

dug-21:
## Summary
Fixes #823 — in multi-project HTTP mode, per-slug vector (HNSW) indices were never persisted to disk. The per-slug index was an in-memory-only dropped local; only the daemon's hash-rooted index was ever dumped. After a restart, per-slug semantic search silently degraded.

## Root cause
`build_project_server` built each per-slug `VectorIndex` over `vector_dir = {base}/{slug}/vector`, but `vector_dir` was a dropped local and `ProjectServerInput` carried only `{slug, store, server}`. The sole production dump (`shutdown.rs`) was wired to the daemon's hash-rooted index only — per-slug indices were never registered for the shutdown dump. The startup `VectorIndex::load` arm already existed but was dead (no `.meta` was ever written).

## Fix (minimal)
- Add `vector_dir: PathBuf` to `ProjectServerInput`; stop dropping it in `build_project_server`.
- Carry `per_slug_vectors: Vec<(Arc<VectorIndex>, PathBuf)>` into `LifecycleHandles` (empty in stdio / single-project).
- In `graceful_shutdown`, dump each per-slug index to its own `{slug}/vector/` in Step 1 (before the Arc drops), **warn-and-continue per entry** (N1) so one bad slug dir cannot abort the others or block the daemon `try_unwrap`.
- Startup `load` arm needs no change once `.meta` is written.

## Out of scope (tracked as #824)
Heal-on-boot from store, `load`-failure fallback to a fresh index, and atomic (temp+rename) dump (N2 from the design review). The dump→load round-trip restores single-project parity, which is the regression being closed.

## Testing
- New regression test `test_graceful_shutdown_dumps_per_slug_vector_index`: asserts the slug's `{slug}/vector` meta has non-zero points, the daemon `{hash}/vector` did NOT gain the slug's vectors, and a load+search round-trip returns the seeded entry. Fails on `main`.
- Per-crate suites (workspace `cargo test` OOMs the linker in this env — #5023): unimatrix-server 4532 passed, all workspace crates green.
- Clippy `-p unimatrix-server --all-targets -D warnings`: clean.
- Integration smoke: 24 passed; restart/lifecycle: 7 passed.

Gate: Bug Fix Validation — **PASS** (9/9, 1 non-blocking WARN).

### DIFF
--- a/crates/unimatrix-server/src/http/router/project_resolver.rs
+++ b/crates/unimatrix-server/src/http/router/project_resolver.rs
@@ -130,6 +130,14 @@ pub struct ProjectServerInput {
     pub store: Arc<Store>,
     /// The assembled per-slug MCP server.
     pub server: UnimatrixServer,
+    /// The slug's own vector dump directory (`{base_dir}/{slug}/vector`).
+    ///
+    /// Carried so the listener wiring can register this slug's `VectorIndex`
+    /// (reachable via `server.vector_index()`) for the per-slug shutdown dump
+    /// (#823). Without this, the per-slug HNSW index was an in-memory-only
+    /// dropped local that was never persisted, silently degrading semantic
+    /// search after a restart in multi-project HTTP mode.
+    pub vector_dir: std::path::PathBuf,
 }
 
 impl std::fmt::Debug for ProjectServerInput {
--- a/crates/unimatrix-server/src/http/router/project_resolver/tests.rs
+++ b/crates/unimatrix-server/src/http/router/project_resolver/tests.rs
@@ -35,10 +35,13 @@ async fn make_slug_input(slug: &str) -> (ProjectServerInput, Arc<Store>) {
     // resolved store IS this one.
     let store = Arc::clone(&server.store);
     let slug = ProjectSlug::try_from(slug).expect("valid slug");
+    // #823: inert in resolver tests (no shutdown drive); a faithful per-slug dir.
+    let vector_dir = std::path::PathBuf::from(slug.as_str()).join("vector");
     let input = ProjectServerInput {
         slug,
         store: Arc::clone(&store),
         server,
+        vector_dir,
     };
     (input, store)
 }
--- a/crates/unimatrix-server/src/http/router/tests.rs
+++ b/crates/unimatrix-server/src/http/router/tests.rs
@@ -2676,11 +2676,15 @@ async fn test_observe_per_slug_funnel_isolation_n2() {
         slug: ProjectSlug::try_from("alpha").expect("valid"),
         store: Arc::clone(&alpha.store),
         server: alpha,
+        // #823: inert here (router tests never drive shutdown); carries the
+        // per-slug dump dir only for the listener wiring path.
+        vector_dir: std::path::PathBuf::from("alpha/vector"),
     };
     let beta_input = ProjectServerInput {
         slug: ProjectSlug::try_from("beta").expect("valid"),
         store: Arc::clone(&beta.store),
         server: beta,
+        vector_dir: std::path::PathBuf::from("beta/vector"),
     };
     let inner = MultiProjectRouter::from_servers(
         vec![alpha_input, beta_input],
@@ -2737,6 +2741,8 @@ async fn test_observe_unregistered_slug_is_loud_404_not_default() {
         slug: ProjectSlug::try_from("only").expect("valid"),
         store: Arc::clone(&only.store),
         server: only,
+        // #823: inert here (router tests never drive shutdown).
+        vector_dir: std::path::PathBuf::from("only/vector"),
     };
     let inner = MultiProjectRouter::from_servers(
         vec![only_input],
--- a/crates/unimatrix-server/src/http_provision.rs
+++ b/crates/unimatrix-server/src/http_provision.rs
@@ -31,14 +31,14 @@ use unimatrix_engine::confidence::ConfidenceParams;
 use unimatrix_observe::domain::DomainPackRegistry;
 use unimatrix_server::error::ServerError;
 use unimatrix_server::http::{
-    build_tls_acceptor, derive_public_url, load_or_generate_cert, Env, ProjectServerInput,
-    ProjectSlug, PublicUrl,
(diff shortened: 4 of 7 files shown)


dug-21 (commented):
## Security Review — PR #825 (fresh context)

**Risk: LOW. No blocking findings. Approve from a security standpoint.**

Reviewed `git diff main...HEAD` cold against OWASP concerns, blast radius, regression risk, dependencies, and secrets.

### Path traversal (the main concern) — SAFE
The only attacker-influenced path component is `slug`, flowing `base_dir.join(slug.as_str())` -> `vector_dir` (`http_provision.rs:161-165`) -> the shutdown `dump(dir)`. `ProjectSlug::try_from` (`http/router/seam.rs:83-116`) enforces an allowlist `^[a-z0-9][a-z0-9-]{0,62}$` at the parse edge, *before* any path join. `.`, `/`, `\`, `%`, whitespace, and uppercase are forbidden by construction, so `..`, absolute paths, and percent-encoded separators (`%2e`/`%2f`) cannot reach `join`. `as_str()` returns the validated raw string unchanged. Escape from `{base_dir}/{slug}/vector` is structurally impossible, not runtime-rejected. No new trust boundary is introduced — the new `vector_dir` is derived from the same already-validated slug.

### Injection / deserialization / secrets — none
- `dump(dir)` takes a `&Path` and writes via `std::fs` + `hnsw_rs::file_dump` — no shell, no format-string-into-command, no SQL. No injection surface.
- No deserialization of untrusted input added.
- No hardcoded credentials/tokens/keys in the diff.

### Dependencies — none added
No `Cargo.toml` changes in the diff. Zero new dependencies; `cargo audit` posture unchanged.

### Blast radius / regression — bounded
- **Arc / try_unwrap:** per-slug `VectorIndex` Arcs hold *per-slug* stores, not the daemon `store` that Step 3 `try_unwrap` compacts. They are dumped in Step 1, then `per_slug_vectors.clear()` runs in Step 2 before the daemon `try_unwrap` — so they add no new outstanding-reference hazard to compaction. Verified against `shutdown.rs:188-241`.
- **Shutdown blocking:** the per-slug loop is warn-and-continue *per entry* (`shutdown.rs:195-206`), mirroring the existing daemon dump. One bad slug dir cannot abort the others or block compaction. `dump` is synchronous local FS I/O with no unbounded wait.
- **Wrong-dir risk:** each pair carries its own `{slug}/vector` dir; the regression test asserts the daemon hash-rooted dir does NOT gain the slug's vectors. Good.
- **Single-project / stdio:** `per_slug_vectors` is `Vec::new()` on both those paths (`main.rs` stdio arm + daemon non-multi-project); the existing daemon dump is byte-for-byte unchanged. No regression to those modes.
- `per_slug_vectors.push(...)` is inside the multi-project HTTP branch only.

### Notes (non-blocking)
- Deferred items (heal-on-boot, load-failure fallback to fresh index, atomic temp+rename dump) are tracked in #824. A non-atomic dump interrupted mid-write could leave a partial `{slug}/vector` that a later `load` rejects; with #824's fallback absent today, a load failure would surface as a startup error rather than a silent rebuild. Acceptable for closing this regression, but worth keeping #824 prioritized.

Self-checks: full diff read; root cause (issue #823) read; surrounding source (`shutdown.rs`, `http_provision.rs`, `persistence.rs`, `seam.rs`, `main.rs`) read in full; OWASP evaluated; worst case named; slug validation independently verified.

PR #826, opened 2026-06-22
fix: atomic vector dump + boot load-fallback for torn per-slug index (#824)

### LINKED ISSUE #824: follow-up: per-slug vector index heal-on-boot + non-atomic dump load-fallback (N2 from #823)

dug-21:
Fixes #824 (N2 follow-up from #823).

## Root cause
`VectorIndex::dump` was non-atomic (wrote `hnsw.graph` + `hnsw.data`, then `.meta` last, no temp+rename). A crash mid-dump left a `.meta` claiming `point_count > 0` over a torn/missing graph. On next boot `VectorIndex::load` returned `Err`, and the boot path `?`-propagated it as a **hard boot abort for that slug** — violating Architectural Principle 5 (graceful degradation).

## Fix (Items 1 + 3; Item 2 deferred)
- **Item 1 — boot load-fallback (3 sites):** at `http_provision.rs:194`, `main.rs:742`, `main.rs:1472`, replace `?` with a `match` that `warn!`s and falls back to `VectorIndex::new` on load `Err`. The slug boots empty and self-heals via the existing capped maintenance heal pass.
- **Item 3 — atomic dump:** write the 3-file set to sibling temp files inside the target dir, then `rename` into place publishing `.meta` **last**; empty-dump path removes stale graph/data files. Covers both daemon and per-slug dumps. Power-loss/fsync durability explicitly out of scope (temp+rename covers the SIGKILL repro).
- **Eval load site (`layer.rs:232`) intentionally left `?`-propagating** — a silent empty index over a corrupt eval snapshot is the bug #2673 was created to prevent.
- **Item 2 (eager boot-time heal-from-store) deferred** as a separate `goal:personal-cloud` feature (re-embed budget, model dependency, boot latency); the existing capped heal pass already delivers incremental self-heal. To be filed as a follow-up.

## Tests (5 new)
- `test_dump_atomic_no_temp_artifacts`, `test_empty_dump_removes_stale_graph_data`, `test_interrupted_dump_keeps_prior_consistent_set` (unimatrix-vector)
- `test_build_project_server_torn_dump_boots_empty_not_err` (server bin) — drives real `build_project_server` over a torn dump, asserts `Ok` + empty
- `test_per_slug_index_heals_after_degrade_to_empty` (server lib) — per-slug, multi-project heal-after-degrade proven directly (not by daemon analogy)

## Verification
5031 unit tests pass, 0 failed. Smoke 24/24. Clippy clean. Gate: Bug Fix Validation — PASS (7/7).

### DIFF
--- a/crates/unimatrix-server/src/http_provision.rs
+++ b/crates/unimatrix-server/src/http_provision.rs
@@ -186,11 +186,27 @@ pub async fn build_project_server(
     let vector_config = VectorConfig::default();
     let meta_path = vector_dir.join("unimatrix-vector.meta");
     let vector_index = if meta_path.exists() {
-        Arc::new(
-            VectorIndex::load(Arc::clone(&store), vector_config, &vector_dir)
-                .await
-                .map_err(|e| ServerError::Core(CoreError::Vector(e)))?,
-        )
+        // Graceful degradation (Architectural Principle 5): a torn/missing dump
+        // (e.g. SIGKILL mid-dump leaving a meta over a missing graph) must NOT
+        // hard-abort this slug's boot. Fall back to an empty index; the capped
+        // run_maintenance heal pass (services/status.rs) re-populates it from
+        // the store on the next tick. See GH-824 / lesson #5272.
+        match VectorIndex::load(Arc::clone(&store), vector_config.clone(), &vector_dir).await {
+            Ok(idx) => Arc::new(idx),
+            Err(e) => {
+                tracing::warn!(
+                    slug = %slug,
+                    vector_dir = %vector_dir.display(),
+                    error = %e,
+                    "failed to load per-slug vector index; booting empty and \
+                     deferring to maintenance heal pass"
+                );
+                Arc::new(
+                    VectorIndex::new(Arc::clone(&store), vector_config)
+                        .map_err(|e| ServerError::Core(CoreError::Vector(e)))?,
+                )
+            }
+        }
     } else {
         Arc::new(
             VectorIndex::new(Arc::clone(&store), vector_config)
@@ -448,3 +464,6 @@ fn flatten_present_keys(raw: &toml::Value) -> Vec<String> {
 
 #[cfg(test)]
 mod slug_config_tests;
+
+#[cfg(test)]
+mod boot_fallback_tests;
--- a/crates/unimatrix-server/src/http_provision/boot_fallback_tests.rs
+++ b/crates/unimatrix-server/src/http_provision/boot_fallback_tests.rs
@@ -1,0 +1,140 @@
+//! GH-824 Item 1 boot-resilience test: a torn per-slug vector dump must NOT
+//! hard-abort `build_project_server`; the slug boots with an EMPTY index and
+//! self-heals via the existing maintenance heal pass.
+//!
+//! Drives the REAL `build_project_server` boot path (the F1 acceptance site
+//! `http_provision.rs`) over a slug whose `{slug}/vector/` dir carries a
+//! `point_count > 0` meta but a DELETED `.hnsw.graph` — the exact on-disk state
+//! a SIGKILL mid-dump leaves. Asserts `Ok` with an empty index, not `Err`.
+
+use std::collections::HashSet;
+use std::sync::Arc;
+
+use unimatrix_core::{VectorConfig, VectorIndex};
+use unimatrix_engine::confidence::ConfidenceParams;
+use unimatrix_observe::domain::DomainPackRegistry;
+use unimatrix_server::http::ProjectSlug;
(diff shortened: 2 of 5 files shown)


dug-21 (commented):
## Security Review — PR #826 (fresh context)

**Risk level: LOW.** No blocking findings. Atomic-dump + boot-fallback is sound, minimal, and adds no dependencies. OWASP surface is clean: no new external inputs, no injection vector, no secrets. Path handling is safe — temp/final filenames derive from `const DUMP_BASENAME`/`METADATA_FILENAME`, not from any untrusted value; the slug join is already allowlist-validated upstream (AC-W2-R6). Eval load site (`layer.rs:232`) correctly still `?`-propagates. EXDEV is correctly avoided by staging temps as siblings inside the target dir.

Findings below are advisory — none block merge.

### F1 (low/advisory) — graph/data publish is two non-atomic renames over stable filenames
`persistence.rs:107-118`. The published basename is the stable `unimatrix`, so the graph and data renames **overwrite the prior dump in place**. A crash *between* `rename(temp_graph→final_graph)` and `rename(temp_data→final_data)` leaves: NEW `unimatrix.hnsw.graph` + OLD `unimatrix.hnsw.data` + OLD meta. Both files exist, so `load` passes its existence checks (`:175-186`) and pairs a **new graph against an old data file** — a mismatched HNSW set the doc comment's invariant ("observer sees either the prior dump or the new complete one") does not actually cover.

- **Blast radius:** bounded. If `hnsw_rs::load_hnsw` rejects the mismatch → `Err` → the new Item-1 fallback degrades to empty + maintenance heal (safe). If it loads silently → a subtly inconsistent index until the next clean dump/heal (no data loss in the store; vectors re-derive from the source of truth). Window is two `rename(2)` syscalls — very small.
- **Why non-blocking:** the SIGKILL-mid-dump repro this PR targets (meta over missing graph) IS fully fixed; this residual is a strictly narrower window than what existed before, and the store remains the source of truth so it self-heals. 
- **Recommendation (follow-up, not this PR):** stage BOTH graph+data under a versioned/temp basename and have the meta point at that basename, flipping the whole set via the single meta rename (meta already carries `basename=`, so load would pick the matching pair atomically). Worth a tracked issue.

### F2 (info) — temp files are predictable, fixed names
`persistence.rs:45-46,55`. Temps are `.unimatrix.hnsw.graph.tmp` etc. — fixed, predictable names inside the (per-slug, 0700-intent) data dir. Not a TOCTOU/symlink concern in the trusted single-tenant data dir, and `rename` over an attacker-planted temp would at worst fail loud. No action; noting for completeness vs. the `mktemp` convention used elsewhere. Concurrent dumps into the same dir would collide on these fixed temp names, but dump is read-locked and per-index, so this isn't reachable today.

### Regression / round-trip
- #823 dump→load round-trip preserved: final filenames (`unimatrix.hnsw.graph/data`, `unimatrix-vector.meta`) and the `basename=unimatrix` meta are unchanged; snapshot/eval consumers that read these names are unaffected.
- Empty-dump now removes stale graph/data (`:119-125`) — strictly improves consistency vs. the old empty-placeholder behavior; covered by `test_empty_dump_removes_stale_graph_data`.
- Fallback does NOT mask genuine load failure inappropriately: it degrades to empty + logs `warn!` with slug+dir+error, and the store-backed heal pass re-populates. The one site where a silent-empty WOULD be a real bug (eval snapshot, #2673) is deliberately left propagating. Correct call.

### Dependencies / secrets
No `Cargo.toml` changes — zero new dependencies. No hardcoded secrets. No `unsafe`.

Recommend merge. Suggest filing F1 as a tracked follow-up (`goal:personal-cloud`) alongside the already-deferred Item 2.

PR #831, opened 2026-06-23
fix: self-heal stale cloud MCP session (keep_alive + bridge re-init) (#830)

### LINKED ISSUE #830: MCP client connection to unimatrix goes stale silently and is not auto-retained — requires manual /mcp reconnect
### Summary

When using the `unimatrix` MCP server from Claude Code over a longer session, the connection appears to go **stale**: tool calls stop succeeding and only recover after a manual `/mcp` reconnect (which reports `Reconnected to unimatrix`). The client does not appear to auto-detect or auto-recover the dropped connection.

### Observed behavior

- Early in a session, `context_*` calls (e.g. `context_status`, `context_search`, `context_correct`) succeed normally.
- After a period / some interaction, subsequent `context_*` calls stop going through — they surface at the Claude Code harness layer as tool-use rejections/interruptions rather than returning a server response.
- The failure persists across retries and across different `agent_id` values.
- Running `/mcp` and reconnecting deterministically restores normal operation; calls succeed again immediately after.

### What recovered it

A manual `/mcp` → reconnect. No server restart was needed, suggesting the server stayed up and the **client-side connection** was the stale party.

### Evidence limitations (so this isn't overstated)

- I do **not** have a captured server-side error string or MCP transport log for the stale window — the failures surfaced as Claude Code harness rejection messages, which are ambiguous (they don't distinguish "transport closed" from other rejections). The strongest signal is that a `/mcp` reconnect reliably fixes it.

### Suggested directions

- Add a keepalive/heartbeat or idle-timeout handling so a dropped connection is detected.
- Auto-reconnect (or surface a clear, actionable transport error) instead of failing opaquely.

*Filed via Claude Code during a working session; happy to attach MCP debug logs if pointed at where to capture them.*

dug-21:
## Summary
Fixes #830 — MCP connection from a remote client to a **cloud-deployed** unimatrix goes stale after ~5 min idle and requires a manual \`/mcp\` reconnect. Never reproduced on local STDIO.

## Root cause
The cloud transport serves MCP over rmcp Streamable-HTTP with \`LocalSessionManager\`, whose \`session_config.keep_alive\` defaults to **300s**. The server built it with \`::default()\` and never overrode it, so after ~5 min idle the session is evicted → the next call returns HTTP **404 "Session not found"**. The bridge captured the session id once on \`initialize\` and replayed it for the process lifetime with no re-init path, and \`dispatch.js\` collapsed the 404 into an opaque \`-32603\` — so every later call repeated the failure and the agent parked. The daemon stays up, which is why \`/mcp\` re-init recovers it without a restart.

## Fix (two complementary layers)
- **Layer A — server** (\`crates/unimatrix-server/src/http/router.rs\`): override \`LocalSessionManager.session_config.keep_alive\` to a **4h finite bound**. Removes the routine idle trigger. Finite (not \`None\`) to avoid zombie-session map growth.
- **Layer B — bridge** (\`packages/unimatrix/lib/hook-client/mcp-bridge/{dispatch.js,lifecycle.js}\`): on a 404 session-not-found, drop the cached session id, **re-initialize once** (preserving \`clientInfo.name\` → vnc-039 audit-attribution contract), and **retry the failed call exactly once**. Single-flight guard prevents re-init storms on concurrent 404s; re-init failure aborts (no loops); a 404 on \`initialize\` itself is fatal. Durable self-heal across evictions, redeploys, and infra idle cutoffs.

## Tests
- Rust: 2 new router tests asserting the keep_alive override lands on the correct field at a long finite bound.
- JS: 6 new bridge tests — 404 classification, re-init-once + retry success, no-loop on second failure, abort on re-init failure, single-flight on concurrent 404s, fatal-on-init-404.
- Full suite green: Rust workspace 6689/0, clippy clean, JS 741/0, integration smoke 24/24.

## Reviews
Diagnosis (v2), architecture design review (APPROVED WITH NOTES), and product review all on the issue. v1 diagnosis targeted the local STDIO bridge and was correctly rejected — the remote bug never traverses that path.

### DIFF
--- a/crates/unimatrix-server/src/http/router.rs
+++ b/crates/unimatrix-server/src/http/router.rs
@@ -17,6 +17,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
+use std::time::Duration;
 
 use bytes::Bytes;
 use http::{Method, Request, Response};
@@ -38,6 +39,23 @@ use crate::services::ServiceLayer;
 
 /// Maximum request body size default (1 MB). Used when no config override.
 const DEFAULT_MAX_BODY_BYTES: usize = 1_048_576;
+
+/// Idle keep-alive (TTL) for rmcp Streamable-HTTP MCP sessions (#830).
+///
+/// rmcp's `SessionConfig::keep_alive` defaults to `DEFAULT_KEEP_ALIVE = 300s`,
+/// which evicts an idle session after 5 minutes. A normal agent think-gap >5min
+/// then loses its server-side session → the next call with the stale
+/// `Mcp-Session-Id` gets HTTP 404 "Session not found" and the bridge hangs until
+/// a manual `/mcp` re-init (#830 cloud staleness).
+///
+/// We override it to a long FINITE bound (4 hours), set on
+/// `LocalSessionManager.session_config.keep_alive` (NOT
+/// `StreamableHttpServerConfig.sse_keep_alive` — a different SSE-ping field that
+/// would NOT fix this). 4h ≫ any realistic idle gap, while keeping rmcp's
+/// finite-TTL safety net so sessions from clients that drop without a clean
+/// DELETE still self-evict (a `None`/unbounded TTL would leak them forever,
+/// growing the session map without bound — #830 design review B2).
+const SESSION_KEEP_ALIVE: Duration = Duration::from_secs(4 * 60 * 60);
 
 // ---------------------------------------------------------------------------
 // ObserveContext — service handle bundle for /observe handler (ADR-001)
@@ -337,7 +355,14 @@ impl McpAdapter {
         allowed_origins: Vec<String>,
         allowed_hosts: Vec<String>,
     ) -> Self {
-        let session_manager = Arc::new(LocalSessionManager::default());
+        // #830: override rmcp's 300s idle session eviction (DEFAULT_KEEP_ALIVE)
+        // with a long finite TTL so an agent think-gap does not silently kill the
+        // server-side MCP session. The TTL lives on the session manager's
+        // `session_config.keep_alive` (a `pub` field) — NOT on
+        // `StreamableHttpServerConfig`. Set it BEFORE wrapping in the `Arc`.
+        let mut session_manager = LocalSessionManager::default();
+        session_manager.session_config.keep_alive = Some(SESSION_KEEP_ALIVE);
+        let session_manager = Arc::new(session_manager);
         let mut config = StreamableHttpServerConfig::default();
         config.allowed_origins = allowed_origins;
         config.allowed_hosts = allowed_hosts;
--- a/crates/unimatrix-server/src/http/router/tests.rs
+++ b/crates/unimatrix-server/src/http/router/tests.rs
@@ -1146,6 +1146,65 @@ fn mcp_post_with_host(host: &str) -> Request<TestBody> {
 
 const HOST_FORBIDDEN_BODY: &str = "Host header is not allowed";
 
+// ===========================================================================
+// bug #830 (Layer A): MCP session idle keep_alive override.
+//
+// rmcp's LocalSessionManager defaults `session_config.keep_alive` to
+// DEFAULT_KEEP_ALIVE = 300s, evicting idle sessions after 5 minutes → the next
+// call with the stale Mcp-Session-Id gets HTTP 404 and the bridge hangs until a
+// manual /mcp re-init. McpAdapter::new must override it to a long FINITE bound.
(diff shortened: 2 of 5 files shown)


dug-21 (commented):
## Security Review — PR #831 (bugfix-830 self-heal stale cloud MCP session)

**Risk: LOW. No blocking findings. Approve from a security standpoint.**

Fresh-context review of the two-layer fix (Layer A: Rust `router.rs` keep_alive; Layer B: JS bridge `dispatch.js`/`lifecycle.js` 404 self-heal). Read the full diff, both modified source files, surrounding `http-session.js`, the server auth layer, and the added Rust + JS tests.

### Layer A — `SESSION_KEEP_ALIVE = 4h` on `LocalSessionManager.session_config.keep_alive`

- **Not an auth boundary.** Bearer auth is a *separate, outermost* Tower layer (`StaticTokenAuthLayer`, wired at `main.rs:1271-1272`) that validates the 32-byte token in constant time (`ct_eq`, `auth.rs`) on **every** request — independent of MCP session lifetime. Extending the session TTL does not extend, weaken, or bypass any credential check. A request on a kept-alive session is still token-authenticated each time.
- **No unbounded-leak / DoS regression.** The override stays `Some(finite)`. The design explicitly rejected `None` (which would leak sessions forever, growing the session map without bound). The added test `test_session_keep_alive_is_long_finite_bound_not_rmcp_default` asserts finiteness and that it is 4h, and confirms rmcp's default is itself `Some(300s)` (also finite) — so this is a 300s→4h finite-to-finite change, not a finite→unbounded one. Worst case: a client that drops without a clean DELETE holds one session slot for ≤4h instead of ≤5min. Bounded, self-evicting, low blast radius. Per-session memory is small and the map remains capped by the finite TTL.
- Correct field (`session_config.keep_alive`, not `StreamableHttpServerConfig.sse_keep_alive`) — verified against the comment + `test_session_manager_keep_alive_overridden_to_long_bound`.

### Layer B — bridge re-init-and-retry on 404 "session not found"

- **No credential replay / privilege escalation.** The Bearer token is fixed on `HttpSession` at construction and re-sent verbatim on the re-init and retry — it is the *same* token, never stale-elevated. The server re-validates it each request. No credential is derived from the (rotating) session id.
- **No cert-pin bypass.** Re-init and retry both go through the same `request()` → `_pinThenFlush` path; the body (incl. Bearer) is flushed only after `secureConnect` fingerprint match, else fail-loud destroy before any byte. A compromised/spoofed server cannot use the 404 path to weaken pinning or exfiltrate the token.
- **404-spoofing is contained, not exploitable.** A malicious/compromised server *can* force one re-init by returning 404 + "session not found". Impact is bounded to exactly ONE re-init + ONE retry per request (single-flight, no loop) — verified by `test_selfHeal_secondConsecutive404_doesNotLoop`, `test_selfHeal_reinitFailure_aborts_noLoop`, `test_selfHeal_concurrent404s_singleReinit`. No re-auth amplification (the token isn't re-prompted/re-derived), no storm. The trust boundary is the cert pin, which the 404 path cannot cross.
- **vnc-039 audit-attribution contract holds.** `clientInfo.name` stays the byte-identical constant `unimatrix-mcp-bridge` on the replayed init (`this.initMessage` is captured already-stamped; `test_selfHeal_404OnToolsCall...` asserts `reinit.body.params.clientInfo.name === CLIENT_INFO_NAME`). 1-client:1-project scoping is unaffected — only the rotating `Mcp-Session-Id` changes, which is expected and benign.
- **Non-idempotency:** the retry re-sends the original call once. For the failed call this is safe — a 404 "session not found" means the server evicted the session and did NOT execute the request, so there is no duplicated side effect (the request never reached application logic). Retry only fires on that narrow sentinel, never on 5xx/timeout where execution status is ambiguous.
- **404 detection is narrowly gated** (`isSessionNotFound`: status===404 AND `/session not found/i`). Not all 4xx (401/403 stay -32001 auth), not 5xx. Regex is a fixed literal over a 1 MiB-bounded body — no ReDoS. Sentinel `-32099` is out of the JSON-RPC reserved range and is normalized to INTERNAL_ERROR before reaching a client (`handle()` tail) — never leaks.

### Regression risk: LOW
- Clean session end (`teardown` DELETE on EOF) unchanged. Normal error propagation preserved: timeout/unreachable → INTERNAL_ERROR; 401/403 → -32001; notifications still return null. `_send` now returns a message[] consistently; the protocolVersion capture was refactored inline but behavior is equivalent (verified against retained semantics + tests).

### Other checks
- **No new dependencies** (Rust adds only `std::time::Duration`). **No hardcoded secrets/tokens/keys** in the diff — token handling is unchanged and token-free in all log/error strings.
- Diff is minimal and on-scope (the docs/protocol-prune churn in the same branch is non-code and out of security scope).

**Verdict: LOW risk, no blocking findings.**

Issue #841, opened 2026-06-24
bug(personal-cloud): bound mid-stream idle read in mcp-bridge (F6, deferred from #839)


dug-21:
## Context
#839 armed connect + request transport timeouts and routed silent (hung-socket) eviction into the #830 self-heal. The architect design review's **F6** — an idle-read deadline on `readBounded` (`dispatch.js`) and `SseParser.collect` (`sse-parse.js`) for the **connect-then-stall-mid-stream** variant — was **deferred (~770 stripped bytes)** to fit the size cap.

F4's `req.destroy(ETIMEDOUT)` socket timeout already covers most mid-body stalls; this is the residual gap: a connection that establishes, begins streaming, then stalls mid-frame with no idle-read deadline.

## Task
Add an idle-read deadline to `readBounded` and `SseParser.collect` so a mid-stream stall is bounded and surfaces `ETIMEDOUT` → single transparent self-heal (same routing #839 added).

## Acceptance
- A mid-stream idle gap is bounded within a configurable deadline (does not hang).
- Regression test: TLS accept → partial write → stall asserts bounded error + one re-init, no loop.
- Existing #839 + #830 self-heal tests still pass.

Depends on size-cap headroom from the #839 reclaim. Also a candidate seed for the future #837 parity suite.
Deferred from #839.

dug-21:
## Scope note from #842 security review (N1)
The security review of the #839 fix confirms the idle-`timeout` arming makes a **mid-stream SSE rejection path** reachable: a >120s mid-body idle stall destroys the read, and `await dispatchResponse(resp)` sits **outside** `_send`'s try/catch, so it rejects up through `handle` rather than routing into the self-heal. The JSON branch is safe; this is SSE-only.

**Add to this issue's scope:** when adding the mid-stream idle-read deadline, also wrap `dispatchResponse` (the SSE read) so a mid-stream timeout routes into the same `{TRANSPORT_TIMEOUT}` → single self-heal path #839 added, instead of rejecting to the caller. The reported connect-stall bug is fully fixed in #839; this closes the connect-then-stall-mid-stream residue.

dug-21:
Per maintainer direction, **F6 is being delivered within #839 itself** (the complete fix), not deferred. This issue will be **closed as folded into #839** once the #839 PR (#842) lands with the mid-stream idle-read deadline + the `dispatchResponse` SSE-heal-routing (N1) included.

dug-21:
[...]

Issue #840, opened 2026-06-24
tech-debt(hook-client): reclaim ≥1KB stripped bytes to restore mcp-bridge primary size cap to 100000 (raised to 101000 for #839)


dug-21:
## Context
#839 (critical availability fix — arm transport/connect timeout + silent-eviction self-heal in the HTTPS mcp-bridge) came in **irreducibly 579 stripped bytes over** the 100,000-byte PRIMARY cap, even after ~1420B of in-bridge reclaim and deferring F6 (see #839). The core (connect/idle deadline + TRANSPORT_TIMEOUT heal routing + double-settle guard) cannot be split further without removing the connect deadline that *is* the fix.

Per **human decision recorded on #839** (the cap-change rule in `test/check-hook-client-size.js` requires a human call recorded on the GH issue), `PRIMARY_LIMIT` was temporarily raised **100000 → 101000** to ship the fix.

## Task
Reclaim stripped bytes across `packages/unimatrix/lib/hook-client/**` (comment/identifier compaction, dedupe) to bring the stripped total back under 100,000, then **lower `PRIMARY_LIMIT` back to 100000**.

## Acceptance
- `node packages/unimatrix/test/check-hook-client-size.js` passes with `PRIMARY_LIMIT = 100000`.
- Hook-client / mcp-bridge test suite still green.
- No behavior change to the shipped client.

Debt created by #839.

dug-21:
**Update:** #839 delivered the COMPLETE fix (F6 included) in one issue, so `PRIMARY_LIMIT` was raised **100000→102000** (not 101000). Stripped total is now **101483/102000**. This issue's target is unchanged: reclaim ≥~1.5KB stripped bytes across `lib/hook-client/**` and lower `PRIMARY_LIMIT` back to 100000.

**Added urgency:** the **raw BACKSTOP** is now tight at **179818/180000 (182 bytes free)** — reclamation here also protects the raw budget, not just stripped. Comment/identifier compaction reduces both.

dug-21:
## Recommendation from ass-086 (#861) — reground the cap before resetting it

Research spike ass-086 grounded the size threshold this issue resets. Bottom line: **`100,000` does not stand — re-derive to 105,000, and move both caps.** Lowering to an inherited round number forces a documentation/logic reclaim fight against a cap with no defensible basis that will get bumped again on the next addition.

**Why** (full evidence: `product/research/ass-086/FINDINGS.md`):
[...]

PR #842, opened 2026-06-24
fix: arm transport/connect timeout + silent-eviction self-heal in HTTPS mcp-bridge (#839)

### LINKED ISSUE #839: bug(personal-cloud): first MCP call after long client dormancy hangs forever — no transport timeout, self-heal unreachable on a half-open socket
**Related:** #830 / #831 (sibling failure mode — *not* covered by that fix).

## Symptom
A Claude Code client on the HTTPS bridge sat dormant ~6h (client machine likely slept). On resume, the first MCP call (`context_correct`) **hung indefinitely** — no timeout, no error, no reconnect — until the user pressed escape.

Note: the *"tool use was rejected"* message reported by the agent is **Claude Code's escape/interrupt artifact, not a server or permission error** — it is not diagnostic. The real behavior is an infinite hang on the first post-dormancy call.

## Root cause (verified in code)
The first post-dormancy call stalls on a dead/half-open TCP path (peer gone, no RST after the client slept). Three independent gaps line up:

1. **No transport timeout.** `mcp-bridge/http-session.js:98-129` builds the request with no `timeout` key (`_options`, `:59-73`) and wires only `response`/`error` handlers. On a half-open socket neither event fires, so the request Promise **never settles** → the `await` at `lifecycle.js:65` blocks forever. (The only `setTimeout` in the bridge is a 1s teardown-DELETE guard, `http-session.js:144`.)
2. **Self-heal can't engage.** #830's re-init fires **only** on a returned `HTTP 404` + body matching `/session not found/i` (`dispatch.js:22-24`, `lifecycle.js:45,109-112`). A hung socket returns *nothing*, so `lostId` is never evaluated and the re-init branch is unreachable. (A clean network with an evicted session *would* 404 and self-heal correctly; this case never gets that far.)
3. **No client keep_alive / liveness probe.** "#830 keep_alive" is purely the **server-side idle TTL** raised 300s→4h (`router.rs:58,364`) — there is **no client-side ping** (`setInterval`/`keepAlive` grep is empty). 6h > 4h, so the session was evicted regardless; nothing on the client detects the suspension gap or probes liveness before the first call.

**The irony:** clean-failure handling already exists — `lifecycle.js:68` maps `ETIMEDOUT` to a proper JSON-RPC "MCP endpoint timed out" error. Nothing ever arms a timeout to produce it.

## Fix direction (for design — not prescriptive)
1. **Arm a request/connect timeout** in `http-session.js:request()` — pass `timeout:` in `_options` AND wire `req.on("timeout", () => req.destroy(<ETIMEDOUT-coded error>))` (Node's `timeout` option does not auto-destroy). Also guard the TLS-connect phase, since the stall can be in the handshake, not just an idle established socket.
2. **Route a transport timeout into re-init**, not just an error return — `lifecycle.js` currently branches to `_reinit` on 404 only. Treat "transport timeout → re-init once" so the user sees a quick reconnect instead of a hang→escape. Guard against the re-init POST itself hanging on the same dead socket (it shares the timeout-less `_send`).
3. Bound the SSE read (`sse-parse.js:10-33`) and `readBounded` (`dispatch.js:32-49`) for the connect-then-stall-mid-stream variant.
4. (Optional, larger scope — could be a follow-up) a wake / idle-gap **liveness probe** before the first call after a long idle, so the first post-sleep call doesn't eat one slow failure.

## Blast radius
100% of live request/response MCP traffic over the cloud bridge shares this chokepoint (`lifecycle.handle → _send → session.request → un-timed Promise`, `http-session.js:98`): every `context_*` tool, **and `initialize`** (a hung init blocks bridge startup and can't self-heal even in principle — `lifecycle.js:42`), **and the self-heal's own re-init POST** (`lifecycle.js:99`, same `_send`). Only the EOF teardown DELETE is guarded.

## Acceptance
After a dormancy exceeding the server TTL (and/or a simulated half-open socket), the first MCP call **fails fast within a bounded timeout and transparently re-inits** — the user sees a brief reconnect, not a hang. No live MCP call can block indefinitely on a dead socket. Regression guard: a clean `404 session-not-found` eviction still self-heals (do not break #830 Layer B).

## Vision alignment
Hits `personal-cloud` reliability head-on — a dormant client that hangs on first use is exactly the ops-friction the goal exists to remove, and it undercuts C0's "first-class, not degraded" promise. Also a live counterexample to **nan-021 (#836) NFR-7**'s assumption that the shipped self-heal covers eviction: it covers *signalled* (404) eviction, not *silent* (hung-socket) eviction.

## Key files
- `packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js:98-129` (fix site — no timeout on `request()`)
- `packages/unimatrix/lib/hook-client/mcp-bridge/lifecycle.js:45,62-70,109-112` (self-heal trigger — 404-only)
- `packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js:22-29` (SESSION_NOT_FOUND sentinel gate)
- `crates/unimatrix-server/src/http/router.rs:58,364` (#830 server keep_alive 4h TTL)

dug-21:
## Summary
Fixes #839 — the first MCP call after long client dormancy hung **forever** on a half-open socket. The HTTPS mcp-bridge armed no transport timeout, so the request Promise never settled and #830's self-heal (gated only on a 404 `SESSION_NOT_FOUND`) was unreachable on a silently-evicted/hung socket. This delivers the **complete** fix (connect/idle/request timeout + self-heal routing + mid-stream idle-read).

## Root cause (3 aligned gaps)
1. `http-session.js:request()` built options with no `timeout` and wired only response/error handlers → half-open socket never settles the Promise → `lifecycle.js` await blocks forever (TLS `secureConnect` with `agent:false` also had no deadline).
2. Self-heal fired only on the 404 `SESSION_NOT_FOUND` sentinel → a hung socket returns nothing → re-init branch dead.
3. `lifecycle.js` already mapped `ETIMEDOUT` → "MCP endpoint timed out" but nothing armed a timeout.

## Fix
- Arm a **request + connect/TLS-handshake timeout** that destroys the socket with `ETIMEDOUT` (Node does not auto-destroy on the `timeout` option); deadlines configurable + generous (15s connect / 120s idle) so slow-but-healthy calls are never aborted.
- Emit a distinct **`TRANSPORT_TIMEOUT` (-32098)** sentinel and widen the self-heal predicate to `{404, transport-timeout}` **only**, routed through the existing **single-flight `_reinit`** (one attempt, normalize-on-exhaustion).
- **Double-settle guard** (vnc-039) — timeout reject gates on settled-state/error-code, not the `flushed` flag.
- **F6 — mid-stream idle-read deadline:** per-chunk-reset idle deadline in `readBounded` (`dispatch.js`) and `SseParser.collect` (`sse-parse.js`), destroying the stream with `ETIMEDOUT`. `_send` now reads the body **inside** its try/catch, so an SSE/JSON mid-stream timeout maps to `TRANSPORT_TIMEOUT` and routes into the same single self-heal (closes the connect-then-stall-mid-stream variant + the security N1 finding). **Supersedes follow-up #841.**

## Tests
- "transport timeout self-heal (#839)" block: **8 tests** — connect-stall fail-fast; timeout → exactly one transparent re-init then success; re-init that also stalls → bounded error, no loop; generic transport error does NOT heal; **+3 mid-stream**: `midStreamStall_boundedByIdleReadDeadline`, `midStreamStall_singleReinitThenSuccess`, `midStreamReinitAlsoStalls_boundedError_noLoop`.
- #830 clean-404 self-heal regression block (4) retained and green.
- Verified: full hook-client **762/0/1**, integration smoke **24/24**, HTTPS-UDS parity **12/12**, lifecycle/protocol green, `cargo build` OK. JS/TS-only diff (no Rust).

## Size cap (#840)
The complete fix is **101483** stripped bytes vs the 100,000-byte hook-client primary cap. Per the cap-change rule (human decision recorded on #839), `PRIMARY_LIMIT` was raised **100000→102000** to land the full fix in one issue. Reclaiming back to 100000 is tracked in **#840**. (Raw backstop now tight at 179818/180000.)

Closes #839. Supersedes #841.

### DIFF
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
@@ -15,9 +15,20 @@ const SESSION_NOT_FOUND = -32099;
 // timed out (#839). Heals through the same single-flight re-init as 404, then
 // normalizes to a generic error if exhausted — never reaches the client.
 const TRANSPORT_TIMEOUT = -32098;
+// Idle-read deadline (#839 F6): a connection that establishes, streams, then
+// stalls mid-frame bounds BYTES but not stall TIME. On expiry destroy + throw an
+// ETIMEDOUT-coded error so _send maps it to the SAME TRANSPORT_TIMEOUT sentinel
+// as the connect/request path (N1: routes the SSE branch into self-heal).
+const DEFAULT_READ_MS = 120000;
 
 function jsonRpcError(id, code, message) {
   return { jsonrpc: "2.0", id: id === undefined ? null : id, error: { code, message } };
+}
+
+// ETIMEDOUT-coded (mid-stream read stall) — same code as the connect/request
+// path, so lifecycle._send routes it into self-heal via TRANSPORT_TIMEOUT.
+function readTimedOut() {
+  return Object.assign(new Error("MCP endpoint timed out"), { code: "ETIMEDOUT" });
 }
 
 // Narrow 404 idle-eviction class (#830): 404 with a "session not found" body.
@@ -32,23 +43,32 @@ function httpStatusToJsonRpc(status, body) {
   return isSessionNotFound(status, body) ? SESSION_NOT_FOUND : INTERNAL_ERROR;
 }
 
-// Read a non-SSE body bounded at `limit`; destroy + return-what-we-have on excess.
-function readBounded(res, limit) {
-  return new Promise((resolve) => {
+// Read a non-SSE body bounded at `limit` bytes AND `readMs` idle-stall (#839 F6).
+// Byte overflow / end / error / close resolve with what we have; an idle-read
+// stall rejects ETIMEDOUT (the throw routes into self-heal via _send). The idle
+// timer resets per chunk.
+function readBounded(res, limit, readMs) {
+  return new Promise((resolve, reject) => {
     const chunks = [];
     let received = 0;
+    let t;
+    const stop = () => { if (t) { clearTimeout(t); t = null; } };
+    const arm = () => { stop(); t = setTimeout(() => { try { res.destroy(); } catch (_e) {} reject(readTimedOut()); }, readMs || DEFAULT_READ_MS); if (t.unref) t.unref(); };
+    const ok = () => { stop(); resolve(Buffer.concat(chunks)); };
+    arm();
     res.on("data", (c) => {
       if (received < limit) {
         chunks.push(c);
         received += c.length;
+        arm();
       } else {
         try { res.destroy(); } catch (_e) {}
-        resolve(Buffer.concat(chunks));
+        ok();
       }
     });
-    res.on("end", () => resolve(Buffer.concat(chunks)));
-    res.on("error", () => resolve(Buffer.concat(chunks)));
-    res.on("close", () => resolve(Buffer.concat(chunks)));
+    res.on("end", ok);
+    res.on("error", ok);
+    res.on("close", ok);
   });
 }
 
@@ -61,11 +81,12 @@ function idFromBody(buf) {
   }
 }
 
-// dispatchResponse({ status, contentType, res }) -> Promise<jsonRpcMessage[]>
-async function dispatchResponse({ status, contentType, res }) {
+// dispatchResponse({ status, contentType, res }, readMs?) -> Promise<jsonRpcMessage[]>
+// An idle-read stall in either branch rejects ETIMEDOUT (caught by _send -> heal).
+async function dispatchResponse({ status, contentType, res }, readMs) {
   const ct = contentType || "";
   if (ct.startsWith("application/json")) {
-    const body = await readBounded(res, BODY_LIMIT_BYTES);
+    const body = await readBounded(res, BODY_LIMIT_BYTES, readMs);
     if (status >= 400) {
       return [jsonRpcError(idFromBody(body), httpStatusToJsonRpc(status, body), "MCP endpoint returned HTTP " + status)];
     }
@@ -76,7 +97,7 @@ async function dispatchResponse({ status, contentType, res }) {
     }
   }
   if (ct.startsWith("text/event-stream")) {
-    const messages = await SseParser.collect(res, BODY_LIMIT_BYTES);
+    const messages = await SseParser.collect(res, BODY_LIMIT_BYTES, readMs);
     if (status >= 400 && messages.length === 0) {
       return [jsonRpcError(null, httpStatusToJsonRpc(status), "MCP endpoint returned HTTP " + status)];
     }
@@ -102,4 +123,5 @@ module.exports = {
   INTERNAL_ERROR,
   SESSION_NOT_FOUND,
   TRANSPORT_TIMEOUT,
+  DEFAULT_READ_MS,
 };
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
@@ -11,6 +11,10 @@ const INTERNAL_ERROR = -32603;
 // 404 "session not found" so lifecycle.js can self-heal (#830, design-review C1).
 // Never reaches the client: lifecycle re-inits + retries before surfacing.
 const SESSION_NOT_FOUND = -32099;
+// Bridge-internal sentinel for a silent (hung-socket) eviction: the transport
+// timed out (#839). Heals through the same single-flight re-init as 404, then
+// normalizes to a generic error if exhausted — never reaches the client.
+const TRANSPORT_TIMEOUT = -32098;
 
 function jsonRpcError(id, code, message) {
   return { jsonrpc: "2.0", id: id === undefined ? null : id, error: { code, message } };
@@ -97,4 +101,5 @@ module.exports = {
   isSessionNotFound,
   INTERNAL_ERROR,
   SESSION_NOT_FOUND,
+  TRANSPORT_TIMEOUT,
 };
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
@@ -12,18 +12,13 @@ const { applyCertPin, verifyPeerFingerprint } = require("../cert-pin.js");
 // rmcp 1.7.0 wire constants (OVERVIEW "VERIFIED WIRE VALUES"; CONFIRM LIVE).
 const SESSION_HEADER = "Mcp-Session-Id";
 const SESSION_HEADER_LC = "mcp-session-id";
-const ACCEPT_VALUE = "application/json, text/event-stream";
-const CONTENT_TYPE_REQUEST = "application/json";
 const BODY_LIMIT_BYTES = 1048576;
+// Generous defaults (ms) so slow-but-healthy calls (large embeddings) are never
+// aborted; the tighter connect/TLS-handshake deadline catches the half-open
+// socket (agent:false → secureConnect can hang; socket `timeout` misses it).
 
 function stripIPv6Brackets(h) {
   return h && h.startsWith("[") && h.endsWith("]") ? h.slice(1, -1) : h;
-}
-
-// Fail-loud exit on pin mismatch; token-free message (NFR-06).
-function failLoud(session, message) {
-  try { session.errOut("mcp-bridge: " + message + "\n"); } catch (_e) {}
-  session.exit(1);
 }
 
 class HttpSession {
@@ -31,7 +26,7 @@ class HttpSession {
     return new HttpSession(opts);
   }
 
-  constructor({ mcpUrl, token, pinnedFp, exit, errOut }) {
+  constructor({ mcpUrl, token, pinnedFp, exit, errOut, connectMs, idleMs }) {
     this.url = new URL(mcpUrl); // POSTed VERBATIM (AC-05)
     this.token = token;
     this.pinnedFp = pinnedFp;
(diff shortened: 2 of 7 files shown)


dug-21 (commented):
## Security Review — 839-security-reviewer (fresh context)

**Risk level: LOW. No blocking findings. Approve to merge.**

Read cold: full `git diff main...HEAD`, all three modified bridge units + `sse-parse.js`/`dispatch.js` in full, the stub-server helper, the investigator report and design review on #839, and ADR-001 (#5115 pin-then-flush) + #5124 (vnc-039 C2 double-settle trap). Ran the gates: 49/49 bridge suite green (incl. the new #839 block AND the retained #830 non-regression block); size gate `stripped=100579/101000` green.

### OWASP / trust-boundary
- **No new trust boundary, no injection/deserialization/access-control surface.** The change adds two timers + an internal sentinel; it parses nothing new and crosses no new boundary. `req.destroy(ETIMEDOUT)` carries no attacker-controlled data.
- **Pin-then-flush invariant (ADR-001 #5115) PRESERVED.** `req.end(body)` is still reached only inside the `secureConnect` match branch (`onPin`). The new `onConnect` callback clears the connect timer and nothing else — it does not move the flush. The Bearer never reaches an unverified or timed-out socket. Verified the `flushed`-guarded swallow still holds for the post-mismatch destroy.
- **Secrets:** none. The only `Bearer` line is the pre-existing `"Bearer " + this.token`; `this.token` is store-sourced at spawn (proven by `tokenReadFromStoreAtSpawn_notFromArgv`). Test literals (`token:"TKN"`) are stubs.

### Heal-predicate narrowness — CONFIRMED restricted to {404, TRANSPORT_TIMEOUT}
`lostId` gates on `m.error.code in SENTINEL_TEXT`, whose keys are exactly `SESSION_NOT_FOUND (-32099)` and `TRANSPORT_TIMEOUT (-32098)`. Auth (`-32001`) and `INTERNAL_ERROR (-32603)` are NOT keys — verified at runtime (`-32001 in T → false`). `_send` maps only `err.code==="ETIMEDOUT"` to the timeout sentinel; every other transport fault → `INTERNAL_ERROR` → no heal. `test_timeout_doesNotHealOnGenericTransportError` (ECONNREFUSED → `initCount===1`, no retry) locks this. **No "any error" widening; no auth/5xx re-init storm.**

### DoS / re-init storm — single-attempt confirmed
Single-flight `reinitInFlight` + retry only `if (ok)` + a timed-out re-init yields `newId==null` → `_doReinit` returns false → original error surfaces, normalized, **no recursion**. `test_timeout_reinitAlsoStalls_boundedError_noLoop` asserts `initCount===2` (one heal attempt) and the sentinel never leaks. A flaky network can at worst induce one transparent re-init per evicted call, then a bounded error — strictly more resilient than main's infinite hang. An attacker who could force a client timeout could already drop the socket; no new amplification.

### Double-settle (vnc-039 C2 / #5124) — safe
Single `settled` flag; `settle()` is the sole resolve/reject path for response, mismatch, and error. The `req.on("error")` swallow now gates on `err.code !== "ETIMEDOUT"` (not `flushed`), so a post-flush timeout reject is NOT swallowed and the trailing error event after `kill`'s `req.destroy` is absorbed by the `settled` guard. Matches #5124's prescription exactly.

### Socket-leak / fire-after-clean-response — safe
Connect timer is `.unref()`'d and cleared on `secureConnect` or in `settle` (both paths). The socket `timeout:idleMs` (120s) stays armed after a clean response — by design — so a mid-stream idle stall destroys the read rather than hanging; the trailing error is swallowed (already settled). No leak, no double-settle, fail-safe.

### Non-blocking observations
- **N1 (low, tracked #841 / F6):** the idle `timeout` arming makes a *new* mid-stream rejection path reachable for the **SSE branch only**: a healthy `text/event-stream` response that idle-stalls >120s gets `req.destroy(ETIMEDOUT)`, rejecting `SseParser.collect` → `dispatchResponse`, which runs **outside** `_send`'s try/catch (`await dispatchResponse(resp)` is unguarded). The JSON branch is safe (`readBounded` resolves via `res.on("close")`). On main this case hung forever; now it can reject up through `handle`. This is exactly the deferred F6 variant (idle-read deadline on `readBounded`/`SseParser.collect`), explicitly tracked in #841. Requires a >120s pathological mid-body stall; the reported bug (connect/handshake stall) is fully fixed and fail-fast (`test_timeout_connectStall_failsFastWithinDeadline`). Recommend #841 also wrap `dispatchResponse` so the timeout routes into the heal path instead of rejecting. **Not blocking** — pre-existing unguard, newly reachable only under the deferred variant.
- **N2 (info, not security):** `PRIMARY_LIMIT` 100000→101000 is the recorded human decision on #839 (reclaim tracked #840). Confirmed irreducible: stripped is 100579, i.e. 579 over the old cap. Noted per scope; not a finding.

### Blast radius
100% of live MCP traffic flows through `request()`. Worst case of a subtle bug here is a too-tight timeout aborting slow-but-healthy calls — mitigated by generous configurable defaults (connect 15s, idle 120s) and the fail-fast connect-stall test. Failure mode is safe (bounded JSON-RPC error or one transparent re-init), never silent data corruption or token exposure.

**Verdict: LOW risk, no blocking findings.** The fix arms what ADR-001's machinery already expected, keeps the heal trigger narrow, caps at one re-init, and preserves the pin-then-flush trust contract.

dug-21 (commented):
## Security Review (DELTA: F6 + N1) — 839-security-reviewer-2 (fresh context)

**Risk level: LOW. No blocking findings. N1 is RESOLVED. Approve.**

Reviewed cold the delta `83dc0540..HEAD` (commits 39bcc3cd F6/N1, 662cd89b size-cap): `dispatch.js` (`readBounded`), `sse-parse.js` (`SseParser.collect`), `lifecycle.js` (`_send`), all three modified files in full, the new mid-stream test block, the prior 839-security-reviewer comment, and `http-session.js` (idleMs/BODY_LIMIT source). Ran gates: bridge suite **52/52** (incl. F6/N1 cases d/e/f), full hook-client **787/0 fail**, size gate green (stripped=101483/102000).

### N1 — RESOLVED
The prior note: `await dispatchResponse(resp)` sat OUTSIDE `_send`'s try/catch, so a >idle SSE mid-body stall rejected up through `handle` instead of routing into self-heal. Now `out = await dispatchResponse(resp, this.session.idleMs)` is INSIDE the try (lifecycle.js:65-71). A mid-stream `ETIMEDOUT` (from `readBounded` reject or `SseParser.collect` rethrow) hits the SAME `if (err.code === "ETIMEDOUT") -> TRANSPORT_TIMEOUT` branch as the connect path. SSE branch now reaches parity with connect/request. Confirmed by `test_timeout_midStreamStall_singleReinitThenSuccess` (initCount===2).

### Single-flight / single-attempt / normalize-on-exhaustion — PRESERVED
Moving the body-read inside the try did NOT alter the heal machinery. `handle` still gates re-init on `lostId` ∈ {SESSION_NOT_FOUND, TRANSPORT_TIMEOUT}, `_reinit` is still single-flight (`reinitInFlight`), retry is still `if (ok)` once. A re-init whose OWN read stalls mid-stream yields a timeout sentinel inside `_doReinit._send`, `newId` stays unchanged -> returns false -> no retry, no recursion. Locked by `test_timeout_midStreamReinitAlsoStalls_boundedError_noLoop` (initCount===2, sentinel normalized, never -32098 to client).

### Double-settle (vnc-039) — SAFE
- `readBounded`: `arm()`/`ok()` both call `stop()` (clearTimeout + null). Promise resolve/reject are idempotent by JS semantics — a late `reject(readTimedOut())` after `ok()` is a no-op. No fire-after-clean-end.
- `SseParser.collect`: the timer callback ONLY calls `res.destroy(err)` — it does NOT settle a Promise. The async iterator is the single settle channel; the destroyed stream makes `for await` rethrow once. `finally { clearTimeout(t) }` fires on every exit (end/error/overflow-break/throw). No second settle path exists.
- `res.destroy()` is idempotent in Node (destroyed-state guarded) and additionally try/caught in both files.

### Timer hygiene — no leaks / dangling sockets / unref
Per-chunk `arm()` clears the prior timer before re-arming; `.unref()` on every armed timer so a stray timer never holds the event loop open. `readBounded` clears on data-overflow/end/error/close; `collect` clears in `finally`. The byte-overflow→finally micro-window is benign: a timer firing there only re-`destroy()`s an already-destroyed stream (no-op), and cannot settle anything.

### Heal-predicate narrowness — INTACT, no slow-healthy false heal
Predicate still `m.error.code in SENTINEL_TEXT` = {-32099, -32098} only. The idle deadline is `this.session.idleMs` (default 120000, configurable; http-session.js:39) threaded through — generous, never undefined (the `|| DEFAULT_READ_MS` fallback is belt-and-suspenders). Per-chunk RESET means only a true >120s mid-frame silence trips it, not a slow-but-steady stream. `test_timeout_doesNotHealOnGenericTransportError` still locks {404,timeout}-only.

### DoS / re-init storm — bounded
Worst case under a malicious server repeatedly stalling mid-stream: exactly ONE transparent re-init per evicted call, then a bounded normalized error. No unbounded retry, no recursion. Strictly more resilient than main (which hung forever). An attacker forcing a mid-stream stall could already drop the socket — no new amplification.

### Pin-then-flush (ADR-001) & secrets — UNAFFECTED
The delta touches only the response read-path (post-`secureConnect`, post-flush). No change to `req.end(body)` / `onPin` / the Bearer path. `req.destroy`/`res.destroy(ETIMEDOUT)` carry no attacker-controlled data and no token. No bearer exposure on a timed-out stream. No new secrets, no new deps, no new trust boundary, no injection/deserialization surface.

### Regression
SSE happy-path (`collect` records emit unchanged; timer is pure-additive and cleared in `finally`), #830 clean-404 self-heal, and the #839 connect/request block all green in the 52/52 run. Tester's 762/0/1 (full), #839 8/8, #830 4/4 are consistent with the 787/0-fail hook-client run observed here.

### Blast radius
100% of MCP traffic flows through `_send`/`dispatchResponse`. Worst case of a subtle delta bug: a too-tight idle abort on a slow-healthy SSE stream — mitigated by the 120s configurable default + per-chunk reset. Failure mode is fail-safe (bounded JSON-RPC error or one transparent re-init), never silent data corruption or token leak.

**Verdict: LOW risk, no blocking findings. N1 resolved, F6 lands the deferred idle-read deadline cleanly within the established heal contract.**

PR #848, opened 2026-06-26
fix: ref F6 idle-read deadline timers so they fire in an idle loop (#847)

### LINKED ISSUE #847: bug(personal-cloud): #839 F6 mid-stream idle-read deadline never fires when event loop is otherwise idle (unref'd timer) — red CI on Node 18/22, blocks v0.8.8 release
**Related:** #839 / #842 (introduced here) · blocks the v0.8.8 release (cancelled — nothing published).

## Symptom
The 3 F6 mid-stream tests added in #842 (commit `39bcc3cd`) **fail on Node 18 & Node 22, pass on Node 24** — across OSes. CI for PR #842 was red on this and was merged anyway (merged on local single-Node green without checking the cross-platform matrix). The failures **cascade** into unrelated test files (`identity across spawns/projects (R-17)`, `store-error posture (R-13)`) via `cancelledByParent`.

Failing tests (`packages/unimatrix/test/hook-client/mcp-bridge.test.js`):
- `test_timeout_midStreamStall_boundedByIdleReadDeadline` (:676)
- `test_timeout_midStreamStall_singleReinitThenSuccess` (:693)
- `test_timeout_midStreamReinitAlsoStalls_boundedError_noLoop`

CI evidence (PR #842, run 28174788059, job `Hook client suite (Node 22 / ubuntu-latest)` 83448088468):
```
not ok 6 - test_timeout_midStreamStall_boundedByIdleReadDeadline
  failureType: 'cancelledByParent'
  error: 'Promise resolution is still pending but the event loop has already resolved'
  code: 'ERR_TEST_FAILURE'
```
Note the suite footer prints `# fail 0` despite the `not ok` lines — the top-level counter is misleading; the job still exits non-zero. Node 24 (windows + macos) PASS; Node 18 (windows, ubuntu) + Node 22 (ubuntu) FAIL.

## Root-cause hypothesis (needs verification, likely a real runtime defect)
`'Promise resolution is still pending but the event loop has already resolved'` means the awaited mid-stream-timeout Promise **never settles because nothing keeps the event loop alive** — the idle-read deadline timer never fires.

Prime suspect: the idle-read deadline timer in `readBounded` (`dispatch.js`) and/or `SseParser.collect` (`sse-parse.js`) is **`.unref()`'d**. An `.unref()`'d timer does not hold the loop open, so when the only other handle is a silent/half-open socket and there is no other pending work, the loop drains and the deadline **never triggers** → the timeout that is supposed to fail fast instead hangs. (The #842 security review specifically praised the `.unref()` for "clean timer hygiene" — that may be exactly what defeats the deadline. Node 24 apparently keeps the loop alive differently, masking it.)

If confirmed, this is **ironic and material**: the whole #839 fix exists to stop hangs, and an `.unref()`'d deadline can itself fail to fire in an otherwise-idle process — i.e., the mid-stream guard may not actually fire in production under the exact "client went quiet" condition it targets. Must determine whether this is test-only or also reachable at runtime.

## Scope
- `packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js` (`readBounded` idle-read timer)
- `packages/unimatrix/lib/hook-client/mcp-bridge/sse-parse.js` (`SseParser.collect` idle-read timer)
- `packages/unimatrix/test/hook-client/mcp-bridge.test.js` (the 3 mid-stream tests — fix any test-harness assumption AND confirm the runtime behavior)

Mind the hook-client **size cap** (currently 102000, #840) — any change must keep `node test/check-hook-client-size.js` green; raw backstop is tight (179818/180000).

## Acceptance
- Mid-stream idle-read deadline **reliably fires ETIMEDOUT → single self-heal** even when the event loop is otherwise idle (the half-open / client-went-quiet case), confirmed by a test that would catch the unref regression.
- Determine + document whether the original defect was runtime-reachable or test-only; if runtime, the fix restores the F6 guarantee #842 claimed.
- **Green across the full Node CI matrix** (Node 18/20/22/24 × ubuntu/macos/windows) — no `cancelledByParent` cascade, no leaked handles.
- #839 connect-stall block + #830 clean-404 self-heal still green.
- Budget a **post-PR CI round-trip** before merge (do not merge on local single-Node green).

## Release impact
Blocks re-release. v0.8.8 was tagged + the version-bump commit (`2662db91`) landed on `main`, but the release run was cancelled before publish — **npm 0.8.8 is unpublished/free, no GitHub release**. Plan: fix here, then re-release (recommend **v0.8.9**; the inert `v0.8.8` tag to be deleted at re-release). Note: `release.yml` does **not** run the JS hook-client suite (only Rust `cargo test`), so the release pipeline will NOT catch this class of failure — CI (`ci.yml`) is the only gate for it.

dug-21:
Fixes #847.

## Root cause
The two F6 mid-stream idle-read deadline timers were `.unref()`'d:
- `dispatch.js` `readBounded` (line 56)
- `sse-parse.js` `SseParser.collect` (line 23)

An unref'd timer doesn't keep the libuv loop alive on its own. With a handle-free mock stream as the sole loop participant, the loop drains before the deadline, so the F6 deadline never fires on Node 18/20/22 (Node 24's test runner masks it) — the awaited timeout promise never settles, cascading `cancelledByParent` into sibling suites (R-17, R-13). This was red CI on PR #842 and blocks the v0.8.8 re-release.

## Fix
Remove `.unref()` from both load-bearing deadlines and mark them intentionally ref'd (Note A comments). Both timers are already cleared on every settle path (`readBounded` via `stop()`, `collect` via its `finally`), so a ref'd timer is only pending during an active read and never delays clean process exit. The connect deadline (`ct`) and teardown timer stay unref'd — a documented two-tier convention (load-bearing read deadlines ref'd, cleanup-grade timers unref'd).

**Runtime impact: test-only.** In production the stalled stream keeps a ref'd socket handle (so the unref'd timer still fired), and `http-session.js` independently bounds a mid-stream stall via its req timeout. The fix hardens the F6 guard to be self-sufficient rather than relying on those coincidences.

## Tests
- 2 new direct regression tests drive a handle-free stall with no competing ref'd handle and assert bounded settle — they fail deterministically if `.unref()` is re-added.
- Full Node suite: 1019 pass / 0 fail / 0 cancelled; the 3 previously-failing F6 mid-stream tests pass; no `cancelledByParent` cascade.
- Hook-client size cap green: stripped 101445/102000, raw 179944/180000.

## Merge gate
Local validation is Node 24 only, which **masks** this bug. The cross-platform `ci.yml` Node 18/20/22/24 × ubuntu/macos/windows matrix is the real gate — **do not merge on local single-Node green.** `release.yml` runs only Rust tests, so `ci.yml` is the sole gate for this class of failure.

### DIFF
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/dispatch.js
@@ -53,7 +53,8 @@ function readBounded(res, limit, readMs) {
     let received = 0;
     let t;
     const stop = () => { if (t) { clearTimeout(t); t = null; } };
-    const arm = () => { stop(); t = setTimeout(() => { try { res.destroy(); } catch (_e) {} reject(readTimedOut()); }, readMs || DEFAULT_READ_MS); if (t.unref) t.unref(); };
+    // Kept REF'd (#847): load-bearing idle deadline; cleared on every settle via stop().
+    const arm = () => { stop(); t = setTimeout(() => { try { res.destroy(); } catch (_e) {} reject(readTimedOut()); }, readMs || DEFAULT_READ_MS); };
     const ok = () => { stop(); resolve(Buffer.concat(chunks)); };
     arm();
     res.on("data", (c) => {
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/sse-parse.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/sse-parse.js
@@ -20,7 +20,8 @@ class SseParser {
     let buffer = "";
     let total = 0;
     let t;
-    const arm = () => { if (t) clearTimeout(t); t = setTimeout(() => { try { res.destroy(Object.assign(new Error("MCP endpoint timed out"), { code: "ETIMEDOUT" })); } catch (_e) {} }, readMs || DEFAULT_READ_MS); if (t.unref) t.unref(); };
+    // Kept REF'd (#847): load-bearing idle deadline; cleared in the finally below.
+    const arm = () => { if (t) clearTimeout(t); t = setTimeout(() => { try { res.destroy(Object.assign(new Error("MCP endpoint timed out"), { code: "ETIMEDOUT" })); } catch (_e) {} }, readMs || DEFAULT_READ_MS); };
     arm();
     try {
       for await (const chunk of res) {
--- a/packages/unimatrix/test/hook-client/mcp-bridge.test.js
+++ b/packages/unimatrix/test/hook-client/mcp-bridge.test.js
@@ -17,7 +17,7 @@ const http = require("http");
 const { startSilentTcpServer } = require("../helpers/stub-server.js");
 const { StdioFramer } = require("../../lib/hook-client/mcp-bridge/stdio-frame.js");
 const { SseParser } = require("../../lib/hook-client/mcp-bridge/sse-parse.js");
-const { dispatchResponse, correlateById, isSessionNotFound, SESSION_NOT_FOUND } = require("../../lib/hook-client/mcp-bridge/dispatch.js");
+const { dispatchResponse, correlateById, isSessionNotFound, SESSION_NOT_FOUND, readBounded } = require("../../lib/hook-client/mcp-bridge/dispatch.js");
 const { Lifecycle, CLIENT_INFO_NAME } = require("../../lib/hook-client/mcp-bridge/lifecycle.js");
 const { HttpSession, SESSION_HEADER_LC } = require("../../lib/hook-client/mcp-bridge/http-session.js");
 const bridge = require("../../lib/hook-client/mcp-bridge.js");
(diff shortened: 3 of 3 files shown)


dug-21 (commented):
## Security Review — Risk: LOW · Non-blocking · Approve

Fresh-context review (847-security-reviewer). Read the full diff cold, both modified lib files in full, the `stallingSse` helper, and re-ran the size gate.

### Verified claims
- **Clear-on-every-settle-path holds — a ref'd timer CANNOT pin the loop open after legitimate completion.**
  - `dispatch.js readBounded`: the only places the promise settles are `ok()` (data-overflow / `end` / `error` / `close`) which calls `stop()` → `clearTimeout`, and the timer callback itself (`reject(ETIMEDOUT)`), which by definition has already fired and is no longer pending in libuv. No settle path leaves an armed timer. ✔
  - `sse-parse.js collect`: the `finally { if (t) clearTimeout(t); }` runs on normal return, `break`, and throw — every exit. ✔
  - Net: the ref'd timer is pending only during an active in-flight read; it is cleared before the promise settles. It cannot keep a CLI/hook process alive past exit (no DoS-by-hang / handle leak).
- **Regression to clean exit: none.** During an in-flight read the socket handle already keeps the loop alive, so ref'ing the timer adds no liveness beyond the read's own duration; after settle it's cleared. Removing `.unref()` does not delay clean process exit in normal (non-stalled) operation.
- **No new dependencies.** Only 3 files touched; no `package.json` / lockfile changes.
- **No secrets introduced.**
- **Size cap green (re-ran `check-hook-client-size.js` against the branch):** stripped 101445/102000, raw 179944/180000 — matches the PR claim.

### OWASP surface
Negligible. No injection (no shell/SQL/format-string with untrusted input), no access-control or trust-boundary change, no new external input (`readMs` is caller-threaded config). Timeout error message is generic ("MCP endpoint timed out") — no internal-state leak. **Net-positive security effect:** the idle-read deadline now fires reliably in an otherwise-idle loop, hardening the F6 guard against a slowloris-style mid-stream stall rather than relying on a coincidental ref'd socket.

### Observations (non-blocking)
1. **Raw size headroom is very tight (56 bytes: 179944/180000).** Not a defect of this PR — the change actually keeps it green — but the next lib edit risks breaching the raw backstop. Worth a comment-density pass at the next touch.
2. The two-tier convention (load-bearing read deadlines ref'd, cleanup/connect timers unref'd) is now load-bearing; the inline Note A comments document it adequately.

**Blocking findings: none.** Recommend merge once the cross-platform `ci.yml` matrix (Node 18/20/22/24 × ubuntu/macos/windows) is green — that is the real gate for this class of failure, per #847.

Issue #853, opened 2026-06-26
feat(personal-cloud): standalone multi-tenant HTTP isolation test — a write for slug A lands only in A's store (N3 #5161)


dug-21:
## Summary

Standalone, single-surface HTTP behavioral cross-tenant isolation test — **test-only**, extending the infra-001 harness cumulatively (no fork, no scaffold). Cross-tenant isolation is an **HTTP-only property** (local is 1-client:1-project by the ADR-006 compile-time guard, so there is no local tenant to be "at parity" with). It was mistakenly scoped as parity dimension **D6** inside the nan-022 / C0 cross-transport parity matrix and was removed from there (#845). This is its proper home: the genuine behavioral isolation test surfaced by ass-084 (#850, GO) and the #845 investigation, decoupled from parity.

Feature ID: **infra-003** (artifacts under `product/test/infra-003/`, matching the infra-001 / infra-002 precedent).

## Scope — standalone, single-surface HTTP behavioral test (NOT a parity comparison, NOT in the parity harness)

The shipped container already multiplexes N slugs via `MultiProjectRouter` (vnc-038 #5082) — this **exercises existing production routing**, it does not change it. Register two allowlist-valid slugs A and B before a single restart; drive uniquely-marked writes through slug A's real routes through the genuine `parse_project_key → resolve_store → dispatch_request` funnel, then do a **genuine two-store content read**.

Per the human Q1 decision (**option a**), infra-003 covers **both** served write surfaces:

1. **Observe surface** — `POST /v1/A/observe`.
2. **HTTP MCP-write surface** — `POST /v1/A/mcp` (`context_store`/`context_correct`). This is load-bearing: the `entry.store == adapter-store` equality is guarded only by a `debug_assert!` that is **compiled out of the release container**, so MCP-write isolation has **zero behavioral coverage in the shipped artifact**.

For each surface:
- **Positive control (load-bearing):** assert A's store grew **with the marker** via a genuine content read (not a `du` size delta).
- **Negative control:** assert B's store did **not** receive A's marker — `landed_only_in_a` — via a genuine two-store read (no dir-count / `other_count` heuristic).
- The positive control **gates** the negative control: a silently-failed A-write fails RED, never passes vacuously as "B unchanged."

**Local-UDS side:** the ADR-006 compile-time guard (`local_binding_guard_tests.rs::FORBIDDEN_IN_LOCAL`) proves a second local route is *unrepresentable* — referenced, not re-run. No UDS behavioral probe.

[...]

Issue #844, opened 2026-06-26
test(infra-001/nan-022): D1/D4 parity flip — fix corpus under-population (5→~25) + stable-prefix assertion; residual is intrinsic HNSW build-RNG (#852)


dug-21:
## Summary (reframed per ass-085 #852 empirical findings — supersedes the wide-margin framing)

The nan-022 parity matrix intermittently classifies **retrieval (D1)** and **briefing (D4)** as
`PARITY_FAIL`. Root cause is **not** a transport defect, and — per the ass-085 measurement
(#852, 300–500 iterations/candidate) — it is **not** primarily a similarity-margin problem
either. There are two distinct causes:

1. **Under-population (the dominant, fixable cause).** The parity corpus is
   `SEED_CORPUS_SIZE = 5` while retrieval asks for top-5. The index returns **fewer than 5
   results in 11–25% of builds** — and a short return is the literal #844 signature
   (`[3,1,5,4]` = only 4 returned). This is the primary driver of the observed failures.
2. **Intrinsic HNSW build randomness (the irreducible residual).** `hnsw_rs` 0.3.x has **no
   seeding API**; the graph is built from per-process entropy, so ~**0.2–0.6%** of runs reorder
   the top-k regardless of corpus design. This is the #746 / #4990 root cause and it is a
   **graph-construction** property — it survives near-exhaustive search and appears even inside
   the top-3 prefix.

## What ass-085 disproved

- **"Wide inter-entry similarity separation eliminates the flip"** — **false**. Margin is
  non-monotonic (wider often flips *more*), is geometrically capped (text embeddings can't be
  pushed arbitrarily far apart), and the residual is build-time, not boundary-driven.
- **"Exact top-k, no tolerance band"** (the framing this issue previously carried) — **not
  achievable**. No corpus/margin/breadth lever reaches a hard zero at the ≥150x sampling bar.
  Exact-zero parity for retrieval ordering is unreachable with `hnsw_rs` as shipped.

## Evidence

Failing-run query[0] (`"nan-022-parity-corpus cross-transport parity"`, k=5):
HTTPS `[3, 1, 5, 4]` vs UDS `[3, 1, 2, 5, 4]` — entry 2 on the cutoff; both legs intra-stable
per run, flip across runs; D4 tracks D1. Measured flip rates (500-iter confirmation): current
N=5 corpus ≈ **13.8% (all short-returns)**; N=25 + 0.20 moat ≈ **0.4–0.6%** (~30× better, zero
[...]

PR #854, opened 2026-06-26
fix: remove D6 isolation from nan-022 parity matrix — isolation is not a parity property (#845)

### LINKED ISSUE #845: test(infra-001/nan-022): remove D6 isolation from the parity matrix — isolation is not a parity property (genuine test → #853)
## Disposition (D1 — supersedes ALL prior framing in this issue, including the asymmetric-reshape + K4-attestation proposal in the comments below)

nan-022/C0 is a **parity** feature: it proves HTTPS behaves identically to local-UDS for the same workload. **Cross-tenant isolation is not a parity property** — it is an HTTP-only function with no local analog (local is 1-client:1-project by the ADR-006 compile-time guard, so there is no local tenant to be at parity with). D6 was a **categorization error** in the original nan-022 scoping. Every prior attempt in this issue — "make the UDS leg two-slug", then "asymmetric comparator + K4 attestation" — was engineering *around* that error rather than fixing it.

Per the uni-zero D1/D2 decision: **isolation leaves the parity matrix entirely**, and the genuine isolation test is built standalone under N3 in **#853**.

## Scope (this issue) — a REMOVAL. Test-only, cumulative on infra-001.

Remove the `isolation` (D6) dimension from the nan-022 parity suite:
- `harness/parity_dimensions.py`: drop `isolation` from `DIMENSIONS` and from `DUAL_SURFACE_DIMENSIONS`.
- `harness/parity_comparator.py`: remove `IsolationComparator`; keep `assert_comparator_contract` consistent with the reduced `DIMENSIONS`.
- Remove the `isolation` key from the `dimension_bundle` schema; update the C5' never-empty node guard (#5321) to drop its isolation clause (so the bundle write neither hard-fails nor silently mis-passes).
- Remove both non-functional probes: the UDS `feature=`-hint probe (`harness/parity_legs_capture.py::capture_isolation`) and the HTTPS `capture_isolation_probe` (`scripts/cloud-bundle-lib.sh`, also a false-RED trap).
- Update the suites that encode isolation as a dimension: `suites/test_parity_{comparator,dimensions,legs,outcome}.py`, `suites/test_transport_health.py`.

**The parity bar becomes 5 dimensions:** retrieval, behavioral, analytics, proactive, precompact.

This is *cleaner* than the asymmetric reshape proposed in the comments — there is nothing to reshape; isolation is removed. **Design-review F1 (attestation-via-K4) is therefore moot.**

## NOT in this issue

The genuine multi-tenant HTTP isolation test (write to slug A lands only in A's store) moves to **#853** (standalone, N3 #5161, outside the parity harness). This issue only removes the non-parity slot.

## Why this stops the recurrence

The investigation, architecture review, and product review (comments below) each did correct work *inside the wrong frame* — refining how to fit isolation into a parity comparator. The frame was the error. Removing isolation from parity fixes it at the source.

## Capability / C0

C0 (#5304→#5324) parity bar corrected to **5 dimensions**; isolation excluded as non-parity, owned by N3 (#853). The C0 flip no longer waits on any isolation test.

dug-21:
## Summary
Removes the **D6 `isolation`** dimension from the nan-022 parity matrix. Per uni-zero D1, cross-tenant isolation is **not a parity property** — parity proves HTTPS behaves identically to local-UDS for the same workload, but local is 1-client:1-project by the ADR-006 compile-time guard, so there is no local tenant to be at parity with. D6 was a categorization error; both prior probes were non-functional (UDS `feature=`-hint always self-found the marker; HTTPS dir-count was vacuous + a false-RED trap).

This is a **pure removal**, test-only, cumulative on infra-001. The parity bar becomes **5 dimensions**: retrieval, behavioral, analytics, proactive, precompact. The genuine multi-tenant HTTP isolation test (write to slug A lands only in A's store) is **out of scope** and owned by **#853** (N3 #5161).

## Changes (19 files, all under `product/test/infra-001/`)
- Dropped `isolation` from `DIMENSIONS` + `DUAL_SURFACE_DIMENSIONS`; removed `IsolationComparator` **and** its `bind_comparators` entry in lockstep (avoids import-time KeyError).
- Removed both probes (`capture_isolation`, `capture_isolation_probe`); dropped the C5' never-empty guard clause + assembly key.
- Synced the second key-list mirror (`transport_health._CANONICAL_CAPTURE_KEYS`), both release-gate `want`-lists (6→5), and removed dead `if dim_id=="isolation"` branches in `parity_matrix_support`/`parity_outcome`.
- Re-pointed the two `del bundle[...]` missing-key→INFRA probes (`isolation`→`behavioral`) so they still exercise the path; re-baselined call-count assertions.

## Verification
- 7 affected parity suites: **187 passed**
- Release-gate logic tests: **10 + 21 passed**
- Mandatory smoke: **24 passed**
- Live Docker HTTPS/UDS parity: matrix emits **exactly 5 dimensions, no isolation**; 4× PARITY_PASS
- The lone Docker `verdict=ERROR` is the **pre-existing** D5/precompact host-side measurability gap (#844), not caused by this removal — this fix removes a former ERROR contributor (strict improvement).
- Test-infra only; zero Rust touched.

## Gate
Bug Fix Validation — **PASS** (7/7).

Closes #845.


dug-21:

**Risk: LOW. No blocking findings.** Test-only, pure removal; cumulative on infra-001. Zero Rust/JS production code touched. Full `main...HEAD` diff read cold (19 files, +108/-344).

### Real security signal NOT weakened (the key check)
The removed D6 probes were genuinely **vacuous**, confirmed by reading the deleted code:
- **UDS `capture_isolation`** searched slug B with a `feature=`-hint that, under the ADR-006 1-client:1-project local guard, has no second tenant store to scope to — it self-found / always saw the healthy case.
- **HTTPS `capture_isolation_probe`** never performed a genuine cross-slug read; it counted "other" per-slug dirs (`other_count`). With no other dir it asserted confinement by *absence*; with a stray dir from a prior run it flipped `visible_to_b=true` → a **false-RED trap**, not a real leak detector.

Removing these does not drop any real production parity invariant — isolation is an HTTP-only property with no local-UDS analog, so it was never a parity property. The genuine multi-tenant test (two-store read, positive+negative control) is **explicitly tracked in #853 (OPEN, N3 #5161 partial)** — not silently dropped. Verified the issue exists and scopes the real work.

### False-green / dangling-reference audit (passed)
- Grepped the whole `product/test/infra-001/` tree: **no surviving** `IsolationComparator`, `capture_isolation`, `slug_a_writes_visible_to_b`, or `landed_only_in_a` references. Remaining "isolation" hits are unrelated (test state-leakage NFR-01, HOME process-isolation, out-of-scope M7 notes).
- All single-source enumerations updated **in lockstep** 6→5: `DIMENSIONS`, `DUAL_SURFACE_DIMENSIONS`, `_CANONICAL_CAPTURE_KEYS`, `DRIVER_WIRE_SURFACES`, and **both** release-gate `want`-lists (bundle-assembly + cloud-cycle).
- The isolation never-empty guard was removed **together with** its bundle key — no orphan guard (would false-RED) and no key emitted without a guard (would false-pass).
- The missing-capture-key → INFRA negative tests were **re-pointed** isolation→behavioral (`test_matrix_missing_capture_key_in_bundle_errors`, `test_load_https_bundle_missing_capture_key_raises_infra`, `_StubDim`), so the "absent dimension errors, never empty-passes" regression guard is **preserved**, not deleted.
- `bridge-cycle-capture.js` (not in diff) carries no isolation refs; no `parity_bundle_contract.md` / architecture schema left stale.

### OWASP / blast radius
- **Injection:** net **reduction** — the deleted shell `capture_isolation_probe` interpolated `$SLUG` into a `vol sh -c '…'` command; removing it deletes that shell-interpolation surface. The deleted Python SQL was already parameterized (`WHERE topic_signal = ?`). No new interpolation introduced.
- **Deserialization / input validation:** unchanged; `JSON.parse` paths operate only on test-internal synthesized fixtures.
- **Dependencies:** none added. **Secrets:** none introduced.
- **Worst case if subtly wrong:** a parity gate flips green with 5 dims — but the negative INFRA-path tests and lockstep want-lists make a silently-missing dimension fail loudly, so the dangerous "silently stops testing" mode is guarded against.

### Non-blocking finding — exec bit lost (755→644)
`scripts/cloud-bundle-lib.sh` and `scripts/release-gate-bundle-assembly-logic-test.sh` lost the exec bit.
- **No functional/pipeline/security impact today:** `cloud-bundle-lib.sh` is **sourced** (`. "$SCRIPTS_DIR/cloud-bundle-lib.sh"`) and the assembly logic test is run via `bash <script>` (matching the `run_smoke_gate … bash …` pattern in release.yml) — both ignore the exec bit. Verified the gate logic ran (PR: 10+21 passed).
- **Recommendation (hygiene):** restore `chmod 755` for consistency with sibling scripts and to future-proof against a `./script` invocation: `git update-index --chmod=+x …` or `chmod +x … && git add`.

**Disposition: APPROVE from a security standpoint.** Single nit (exec bit) is non-blocking.

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. Test-only pure removal, independently verified clean, security APPROVE; the one security nit is already fixed on the branch.

### Vision / roadmap fit
Advances `goal:personal-cloud` and its marquee capability **C0** (full HTTPS==local intelligence-pipeline parity). This PR is the code-side execution of a decision the corpus already records: the active C0 entry **#5330** carries the corrected **5-dimension** parity bar (retrieval, behavioral, analytics, proactive, precompact) with per-slug isolation `excluded-by-design` — removed (#845), no local analog under the ADR-006 1-client:1-project guard, genuine test owned by N3/#853. Corpus and code converge exactly; nothing here contradicts the map. The net effect is a strict improvement toward an *honest* parity bar — it deletes a false-RED contributor that was muddying C0-flip readiness, rather than narrowing the bar.

### Approach commentary
- **Removal (D1) over the earlier reshape/K4-attestation is the right call** and the issue documents why cleanly: the prior attempts were faithful work inside the wrong frame. Fixing the categorization at source is less code and less risk.
- **The real risk in a removal like this is lockstep** — a dangling reference that leaves a gate silently false-green, an orphaned guard that false-REDs, or a negative test that quietly becomes a no-op. Every reviewer independently confirmed the dangerous modes are closed: registry + comparator + `bind_comparators` + both probes + the C5' never-empty guard + both key-list mirrors + both release-gate `want`-lists all moved 6→5 together, and the missing-key→INFRA negative probes were **re-pointed isolation→behavioral (not deleted)**, so the "absent dimension errors loudly, never empty-passes" regression guard survives. That is the load-bearing property and it is covered.
- **Security review is independent and clean** (LOW, APPROVE) — it read the deleted code cold, confirmed both probes were genuinely vacuous (UDS self-found the marker; HTTPS counted dirs = a false-RED trap), and noted a net injection-surface *reduction* (deleted shell `$SLUG` interpolation). Its sole nit (exec bit 755→644 on `cloud-bundle-lib.sh` + `release-gate-bundle-assembly-logic-test.sh`) is **already resolved** by branch commit `04863303` — both files are `100755` at HEAD. No open security action remains.

### Capability coverage (delivery assessment — C0)
Target: **C0 #5330 — Full intelligence-pipeline fidelity over HTTPS == local**   Coverage: this PR delivers a corrective slice, not a flip.
- 5-dim parity bar (isolation removed) → covered by this PR + already reflected in #5330.
- Isolation as a genuine property → DEFERRED (named gap) to N3/#853 (OPEN) — correctly decoupled from the flip.
Verdict: meets the defined part. No undeclared gap. This PR is **test-only and does NOT flip C0** (it does not claim to) — C0 stays `partial`, correctly blocked on #844 (D1/D4 retrieval HNSW-boundary decision) and the D5 host-side documented exception. No capability-map mutation is needed; #5330 already encodes this outcome.

### Note — PR file list looks larger than the change (benign)
GitHub's file list / `gh pr diff` shows **22 files**, including 3 `.claude/` protocol+agent docs — which is why the body's "19 files, all under `product/test/infra-001/`, zero Rust" can look inconsistent. It is not: `main` is an ancestor of the branch, the branch-only commits touch **exactly 19 files all under `product/test/infra-001/`**, and the 3 `.claude/` files are byte-identical to `main` (already landed in 45191aff). They are a stale-PR-base rendering artifact and a **merge no-op** — not scope creep, not production code. The body's scope statement is accurate as to merge effect.

No open questions remain at this gate. Disposition D1 is decided (uni-zero D1/D2, reflected in #5330); security review has no blocking findings; the exec-bit nit is already fixed.

### Recommended actions
1. **Merge as-is.** Pure removal, lockstep verified by fix/verify/gate, live Docker emits exactly 5 dimensions, security APPROVE, nit resolved.
2. Disregard the apparent 3-extra-files in the GitHub diff (stale-base no-op, explained above).
3. **Keep #853 on the critical path** — this PR removes a *false* isolation signal; the only genuine multi-tenant HTTP isolation proof is still OPEN on an unrollbackable failure mode. Removing the false signal must not lower the urgency of the real test.
4. C0 flip stays correctly gated on #844 (D1/D4) + the D5 host-side exception; the capability map (#5330) is already current — no further map action.

Issue #849, opened 2026-06-26
release lane: provision sqlite3 for nan-022 live parity matrix (D2 behavioral read) before first tag round


dug-21:
## Problem

The nan-022 cross-transport parity matrix (`pytest -m parity`, release-gate Docker lane) performs the **D2 behavioral** capture by reading the per-slug DB host-side via `sqlite3` (the C5 shell smoke gate, `cloud-bundle-lib.sh`). By design this is a **HARD INFRA fail** (distinct exit, never a vacuous/empty pass) when `sqlite3` is absent — so on a clean CI runner the first live tag round will error out as INFRA-ERROR rather than producing a parity verdict.

`sqlite3` is **not guaranteed** on the release-lane runner / distroless image (it is not on the shipped image, and `node` is provisioned explicitly). It happened to be installed ad-hoc in the local dev container during the nan-022 Stage 3c first live run; CI has no such guarantee.

## Impact

The first live `pytest -m parity` tag round will burn on a mis-provisioned lane (INFRA-ERROR), not on a real parity result — wasting a tag round and masking the actual matrix outcome until the lane is fixed.

## Fix options

1. **Provision `sqlite3` in the release-gate lane** alongside `node` (recommended — minimal, keeps the host-side DB read).
2. Alternatively, route the D2 behavioral read through the MCP bridge / analytics surface instead of a host-side `sqlite3` query (larger change; reconsider only if provisioning is undesirable).

## Context

- Surfaced at nan-022 Stage 3c first-live-run (PR #846, merged `6a8282d3`); flagged by the C5 implementer and the security review as a non-blocking ops note.
- The HARD-INFRA-on-missing-`sqlite3` behavior is correct and should stay — this issue is about lane provisioning, not changing the guard.
- Files: `product/test/infra-001/scripts/cloud-bundle-lib.sh` (the D2 read), release lane job in `.github/workflows/release.yml`.

## Acceptance

The release-gate lane that runs `pytest -m parity` has `sqlite3` available, so a clean tag round reaches a real per-dimension parity verdict instead of an INFRA-ERROR on a missing binary.

dug-21:
# Bug Investigation Report: 849-investigator

## Bug Summary
[...]

Issue #837, opened 2026-06-24
feat(personal-cloud): cross-transport parity suite — prove the full intelligence pipeline behaves identically over HTTPS and stdio/UDS (C0)


dug-21:
**[nan-022]** Cross-transport parity suite — C0's (#5304) proof artifact. TEST-ONLY.

Design complete (Session 1). This issue tracks the implementation.

## Summary

Generalize nan-021's single-`MetricVector` HTTPS-vs-UDS parity gate (`product/test/infra-001/`) into a **dimension-keyed parity matrix** that drives ONE canonical workload over BOTH transports (HTTPS bridge / stdio-UDS) in ONE pytest invocation and asserts *measured* parity across all six C0 (#5304) dimensions: **retrieval, behavioral signals, analytics/learning, proactive delivery, PreCompact restoration, per-slug isolation**. nan-021 proved exactly one (analytics/learning); this proves the remaining five and composes all six into one false-green-proof release-gate matrix. Passing the full matrix is the behavioral evidence an authorized session uses to flip C0 → `proven` — this feature does NOT itself flip C0.

## Approach

One authoritative dimension registry (`DIMENSIONS`); leg drivers return a dimension-keyed bundle; the orchestrator runs one comparator per dimension and emits a per-dimension PASS/FAIL evidence table keyed by the run-correlation token. Analytics is CONSUMED verbatim, not re-proven. Net-new (pure-Python, off-Docker unit-testable): comparator framework + cross-dimension drift guard, one ranking-tolerance policy (retrieval + briefing), a four-valued outcome-class model, and a transport-health preflight.

## Scope / Non-Goals

- **No production code.** Test-only, cumulative on `infra-001`; no fork, no parallel scaffold. Any defect surfaced is FILED as a GH bug, not fixed here.
- Not a re-prove of nan-021's analytics slice (consumed verbatim). Not a soak/load test. Not a Claude-Code-driven integration. Does not broaden C0's surface beyond the six named dimensions. Does not itself flip C0. Not wired into the JS-only `ci.yml` PR matrix (release-gate Docker lane only).

## Dimensions × outcome classes

[...]

PR #860, opened 2026-06-28
fix: provision UDS-leg binary for nan-021 parity gate (#851)

### LINKED ISSUE #851: release.yml nan-021-https-uds-parity gate fails: pytest conftest requires local unimatrix binary not provisioned in release job
## Summary
The `nan-021-https-uds-parity` gate in the Release workflow fails at every release run. **Environmental / CI-wiring bug** — not a parity regression and not a bad release. v0.8.9 published correctly (npm, container manifest, GitHub release all green); this gate is intentionally non-blocking ([release.yml:560-572](../blob/main/.github/workflows/release.yml#L560-L572), kept out of `create-container-manifest` `needs:` until first-green).

## Symptom
Run [28234715166](https://github.com/dug-21/unimatrix/actions/runs/28234715166) (v0.8.9), job `nan-021-https-uds-parity`:

```
RuntimeError: Cannot find unimatrix binary.
Set UNIMATRIX_BINARY env var or build with: cargo build --release
harness/conftest.py:51
→ 10 deselected, 3 errors in 0.11s
```

All three parity tests error at fixture setup — collection-time, before any parity logic executes.

## Root cause
The job (`release.yml:573`) provisions Node, Python, GHCR login — but **never builds the Rust binary nor exports `UNIMATRIX_BINARY`**. The harness `conftest.py:_resolve_binary()` unconditionally looks for `target/release/unimatrix` / `target/debug/unimatrix` (or the env var) at fixture setup and aborts when none exist.

The gate's *intent* is to test against the container `IMAGE` (HTTPS-vs-UDS parity), so the conftest's hard requirement for a **local** binary conflicts with the IMAGE-based design. The local-binary requirement is satisfied in PR CI (where the binary is built) but not in the release job.

## Fix options
1. Provision the binary in the release job — download the existing build artifact (`unimatrix-linux-x64` is already produced upstream) and `export UNIMATRIX_BINARY`, rather than rebuilding.
2. Or make `conftest._resolve_binary()` lazy / skip the local-binary requirement for the IMAGE-based parity path so it doesn't error at collection when the gate targets a container.

Option 1 (reuse the existing artifact) is cheapest — avoids a redundant `cargo build --release` in the release pipeline.

## Impact
None on releases — gate is non-blocking by design. It just keeps the gate red and blocks "first-green" promotion into the manifest `needs:`.

dug-21:
## Summary
Fixes #851 — the `nan-021-https-uds-parity` gate in the Release workflow errored at pytest collection on every release run (`RuntimeError: Cannot find unimatrix binary`).

## Root cause
The gate is **cross-transport**, not pure-IMAGE. The HTTPS leg runs against the container image, but the **UDS leg spawns a live local daemon** via the `daemon_server` fixture, so a host-compatible local binary is a genuine requirement. The parity job's `needs:` only guaranteed the container image; the host-built `unimatrix-linux-x64` artifact (from `build-linux-x64`) was neither in `needs:` nor produced on `workflow_dispatch`, so `conftest._resolve_binary()` aborted at setup.

## Fix (full — both tag-push and workflow_dispatch)
- **`.github/workflows/release.yml`**
  - `nan-021-https-uds-parity`: add `build-linux-x64` to `needs:`; download `unimatrix-linux-x64` to `${{ github.workspace }}/bin`; `chmod +x`; export `UNIMATRIX_BINARY` via `$GITHUB_ENV`; `model-download` warm-up before pytest so the UDS daemon binds within the 15s deadline. ORT resolves via co-located lib (`RUNPATH $ORIGIN`).
  - `build-linux-x64`: relax the `workflow_dispatch`-excluding `if:` so the host artifact is produced on dispatch too. Manifest/npm/release/arm64 guards untouched.
- **`product/test/infra-001/scripts/release-gate-cloud-cycle-logic-test.sh`**: two new static assertions — `test_c5_lane_provisions_uds_leg_binary` and `test_c5_build_linux_x64_runs_on_dispatch`.

## Verification
- New static assertions: pre-fix FAIL (21/2) → post-fix PASS (23/0) — reproduces the gap, verifies the fix.
- `cargo test --workspace` rc=0; clippy clean; integration smoke 24 passed / 0 failed; release.yml parses.
- The live parity job cannot run locally (needs container image + runner artifact); the static gate test is the design-intended proxy.

## Scope / notes
Gate is non-blocking by design; this fix cannot break a real release (all manifest/publish edges untouched). Clears the binary-provisioning blocker — coordinate with the #849/#844 first-green sweep for any next sequenced failure.

Diagnosis, design review, product review, fix, test, and gate reports are on #851.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,8 +18,12 @@ env:
   ORT_VERSION: "1.20.1"
 
 jobs:
+  # NOTE: runs on BOTH tag-push and workflow_dispatch. The dispatch path is
+  # load-bearing for the nan-021-https-uds-parity gate, whose UDS leg spawns a
+  # live LOCAL daemon and needs this host-built `unimatrix-linux-x64` artifact
+  # (#851). The only other consumer, package-npm, stays dispatch-guarded, so
+  # producing the artifact on dispatch causes no extra publish.
   build-linux-x64:
-    if: github.event_name != 'workflow_dispatch'
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
@@ -571,10 +575,34 @@ jobs:
   # by release-gate-cloud-cycle-logic-test.sh (R-12) so this is not the first
   # place the gate logic runs.
   nan-021-https-uds-parity:
-    needs: [build-container-x64]
+    needs: [build-container-x64, build-linux-x64]
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
+
+      # The UDS leg spawns a live LOCAL daemon (daemon_server fixture), which
+      # needs a HOST-compatible binary. Use the ubuntu-22.04-built artifact
+      # (glibc-compatible with this runner) — NOT a binary extracted from the
+      # bookworm/distroless image. Download to an absolute, working-directory-
+      # independent path (download-artifact's `path:` is repo-root relative and
+      # ignores the job-level working-directory) (#851, F2).
+      - name: Download host-built UDS-leg binary
+        uses: actions/download-artifact@v4
+        with:
+          name: unimatrix-linux-x64
+          path: ${{ github.workspace }}/bin
+
+      # chmod (upload-artifact drops the exec bit) + export UNIMATRIX_BINARY via
+      # $GITHUB_ENV so the pytest step's daemon_server/binary_version_preflight
+      # resolve it. Warm the embedding model up front (model-download) so the
+      # cold first-boot HF fetch does not blow the 15s daemon socket-bind
+      # deadline; ORT resolves via the co-located lib (RUNPATH $ORIGIN) (#851, F1).
+      - name: Provision UDS-leg binary and warm embedding model
+        run: |
+          set -euo pipefail
+          chmod +x "${{ github.workspace }}/bin/unimatrix"
+          echo "UNIMATRIX_BINARY=${{ github.workspace }}/bin/unimatrix" >> "$GITHUB_ENV"
+          "${{ github.workspace }}/bin/unimatrix" model-download
 
       - name: Provision pinned Node for the shipped mcp-bridge.js leg
         uses: actions/setup-node@v4
--- a/product/test/infra-001/scripts/release-gate-cloud-cycle-logic-test.sh
+++ b/product/test/infra-001/scripts/release-gate-cloud-cycle-logic-test.sh
@@ -491,6 +491,43 @@ test_c5_lane_resolves_image_via_shipped_lib() {
 }
 test_c5_lane_resolves_image_via_shipped_lib
 
+test_c5_lane_provisions_uds_leg_binary() {
+  # The parity gate is CROSS-TRANSPORT: the UDS leg spawns a live LOCAL daemon
+  # (daemon_server) that needs a HOST-compatible binary (#851). Assert the lane
+  # provisions it: (a) downloads the unimatrix-linux-x64 artifact, (b) exports
+  # UNIMATRIX_BINARY, (c) has build-linux-x64 in `needs:`, (f1) warms the model
+  # so the daemon binds within the 15s socket deadline. (d) below asserts the
+  # build job actually produces the artifact on dispatch — a `needs:` edge on a
+  # guarded-skipped job would SILENTLY SKIP the parity lane on workflow_dispatch.
+  local blk
+  blk="$(job_block nan-021-https-uds-parity)"
+  if printf '%s\n' "$blk" | grep -q 'download-artifact' \
+     && printf '%s\n' "$blk" | grep -q 'name: unimatrix-linux-x64' \
+     && printf '%s\n' "$blk" | grep -q 'UNIMATRIX_BINARY' \
+     && printf '%s\n' "$blk" | grep -q 'model-download' \
+     && printf '%s\n' "$blk" | grep -qE 'needs:.*build-linux-x64'; then
+    pass "test_c5_lane_provisions_uds_leg_binary (UDS-leg binary downloaded + UNIMATRIX_BINARY exported + model warmed + needs: build-linux-x64)"
+  else
+    oops "test_c5_lane_provisions_uds_leg_binary" "parity lane does not provision the UDS-leg binary (download unimatrix-linux-x64 + export UNIMATRIX_BINARY + model-download + needs: build-linux-x64)"
+  fi
+}
+test_c5_lane_provisions_uds_leg_binary
+
+test_c5_build_linux_x64_runs_on_dispatch() {
+  # (d) The artifact must EXIST on workflow_dispatch, not just tag-push. A job
+  # whose `needs:` dependency is guard-SKIPPED is itself skipped (not failed),
+  # so a dispatch-excluding `if:` on build-linux-x64 would silently dark the
+  # parity gate on dispatch (#851 / F3).
+  local blk
+  blk="$(job_block build-linux-x64)"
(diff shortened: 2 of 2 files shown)


dug-21 (commented):
## Security Review — bugfix-851-security-reviewer (fresh context)

**Risk: LOW. No blocking findings.** CI/release-workflow-scoped change (release.yml + one gate-logic test). No Rust/JS source touched.

### Supply-chain / artifact trust — OK
The parity job downloads `unimatrix-linux-x64` via `actions/download-artifact@v4` with no `run-id`/`github-token`, so it is **scoped to the current workflow run** (v4 default). It is a first-party, same-run, same-ref artifact produced by `build-linux-x64` in this very run — the identical trust model `package-npm` already uses for the same artifact (lines 225-234). It never crosses the run boundary, so no attacker-controllable/cross-run substitution is possible and a sha256 pin would add no integrity guarantee here. `chmod +x` + executing the binary (`model-download`) is executing the first-party artifact under release — expected.

### Privilege / token surface — OK
Relaxing only `build-linux-x64`'s `workflow_dispatch` guard exposes no new secret to a new path: that job pushes nothing (checkout/build/test/upload-artifact only) and uses no OIDC/GHCR/npm token. Confirmed every publish edge stays guarded:
- `package-npm` keeps `if: != workflow_dispatch` (line 211) — npm `--provenance`/id-token never runs on dispatch.
- `create-release` is `needs: package-npm` — transitively skipped on dispatch.
- `create-container-manifest` keeps its guard (line 696).
- `build-linux-arm64` guard untouched (line 118).
So dispatch still cannot trigger any publish/manifest/release step. The only added dispatch behavior is producing a retention-1d artifact + running the (non-publishing) parity gate.

### Injection — OK
New `run:` blocks interpolate only `${{ github.workspace }}` (runner-controlled, not attacker input) and write `$GITHUB_ENV`. Workflow triggers are `push: tags v*` and `workflow_dispatch` — both require write access; no fork-PR / issue-body / branch-name interpolation. `set -euo pipefail` + quoted paths throughout.

### Blast radius — contained
`nan-021-https-uds-parity` is a **leaf** job — no job lists it in `needs:` (verified). It is intentionally NOT in `create-container-manifest`'s `needs:` (first-green budget). Worst case of subtly-wrong wiring = the non-blocking parity gate (or `build-linux-x64` on a manual dispatch) goes red. It **cannot break or publish a bad real release**: the tag-push publish path is byte-for-byte unchanged and structurally decoupled from this job.

### Secrets — none
No credentials/tokens/keys in the diff.

### Minimal-change — clean
Change set is confined to the bug: one `if:` removed, `needs:` extended, two provisioning steps, two static gate assertions. No unrelated edits.

Minor (non-blocking): the parity gate's `model-download` adds a HuggingFace network dependency at gate time; an HF outage reddens the gate. Acceptable — non-blocking gate, and `build-linux-x64` already performs the same fetch.

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready as a first-green blocker-clearing fix. Diagnosis, fix, and tests are sound and the blast-radius claims verify against ground truth. Two things to accept consciously before merge: the F4 dispatch trade-off, and confirm the security verdict is captured (no dedicated security-review comment is in the inputs).

### Vision / roadmap fit
This is on the `personal-cloud` goal's critical path. `nan-021-https-uds-parity` is the behavioral proof that the intelligence pipeline behaves identically over HTTPS and UDS (capability C0 / #837) — the load-bearing evidence for "full pipeline fidelity over HTTPS." The lane could never green because the UDS leg's live local daemon had no binary, so C0 cannot flip to proven and the gate cannot be promoted into `create-container-manifest`'s `needs:`. This PR removes that blocker. It does NOT by itself green the lane — per the ADR-005 first-green budget it will likely surface the next sequenced failure (cert/bridge/cycle/review/parity). Treat this as one step in the coordinated first-green sweep with #849/#844, not a standalone green. The PR body and every gate report frame the success criterion correctly ("clears the binary-provisioning blocker"), which is the right altitude.

### Approach commentary
- Option 1 (provision the host-built `unimatrix-linux-x64` artifact) is the correct call; image-extraction (glibc 2.36 vs runner 2.35) and lazy-skip-conftest were correctly rejected. The UDS leg genuinely needs a host-compatible local daemon — the binary is a real requirement, not an accident.
- The architect amendments all landed: F2 absolute `${{ github.workspace }}/bin` path (avoids the `working-directory` mismatch), F1 `model-download` warm-up (clears the 15s socket-bind deadline), F3 guard relaxation + the (d) static assertion that catches the silent-skip-on-dispatch trap. Good closure.
- Blast radius verified against current main: `package-npm` keeps `needs: [build-linux-x64, build-linux-arm64]` + `if: != workflow_dispatch`, `build-linux-arm64` stays guarded, and `create-container-manifest.needs` excludes both parity and `build-linux-x64`. Relaxing only the x64 guard produces no new publish on dispatch, and a wrong/failed parity job cannot block a release. The "cannot break a real release" claim holds.

### Open items the human should weigh
**1. F4 trade-off (accepted in design review, worth a conscious nod at merge).** Removing `build-linux-x64`'s dispatch guard drags its full `cargo test --release` (release.yml:97) + `model-download` onto every `workflow_dispatch`, not just the artifact build. Two consequences on dispatch only: (a) the critical path to the parity gate lengthens to `max(container build, linux-x64 build+test)`; (b) a `cargo test` flake fails `build-linux-x64` → the parity job is SKIPPED (not failed) → parity signal silently lost on that dispatch run. On tag this is immaterial (a `build-linux-x64` failure already blocks via `package-npm`). Given the gate is non-blocking and first-green-budgeted, the simpler relax-the-guard form is acceptable — but it is a real behavioral change to `workflow_dispatch`. The cleaner long-term shape (a lean build+bundle-artifact-only job, heavy test left tag-only) is a larger refactor; fine to defer as a tracked follow-up rather than expand this PR.

**2. Security verdict not in the inputs.** No dedicated security-review comment is present on the issue or PR. The available security analysis is design-review F6: the binary is built from the same checked-out ref and consumed within the same run (download-artifact@v4 is run-scoped, retention 1 day), so there is no new external trust boundary — unlike the ORT `curl|tar` that #4274 flagged for a missing sha256. For a YAML-only, same-run first-party-artifact change that conclusion is appropriate and no sha256 gate is warranted. Recommend the human confirm whether the protocol expects a standalone security-reviewer verdict here; if so, ensure it is captured before merge. I do not re-run it.

### Right-sizing
The fix is scoped to the diagnosed root cause — no gold-plating, no under-build. It delivers the OUTCOME (the parity gate can actually run its UDS leg on both triggers), not just a mechanism, and is honest that greening the lane is a separate downstream step. The two new static assertions extend the existing C5 lane block (cumulative test-infra idiom, #5192) rather than scaffolding new infra, and they reproduce the gap (pre-fix 21/2 FAIL → post-fix 23/0 PASS).

### Recommended actions
1. Merge as-is once items 1-2 are consciously accepted — the fix is narrow, verified, and cannot break a release.
2. Set team expectation: this clears the binary blocker; the lane greens over the shared #849/#844 first-green sweep, expect the next sequenced failure to surface on the next real run (by design, ADR-005).
3. Optionally file a follow-up for the F4 lean-artifact-job refactor if dispatch-run cost / parity-signal coupling becomes a concern.
4. Capability map: leave C0 (#837) `partial` — this PR enables the proof path but does not yet produce a green parity run. Capability-status mutation is the vision session's call, not mine.

_Independence: spawn prompt was clean (ID, gate, issue, PR, inputs only) — no contract violation. Capability-coverage / right-sizing-by-clause sections are scope/design-review concerns; at pr-review I assess merge-readiness of the whole._

PR #873, opened 2026-07-02
fix: guard context_get/context_graph tool-description literal drift (#869)

### LINKED ISSUE #869: context_get: guard CONTEXT_GET_DESCRIPTION const against tool-description drift (vnc-042 INFO-1)
Non-blocking finding from the vnc-042 security review (#843, PR #867).

## Problem
`CONTEXT_GET_DESCRIPTION` is `#[allow(dead_code)]` and its test asserts only **substrings**, not byte-equality with the live `#[tool(description = ...)]` literal. The two match today but can silently drift. Its doc-comment is also a copy-paste leftover referencing `context_graph`.

## Task
- Add a byte-equality assertion between `CONTEXT_GET_DESCRIPTION` and the live `#[tool(description=...)]` literal (or wire the const into the attribute so there is a single source of truth).
- Fix the stale `context_graph` reference in the const's doc-comment.

## Acceptance
- A test fails if the tool description literal and the const diverge by a single byte.

Ref: product/features/vnc-042/agents/vnc-042-security-reviewer-report.md (INFO-1)

dug-21:
Fixes #869.

## Root cause
The client-facing tool description for `context_get` existed as **two independent, unlinked literals** for one logical string: the live `#[tool(description = "...")]` attribute on the handler and a mirror `const`. They were byte-identical but nothing enforced it — the existing test asserted only *substrings against the const*, never the live macro-generated literal, so the description could drift silently. `context_graph` carried the identical latent twin, and its test comment **falsely claimed** to be the "single source of truth."

## Fix
Structurally enforce the invariant by making it CI-checked (the rmcp 1.7.0 `#[tool]` macro accepts only a string literal in `description =`, so a shared const cannot be single-sourced; the byte-equality guard is the enforceable link):

- Add byte-equality guard tests comparing the **live** `UnimatrixServer::{context_get,context_graph}_tool_attr().description` against the mirror const — fails on a single-byte divergence (the acceptance criterion).
- Fix the stale `context_get` doc-comment (copy-paste leftover referencing `context_graph`).
- Correct the false "single source of truth" comment on the `context_graph` substring test.
- Keep the existing substring tests (they guard semantic content).

Scope held to `context_get` + `context_graph`; the `include_str!` sidecar was rejected in design review (no precedent, doc-concat + trailing-newline footguns). The remaining ~14 tool descriptions carry the same latent structure and are deferred to a follow-up issue.

## Tests
- New guards: `test_get_tool_attr_description_matches_const`, `test_graph_tool_attr_description_matches_const` — both pass.
- `cargo test --workspace --lib`: 6128 passed, 0 failed. Clippy clean. Smoke 26/26. `protocol` + get/graph `tools` suites green.
- Full-workspace `cargo test` rc=101 is an environmental linker OOM on integration-test binaries (SIGKILL on `ld`, swap exhausted) — not a defect and not caused by a test-only lib change.

### DIFF
--- a/crates/unimatrix-server/src/mcp/tools.rs
+++ b/crates/unimatrix-server/src/mcp/tools.rs
@@ -44,11 +44,15 @@ use crate::uds::hook::MAX_GOAL_BYTES;
 // pub(super) so the sibling test module can reference it via use super::REDIRECT_CEILING.
 pub(super) const REDIRECT_CEILING: usize = 50;
 
-/// Tool description for context_graph (vnc-018, vnc-019, vnc-020).
+/// Tool description for context_get (vnc-042).
 ///
-/// Exposed as `pub(crate)` so sibling test modules can assert on disclosure text
-/// without re-embedding the string (AC-13, R-11 vnc-019 test plan; AC-19, R-01 vnc-020).
-/// Must match the `#[tool(description = "...")]` attribute literal on `context_graph`.
+/// Exposed as `pub(crate)` so sibling test modules can assert on the description text
+/// without re-embedding the string (FR-13, R-09 proxy).
+/// Must stay byte-identical to the `#[tool(description = "...")]` attribute literal on
+/// `context_get`; that invariant is enforced by the byte-equality guard test
+/// `test_get_tool_attr_description_matches_const` (the #[tool] macro cannot consume a
+/// const in `description =`, so the guard is the only enforced link between the two
+/// literals — the substring test alone cannot detect drift).
 // rationale: the #[tool] macro requires an inline string literal, so this single-source
 // constant is asserted against only by sibling test modules.
 ///
@@ -6037,8 +6041,12 @@ mod tests {
         // handler must contain the exact staleness documentation text mandated by
         // ARCHITECTURE.md and ADR-005.
         //
-        // Uses CONTEXT_GRAPH_DESCRIPTION directly — the single source of truth — so this
-        // test always reflects the live constant rather than a stale inline copy.
+        // Asserts substrings against the CONTEXT_GRAPH_DESCRIPTION mirror const to guard the
+        // SEMANTIC content of the description. NOTE: this const is NOT the single source of
+        // truth — the live client-facing description is a separate `#[tool(description = "...")]`
+        // literal on the handler (rmcp 1.7.0 cannot consume a const there). Byte-equality
+        // between the two literals is enforced by test_graph_tool_attr_description_matches_const;
+        // this substring test alone cannot detect drift in the live literal (#869).
         //
         // vnc-019: Extended to assert AC-13 subgraph disclosure facts.
         // vnc-020: Updated from "four modes" to "seven modes".
@@ -6103,6 +6111,23 @@ mod tests {
         assert!(
             description.contains("seven modes"),
             "description opening must say seven modes (vnc-020 added inverse, filter, path)"
+        );
+    }
+
+    #[test]
+    fn test_graph_tool_attr_description_matches_const() {
+        // #869: identical twin of context_get. The live `#[tool(description = "...")]` literal
+        // on the context_graph handler and the CONTEXT_GRAPH_DESCRIPTION mirror const are two
+        // independent literals; rmcp 1.7.0 cannot single-source the const into the attribute.
+        // This guard makes the byte-equality invariant CI-enforced by comparing the LIVE
+        // macro-generated description (what MCP clients receive) against the mirror const the
+        // substring test asserts on. A single-byte divergence now fails here.
+        assert_eq!(
+            UnimatrixServer::context_graph_tool_attr()
+                .description
+                .as_deref(),
+            Some(CONTEXT_GRAPH_DESCRIPTION),
+            "live context_graph tool-attr description must be byte-identical to CONTEXT_GRAPH_DESCRIPTION"
         );
     }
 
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — 869-security-reviewer (fresh context)

**Risk level: LOW.** No blocking findings.

Reviewed `git diff main...bugfix/869-context-desc-drift-guard` cold against the investigation/design artifacts in #869.

### Scope verification
- Single file, +50/-8. Every changed line is inside `#[cfg(test)]` modules (`mod tests`, `mod get_resolution_tests`) or doc/inline comments. Confirmed: two byte-equality guard tests added, one test import extended (`UnimatrixServer`), and two source comments corrected (stale `context_graph` copy-paste on the `context_get` const doc; false "single source of truth" claim on the `context_graph` substring test).
- The mirror consts (`CONTEXT_GET_DESCRIPTION` :63, `CONTEXT_GRAPH_DESCRIPTION` :65) are `pub(crate)` and referenced only from tests. Their string values are unchanged by this diff — no client-facing tool description is altered.
- No production code path touched. No `Cargo.toml` change; no new dependency. No I/O, no deserialization, no path handling, no shell/format-string construction introduced.

### OWASP concerns
- Injection / path traversal / deserialization / access control: **N/A** — test-only assertions comparing two static, compile-time-embedded string literals. No untrusted input, no boundaries crossed.
- Secrets: none. Both consts are user-facing MCP tool documentation strings.

### Blast radius (worst case if the new tests have a subtle bug)
Contained to CI signal. A wrong assertion could only (a) fail spuriously and block a merge, or (b) fail to detect drift it was meant to catch. Neither can corrupt data, leak information, or affect the serving path — the guards run in the test binary and call the macro-generated `*_tool_attr()` constructor (cheap, `Cow::Borrowed<'static>`). The guards strictly *increase* safety: they close the exact silent-drift gap #869 identified, where the client-facing `#[tool(description=...)]` literal could diverge from the tested mirror const with zero failing test.

### Regression risk
Negligible. Test-only + comment edits cannot change runtime behavior. Fix summary reports both guards pass on first run (const and live literal byte-identical today), workspace `--lib` green, clippy clean. The full-workspace `cargo test` rc=101 is a documented environmental linker OOM, not a defect in this change.

### Note (non-blocking, already tracked)
The remaining ~14 `#[tool]` descriptions carry the same unguarded-twin structure and are correctly deferred to a follow-up per design review. Not a gap in this PR.

**Blocking findings: none.** Approve from a security/risk standpoint.

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. The delivery matches the approved fix-approach exactly; security review is LOW with no blocking findings. One loose end: file the promised follow-up issue for the remaining ~14 tool descriptions before/at merge so the deferral is tracked, not lost.

### Vision / roadmap fit
Correctness-integrity hardening on the client-facing MCP tool surface — the contract of a knowledge engine whose value proposition is trustworthy, consistent knowledge (`self-learning` #5219, `personal-cloud` #4946). Maps to standing knowledge #4301 (verify each tool description matches its code path) and #4303 (a description that misrepresents its tool silently misleads every client). No roadmap item is advanced or blocked; an existing surface is hardened. Low strategic weight, correct to fix, correct to keep small.

### Delivery-as-a-whole assessment (incl. security review)
- **Root cause closed, not symptom.** The defect was a missing *enforced* link between each tool's live `#[tool(description=...)]` literal (what clients receive) and its mirror const; the old substring tests asserted only against the const. Both new guards compare the LIVE macro-generated `*_tool_attr().description` against the mirror — the exact comparison that was absent. This delivers the OUTCOME (drift cannot pass CI), not just a mechanism.
- **Matches the approved approach precisely.** Guarded-literal path taken; `include_str!` sidecar rejected (correctly — unprecedented single-tool convention plus `///`-concatenation and no-trailing-newline footguns); both existing substring tests retained (they guard content semantics byte-equality does not). The `context_graph` last-mile the fix-approach review flagged — the FALSE "single source of truth" comment plus the identical unguarded twin — was pulled into THIS fix. A false guarantee is worse than a missing one, so correcting it here rather than deferring it was the right call.
- **Security review (869-security-reviewer): LOW, no blocking findings — concur.** Test-only + comment edits, single file +50/-8, no production path, no dependency, no I/O. Worst case of a buggy guard is contained to CI signal (spurious fail or missed drift); it strictly increases safety. Nothing to action.
- **Validation is sound.** Both guards pass (proving byte-identity today), `cargo test --workspace --lib` rc=0 (6128 passed; server 4327), clippy clean, smoke 26/26, protocol + get/graph tools suites green. The single red (`cargo test --workspace` rc=101) is correctly classified as an environmental linker OOM on integration-test binaries — a test-only lib change cannot introduce a linker error and does not touch those `tests/*.rs` binaries. GH CI (JS client matrix + GitGuardian) all green; per this repo's model the Rust validation is the protocol gates, which passed.

### Capability coverage
N/A — bugfix, not a capability delivery. This hardens an existing surface (the `tools/list` description contract); it does not advance a capability's `done_when`.

### Right-sizing
Outcome this had to achieve: the client-facing `context_get`/`context_graph` descriptions cannot silently diverge from what tests/clients see — an ENFORCED invariant, not "a const exists."
- Size verdict: **right-sized.** Two guards + two comment corrections + retained substring tests is the outcome; no gold-plating, no mechanism-without-outcome.
- Deferral — the systematic sweep of the remaining ~14 `#[tool]` descriptions is a genuinely DIFFERENT, larger outcome (legit defer). Not the hidden point of this bug.
- Follow-up smell: the deferral is honest, BUT the promised follow-up issue does not exist yet (searched open+closed — only #869 itself). Multiple artifacts (investigator, design review, fix-approach, PR body) commit to it; if it is not filed, the ~14-tool latent-twin structure silently rides along with no tracker. Recommend filing it before/at merge — cheap, and it is the difference between an honest deferral and a dropped one.

No open questions remain in the artifacts — the fix-approach gate resolved all three (guarded-literal over sidecar; include context_graph; keep substring tests), and the delivery honored each.

### Recommended actions
- **Merge as-is** once the follow-up is tracked. The diff, gate PASS, and security review support ship.
- **File the deferred follow-up GH issue** (guard the remaining ~14 `#[tool]` descriptions; this PR's guard shape is the template) — the one promise not yet kept.
- At retro, capture the reusable rmcp-1.7.0 insight (`#[tool(description=…)]` accepts only a string literal; `{fn}_tool_attr().description` exposes the live value for byte-equality tests) via /uni-store-pattern — as a pattern, not a bug-lesson.

PR #875, opened 2026-07-02
fix: client-side dormancy reconnect + session-id restore + errors-only observability (#872)

### LINKED ISSUE #872: Personal Cloud: client still hangs on long non-interaction Does not recover.
On the client, after a long period of inactivity, the next call to unimatrix will hang.  the behavior is same s one time we tried to fix this before: 

The user doesn't want to proceed with this tool use. The tool use was rejected
  (eg. if it was a file edit, the new_string was NOT written to the file).
  STOP what you are doing and wait for the user to tell you how to proceed.

  Followed by a system line:
  [Request interrupted by user for tool use]

  The critical distinction

  That message is NOT from Unimatrix. It's from the harness/permission layer, and it means the call was
  stopped before it ever reached the Unimatrix MCP server. So on these recent attempts I have received zero 
  response from Unimatrix itself — no data, no error, no timeout string. Nothing from the server reached me,
  because the call didn't get there.

  From my side I can only ever see one of two things: (a) a real tool result, or (b) that
  interrupt/rejection text above. I cannot tell "hung for 10 minutes then you killed it" apart from 
  "rejected instantly" — both surface to me identically as that interrupt message. I have not seen a hang, a
  timeout error, or any Unimatrix error string in these attempts.

The expectation is that the client can recover from a short medium/long term periods of inactivty.  Nothing ever has to be done to the server during these events, just reconnect to mcp server from the client... then everything works fine.

dug-21:
Fixes #872.

## Problem
The Personal Cloud edge client hangs on the first call after a long idle period and never recovers. The reporter expects the **client** to recover on resume by reconnecting — no server action.

## Why prior fixes didn't close it
The #831 / #842 / #848 lineage armed reactive timeouts, routed `ETIMEDOUT` into self-heal, and ref'd the F6 idle-read timers. That machinery is **reactive and single-shot with no notion of elapsed idle time** — every recovery path only fires *after* a live call has already burned a connect (15s) or idle (120s) deadline, and one path could wedge the session.

## Root cause (two gaps)
- **Gap 1 — no proactive dormancy reconnect.** Recovery never fires until a deadline is paid, so it reads as a hang. This was #839's explicitly-deferred item 4; never built.
- **Gap 2 — session-id null-poisoning wedge.** `_doReinit` nulled `sessionId` before the re-init POST; if that POST also timed out, `sessionId` stayed `null`, and the bridge then dispatched every later call session-less. rmcp 1.7.0 answers a session-less non-init POST with **HTTP 422** (not a heal-triggering 404) — a permanent wedge.

## Fix (client-side, JS/TS mcp-bridge only)
1. **Gap 1** — track `lastActivityMs` (seeded to `now`); in `handle()`, run the existing single-flight `_reinit()` proactively when `now - lastActivityMs > STALE_MS`, **fail-open** (fall through to normal dispatch if it fails), on the short connect budget. `STALE_MS` default **2.5h** (62.5% of the 4h server keep-alive), override `UNIMATRIX_HOOK_MCP_STALE_MS`.
2. **Gap 2** — restore `prevId` on a failed re-init; never leave `sessionId === null`.
3. **Observability (errors-only, always-on)** — via the existing `errOut`→stderr seam, one dialect (`mcp-bridge: <event> key=value`):
   - `transport_timeout phase=connect|idle elapsed_ms=<int> idle_ms=<int>` — `idle_ms` distinguishes dormancy timeout from server-down.
   - `reinit_fail session=restored`.
   - Redacted (no Bearer token, no full session id); every emit wrapped in a true no-op try/catch so a logging throw can never abort the heal — enforced by a test.

## Tests (DoD)
7 new tests in `mcp-bridge.test.js`:
- `test_dormancy_proactiveReconnectFiresBeforeAnyTimeout` — proactive reconnect fires before a slow failure is paid.
- `test_healExhaustion_preservesSessionId_notWedgedSessionless` — session preserved after heal-exhaustion.
- `test_proactiveReconnect_boundedToConnectBudget` — recovery bounded to the connect budget.
- `test_b1_errOutThrows_healStillCompletes` — a throwing `errOut` never aborts the heal.
- `test_dormancy_freshSessionNoProactiveReinit`, `test_observability_transportTimeout_hasIdleMs_noSecrets`, `test_observability_reinitFail_sessionRestored_noSecrets`.

Results: bridge 61/0 (7 new), hook-client 771 pass/1 platform-skip, integration smoke 26/0, cross-transport parity (layer2 HTTPS-vs-UDS) 24/0, `cargo build --workspace` clean (0 Rust files changed).

## Convergence (on the record)
This session believes #872 is the **terminal fix for the reported dormancy/wake hang mode**: it builds the proactive reconnect the whole lineage deferred and closes the non-recovery wedge that made prior reactive patches fail to recover. It does **not** claim closure of the entire half-open-transport class — the errors-only observability (`idle_ms` / `transport_timeout` / `reinit_fail`) is shipped deliberately so any future instance arrives with evidence instead of guesswork.

### DIFF
--- a/packages/unimatrix/lib/hook-client/mcp-bridge.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge.js
@@ -61,7 +61,9 @@ function run(session, deps) {
   const exit = (deps && deps.exit) || ((c) => process.exit(c));
 
   const framer = new StdioFramer(stdin, stdout);
-  const lifecycle = new Lifecycle(session);
+  // Forward deps so the lifecycle's errOut floor + injectable clock are wired
+  // (N3: run() previously built Lifecycle with no deps → stderr-only, untestable).
+  const lifecycle = new Lifecycle(session, deps);
 
   framer.onMessage(async (msg) => {
     let response = null;
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
@@ -130,8 +130,10 @@ class HttpSession {
       // Connect/TLS-handshake deadline (F4): armed before connect, cleared in
       // secureConnect (clearCt) — covers the half-open-socket stall the socket
       // `timeout` option does NOT. Idle (post-connect): Node emits "timeout".
+      // REF'd on purpose (not unref'd): #872 recovery runs on an idle/dormant
+      // loop where an unref'd deadline never fires (#847/#848). Every terminal
+      // path routes through settle() -> clearCt(), so it never leaks a handle.
       const ct = setTimeout(() => kill("connect"), this.connectMs);
-      ct.unref();
       const clearCt = () => clearTimeout(ct);
       const settle = (fn, v) => { if (settled) return; settled = true; clearCt(); fn(v); };
       req.on("timeout", () => kill("idle"));
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/http-session.js
@@ -37,6 +37,10 @@ class HttpSession {
     this.requester = (o) => https.request(o); // injectable for tests
     this.connectMs = connectMs || 15000;
     this.idleMs = idleMs || 120000;
+    // Last successful round-trip wall-clock; mirrored by the lifecycle so a
+    // transport timeout can report idle_ms (dormancy vs server-down). Seeded to
+    // now so an early failure reports a small, sane idle span (#872).
+    this.lastActivityMs = Date.now();
   }
 
   _headers(body, opts) {
@@ -104,16 +108,33 @@ class HttpSession {
       const req = this.requester(this._options("POST", this._headers(body, opts)));
       let flushed = false;
       let settled = false; // single settle-guard (F3, vnc-039 C2 double-settle)
+      let killed = false; // one-shot: emit the timeout event + destroy at most once
+      const startMs = Date.now();
+      // Errors-only observability (#872, ALWAYS-ON, stderr seam only). connect vs
+      // idle is ONLY distinguishable here. idle_ms distinguishes dormancy timeout
+      // from server-down. NO token, NO session id (redaction). B1: try/catch no-op.
+      const emitTimeout = (phase) => {
+        try {
+          this.errOut("mcp-bridge: transport_timeout phase=" + phase +
+            " elapsed_ms=" + (Date.now() - startMs) +
+            " idle_ms=" + (Date.now() - this.lastActivityMs) + "\n");
+        } catch (_e) {}
+      };
       // ETIMEDOUT-coded so lifecycle.js maps it to "MCP endpoint timed out".
-      const kill = () => { try { req.destroy(Object.assign(new Error("timed out"), { code: "ETIMEDOUT" })); } catch (_e) {} };
+      const kill = (phase) => {
+        if (killed) return;
+        killed = true;
+        emitTimeout(phase);
+        try { req.destroy(Object.assign(new Error("timed out"), { code: "ETIMEDOUT" })); } catch (_e) {}
+      };
       // Connect/TLS-handshake deadline (F4): armed before connect, cleared in
       // secureConnect (clearCt) — covers the half-open-socket stall the socket
       // `timeout` option does NOT. Idle (post-connect): Node emits "timeout".
-      const ct = setTimeout(kill, this.connectMs);
+      const ct = setTimeout(() => kill("connect"), this.connectMs);
       ct.unref();
       const clearCt = () => clearTimeout(ct);
       const settle = (fn, v) => { if (settled) return; settled = true; clearCt(); fn(v); };
-      req.on("timeout", kill);
+      req.on("timeout", () => kill("idle"));
       this._pinThenFlush(
         req,
         () => { flushed = true; req.end(body); }, // pin OK — flush Bearer body
--- a/packages/unimatrix/lib/hook-client/mcp-bridge/lifecycle.js
+++ b/packages/unimatrix/lib/hook-client/mcp-bridge/lifecycle.js
@@ -20,12 +20,36 @@ const CLIENT_INFO_NAME = "unimatrix-mcp-bridge"; // STABLE, fixed (FR-03a/AC-12)
 // Exhausted-heal normalization: an internal sentinel never reaches the client.
 const SENTINEL_TEXT = { [SESSION_NOT_FOUND]: "MCP session lost, re-init failed", [TRANSPORT_TIMEOUT]: "MCP endpoint timed out" };
 
+// Proactive dormancy reconnect (#872 Gap 1). After the client sleeps past the
+// server's rmcp keep_alive (4h), the pre-sleep socket is silently dead; the next
+// call would eat a full connect/idle deadline before healing. If we have been
+// idle beyond STALE_MS, re-init BEFORE dispatch. Default ~2.5h (62.5% of the 4h
+// server TTL) so we preempt eviction with margin; overridable for tuning/tests.
+const DEFAULT_STALE_MS = 9000000; // 2.5h in ms
+function resolveStaleMs() {
+  const n = parseInt(process.env.UNIMATRIX_HOOK_MCP_STALE_MS, 10);
+  return Number.isFinite(n) && n > 0 ? n : DEFAULT_STALE_MS;
+}
+
 class Lifecycle {
   constructor(session, deps) {
     this.session = session;
     this.initMessage = null; // captured client initialize, replayed on re-init (B3)
     this.reinitInFlight = null; // single-flight guard for concurrent 404s (C2)
     this.errOut = (deps && deps.errOut) || ((s) => process.stderr.write(s));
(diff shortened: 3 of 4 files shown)


dug-21 (commented):
## Security Review — bugfix-872 (uni-security-reviewer, fresh context)

**Risk level: LOW. No blocking findings.** Diff read cold against `main...HEAD` (4 files, all under `lib/hook-client/mcp-bridge/` + its test). Lockfile/package.json unchanged — no new dependencies.

### Verified (the things the spawn asked me to prove from the diff, not the tests)
- **No secret leakage in the new stderr observability.** Both new emitters interpolate non-sensitive values only:
  - `transport_timeout phase=<connect|idle> elapsed_ms=<int> idle_ms=<int>` — integers + a fixed phase literal.
  - `reinit_fail session=<restored|null>` — the tag is `sessionId != null ? "restored" : "null"`, i.e. the *word*, never the id value.
  - No `Bearer`/token and no raw `Mcp-Session-Id` anywhere on the new paths. Redaction holds in the source, not just fixtures.
- **stdout MCP channel untouched.** Every new write routes through `this.errOut` (defaults to `process.stderr.write` in both `lifecycle.js:28` and `http-session.js:36`). No `process.stdout` write introduced. The `deps` forwarding in `mcp-bridge.js` only wires the errOut floor + injectable clock.
- **Input validation on the one new external input.** `UNIMATRIX_HOOK_MCP_STALE_MS` is `parseInt` + guarded `Number.isFinite(n) && n > 0` before use, else falls back to `DEFAULT_STALE_MS`. Rejects NaN/negative/zero. Numeric-only — no injection surface. No path/shell/SQL anywhere in the change.

### Blast-radius assessment (worst realistic case)
- **Reconnect storm — bounded, self-limiting.** Proactive `_reinit()` reuses the existing single-flight `reinitInFlight` guard, and a *successful* proactive re-init resets the dormancy clock via `_touch()` inside `_send`, so it cannot re-trigger until `STALE_MS` elapses again. On a *failed* proactive probe during a dormant-and-server-down window the clock is not reset, so each subsequent client call attempts one extra re-init (≈2x transport attempts/call) — but it is fail-open, connect-budget-bounded, and paced by client call rate, not a runaway loop. Acceptable; arguably desired retry behavior.
- **Restored-but-dead session — intentional and safe.** Gap 2 restores `prevId` (the evicted id) rather than leaving `sessionId === null`. Worst case the restored id is stale → server 404 → existing #830 heal re-fires. That is strictly better than the null→422 permanent wedge it replaces. Session ids are opaque server-minted values; restoring an old one has no auth consequence. Ordering is correct: `sessionId = prevId` is assigned *before* the wrapped `reinit_fail` errOut, so even a throwing logger cannot leave the session un-restored.

### Regression risk vs the #842/#848 timeout/self-heal machinery
- Low. `kill()` gained a `phase` arg and a `killed` one-shot guard; both call sites updated (`setTimeout(()=>kill("connect"))`, `req.on("timeout", ()=>kill("idle"))`). ETIMEDOUT coding is unchanged, so the TRANSPORT_TIMEOUT sentinel → self-heal routing is preserved. The `killed` guard is additive (emit+destroy at most once); the separate `settled` guard is untouched.
- Every heal-path `errOut` is now wrapped in a true no-op try/catch (including the pre-existing `session evicted (404)` line), closing the "a logging throw aborts the heal it is observing" failure mode.

### Note (non-blocking, pre-existing)
The proactive path nulls `sessionId` during the inner re-init POST, widening the pre-existing window in which a *concurrently pipelined* request could observe `sessionId === null` and dispatch session-less. This window already exists in the #830 reactive `_doReinit`; the change adds a new trigger, not a new class. For a single-client stdio bridge this is tolerated by design. No action required; flagging for the record.

**Verdict: approve on security grounds. No changes requested.**

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. The delivery closes the reported "does not recover" wedge (Gap 2, server-422 verified) and ships the reporter's actual asked-for behavior (Gap 1, #839's deferred item), correctly scoped and clean on security. The one residual the human should hold: this bug class cannot be reproduced live in CI, so merge confidence rests on the diagnosis + proxy tests, and the shipped observability is the real-world validation instrument.

### Vision / roadmap fit
Squarely on `personal-cloud` (#4946) — "full pipeline fidelity over HTTPS, one bearer token, just works." The defect is the HTTPS bridge silently failing the "just works" contract after client dormancy, which is exactly the personal-cloud reliability promise. Fixing it client-side is correct per Architectural Principle 6 (the client is an adapter; resilience lives there) and keeps parity with the local-UDS path. No collision with in-flight `personal-cloud` work — this is reliability hardening of already-shipped serving, continuing the #830 → #839/#842 → #848 self-heal lineage and closing the dormancy/silent-eviction gap #839 explicitly deferred.

### Approach commentary
Sound and minimal. Reuses the shipped single-flight `_reinit()`/sentinel machinery instead of a parallel reconnection engine; no background `setInterval` keepalive (correct for a stdio bridge process — the staleness check is event-driven on dispatch). Gap 2 (`prevId` restore) is the load-bearing fix — it alone converts the null→422 permanent wedge into a restored-id→404→#830 heal. Gap 1 is the reporter-facing latency/observability win. The whole reviewer chain converged (design APPROVED-WITH-NOTES, observability APPROVED-WITH-NOTES, security LOW/no-blockers, gate 7/7 PASS, tester all-green) and every blocking note (F1 seed-to-now, F2 fail-open, B1 wrapped errOut, B2 redaction, single dialect) is visibly implemented in the diff. Two observations worth the human's eye, neither blocking:

- **Composed worst-case latency is bounded but not tiny.** A fully-dead-and-server-down wake now chains proactive re-init → restore `prevId` → dispatch → reactive re-init, i.e. up to ~3× `connectMs` (~45s at the 15s default) before a bounded error surfaces. It is loop-free and single-shot ("heal once then surface"), and the common case (server alive on wake) resolves in ≈1 connect round-trip. Design-review F4 asked for this number to be "stated, not discovered in the field" — it is implied by the diff but not written down. Worth a one-line note on the issue so it isn't re-litigated.
- **`idle_ms` on `transport_timeout` slightly exceeds the trimmed 2-event floor** the observability reviewer specified (their final block showed only `elapsed_ms`). It is a redacted integer, security-verified to leak nothing, and it is what actually distinguishes dormancy-timeout from server-down — a benign, useful addition, not scope creep to worry about. It does introduce a small `lastActivityMs` mirror across Lifecycle↔HttpSession (seeded in both ctors, Lifecycle overwrites on construction); harmless redundancy.

### Capability coverage
Not applicable at a bugfix pr-review — no new capability with a `done_when` to map; this is a defect fix against already-delivered `personal-cloud` serving. The correct frame is regression-vs-outcome (below), consistent with the fix-approach review.

### Right-sizing
Outcome / DoD altitude: the bridge must ACTUALLY recover (first, or at worst next, call after long dormancy) within a bounded budget — the property WORKS, not "a reconnect path exists." The tests target that behaviorally (proactive re-init fires before a slow deadline; sessionId preserved after heal-exhaustion; recovery bounded to the connect budget; throwing errOut never aborts the heal). 
Size verdict: right-sized. No gold-plating (no periodic keepalive engine), and the last mile (Gap 1 proactive path + Gap 3 short budget) was pulled in rather than deferred.
Follow-up smell: none that hides the point. The convergence note is honest — it claims closure of the reported dormancy/wake mode only, NOT the whole half-open-transport class, and ships errors-only observability precisely so any future instance arrives with evidence. That is a legitimately different outcome to defer, not a deferred last mile.

### Residual risk (the one thing to weigh before merge)
This defect is intermittent and requires a genuine half-open socket after a >4h sleep — it cannot be reproduced in CI. Every test is a deterministic proxy (injected clock, silent TCP listener, injected requester). Merge confidence therefore rests on (a) the diagnosis, including the server-422/404 behavior uni-zero verified at fix-approach, and (b) proxy coverage — not a live dormancy repro. That is acceptable and inherent to this bug class, but it means the true validation is the first real post-dormancy recovery observed in the field via the new `transport_timeout idle_ms=…` / `reinit_fail session=restored` lines. Recommend treating those log lines as the acceptance evidence and confirming one real recovery post-merge before considering the dormancy mode closed.

No open questions remain in the PR. The one prior open item — server response to a session-less POST — was answered at fix-approach (rmcp 1.7.0: session-less non-init → 422; unknown/evicted id → 404) and the fix correctly does NOT teach the bridge to heal on 422 (that would reintroduce the re-init-storm risk #839 warned against); it fixes the root cause by never going session-less.

### Recommended actions
1. Merge as-is — no rework indicated. All blocking review notes are implemented; security is LOW with no changes requested.
2. Add a one-line note (issue or PR) stating the composed worst-case bound (~3× `connectMs`) so F4's "state, don't discover" ask is on the record.
3. Post-merge: watch for the first real `reinit_fail session=restored` / `transport_timeout … idle_ms=…` in the field as the live-recovery evidence; only then consider the dormancy mode validated.
4. Post-merge knowledge step (vision session, not this PR): extend #5303 via `context_correct` tagged `caused_by_feature:bugfix-872` rather than a new entry — the generalizable insight (reactive-only self-heal is insufficient without proactive dormancy reconnect; nulling sessionId on a failed re-init poisons recovery into a 422 wedge) becomes validated once the field evidence lands.

_Independence note: spawn prompt was clean (agent ID, gate, issue/PR numbers, artifact pointers only) — no summaries or conclusions to discount._

Issue #865, opened 2026-06-29
parity lane: fold D5 precompact documented exception into exit-code exclusion set so the matrix can reach green


dug-21:
## Problem

On the first live `workflow_dispatch` round of the nan-022 cross-transport parity matrix (run [28371921936](https://github.com/dug-21/unimatrix/actions/runs/28371921936), lane `nan-021-https-uds-parity`, after #849 provisioned `sqlite3`), the matrix reached a **real per-dimension verdict** — but the lane still exits RED:

| Dimension | Outcome |
|-----------|---------|
| retrieval | PARITY_PASS |
| behavioral | PARITY_PASS |
| analytics | PARITY_PASS |
| proactive | PARITY_PASS |
| **precompact** | **INFRA_ERROR** (`exit_code: 7`) |

Result: `1 failed, 2 passed, 10 deselected`; overall `verdict: ERROR`.

The `precompact` (D5) `INFRA_ERROR` is **by design** — CompactContext restoration has a Claude-Code host-side component the harness cannot drive test-only (OQ-2 / ADR-006). It is already a human-signed **documented exception** (see #837): *"measured-where-drivable + documented host-side gap ... never rounded up to fully measured ... host-side gap → documented call-out, never a drop."* The matrix correctly lists it under `documented_exceptions`, and the harness itself labels the exit as *"distinct exit 7=7 — NOT a parity RED."*

## The gap

The other four dimensions report *"parity clean modulo closed exclusion set"* — i.e. an exclusion-set mechanism already exists. But the D5 precompact documented exception is **not** in that exit-code exclusion set, so it drives `exit_code: 7` and reds the whole lane. The result: the lane cannot reach green even though every drivable dimension is clean.

This tension was anticipated in #837's design (documented exception, "never a drop") but #837 is CLOSED and did not resolve the **exit-code treatment** of a human-signed documented exception.

## Proposed fix

Fold the D5 precompact human-signed documented exception into the exit-code exclusion set (the same "closed exclusion set" the other dimensions already honour), so the matrix can reach green on a clean drive **without** rounding measurability up to "fully measured." The dimension must still:
- report its true `INFRA_ERROR` / documented-exception outcome in the evidence table (never silently dropped, never rounded up),
- keep `blocks_c0_proof: true` so the C0 flip bar is unchanged (the flip session still requires the human-signed exception, per #837).

[...]

PR #877, opened 2026-07-02
fix: observe SessionClose 204-no-op is observable (#819)

### LINKED ISSUE #819: Observability: observe that Acks 204 but persists nothing is silent

dug-21:
## Summary
Closes the blind spot where an \`/observe\` **SessionClose Acks 204 while persisting nothing** (absent/already-drained session) with no trace or counter — the silent no-op that hid the Gate 7 bug (#818). **Observability only; the idempotent 204 contract is unchanged.**

## Changes
- **Server (adjunct signal)** — `process_session_close` (`crates/unimatrix-server/src/uds/listener.rs`): emit one `debug!(target: "unimatrix_server::obs", session_id = %session_id, ...)` when `maybe_output.is_none()` (the complete zero-write predicate) before the `Ack`. Uses the mandatory `unimatrix_server::obs` target per ADR #3467. No change to `observe.rs`, the `Ack→204` mapping, or `dispatch_request` signatures.
- **Automated blind-spot closure (mandatory core)** — two integration smoke assertions in `product/test/infra-001/suites/test_lifecycle.py`:
  - `test_session_close_drain_persists_sessions_row_but_ghost_close_is_zero_delta` — **#818's exact escape path.** Positive: `SessionRegister → SessionClose` (no PostToolUse between) flips `sessions.ended_at`/`outcome` NULL→set — the `store.update_session()` drain flush #818 dropped. Negative: never-registered ghost `SessionClose` yields **zero delta on `sessions` and `signal_queue`**, still 204. Load-bearing: reads `ended_at` NULL before / set after, so a silently-no-op'd registered drain leaves it NULL → RED.
  - `test_observe_row_count_delta_persist_vs_noop_close` — `observations` row-count delta (0 no-op / +1 persist) for the PostToolUse observe arm.
- **Rust controls** — two `#[traced_test]` tests: absent session → `Ack` + trace; real close → `Ack`, no trace.

## Validation
- Rust lib (`cargo test --workspace --lib`): 6130 passed, 0 failed
- Clippy (`--workspace -D warnings`): clean
- Smoke (`pytest -m smoke`, incl. both new tests): 28 passed, 0 failed
- Full `cargo test --workspace` hit a pre-existing **link-step OOM** (`ld` killed) on integration binaries unrelated to the observe path — environmental memory ceiling; observe coverage via lib controls + the Python end-to-end smokes.

## Design decisions
- **Trace not counter**: no metrics registry / `/metrics` endpoint; ADR #865 counters are DB rows, not runtime metrics.
- **`debug!` not `warn!`**: correct idempotent behavior — a warn would be a false alarm (inverts goal N4). Not `trace!` — the UDS obs group standardizes on `debug!`.

Gate: Bugfix Validation — **PASS** (initial 13/13; focused re-gate 5/5 confirming exact-#818-path coverage). Security Review — **LOW**, no blocking findings.

Closes #819.

### DIFF
--- a/crates/unimatrix-server/src/uds/listener.rs
+++ b/crates/unimatrix-server/src/uds/listener.rs
@@ -2341,6 +2341,21 @@ async fn process_session_close(
         run_retrospective_consumer(store, pending, entry_store, fc_key).await;
     }
     // If session absent (already cleared): no-op (idempotent — AC-03)
+    // #819 (F2): maybe_output.is_none() alone is the complete zero-write predicate
+    // for the closing session — Steps 3/4 (SESSIONS update, SIGNAL_QUEUE write,
+    // consumers) were all skipped. Emit ONE obs-target debug event so this
+    // acked-but-persisted-nothing close is distinguishable from a fully-persisted
+    // one (the #818 blind spot). F1 (ADR #3467 / pattern #4602, GH #381): the
+    // target: "unimatrix_server::obs" is MANDATORY — it keeps this new UDS
+    // observation log point in the operator-togglable obs group. debug! (not
+    // warn!) because this is correct idempotent behavior, not a fault (goal N4).
+    if maybe_output.is_none() {
+        tracing::debug!(
+            target: "unimatrix_server::obs",
+            session_id = %session_id,
+            "UDS: SessionClose acked (204) with zero store writes for the closing session: session absent or already drained"
+        );
+    }
 
     HookResponse::Ack
 }
@@ -3634,6 +3649,90 @@ mod tests {
 
         // col-008: verify session cleared
         assert!(registry.get_state("s1").is_none());
+    }
+
+    /// #819 (a) — negative control: a SessionClose against an EMPTY registry
+    /// (absent/already-drained session) still returns Ack (204 idempotency
+    /// contract unchanged) AND emits the zero-store-writes obs trace so the
+    /// no-op is distinguishable from a fully-persisted close (the #818 blind spot).
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn test_session_close_absent_session_returns_ack_and_emits_zero_write_trace() {
+        let store = make_store().await;
+        let embed = make_embed_service();
+        let registry = make_registry();
+        let (vs, es, adapt) = make_dispatch_deps(&store);
+
+        // No register_session — the registry is empty, so drain returns None.
+        let response = dispatch_request(
+            HookRequest::SessionClose {
+                session_id: "absent-819".to_string(),
+                outcome: Some("success".to_string()),
+                duration_secs: 60,
+            },
+            &store,
+            &embed,
+            &vs,
+            &es,
+            &adapt,
+            "0.1.0",
+            &registry,
+            &make_pending(),
+            &make_services(&store, &embed, &vs, &es, &adapt),
+            crate::uds::UDS_CAPABILITIES,
+        )
+        .await;
+
+        // Idempotency contract unchanged: still Ack (→ 204).
+        assert!(matches!(response, HookResponse::Ack));
+        // The distinguishing signal: the zero-store-writes obs trace fired.
+        assert!(
+            logs_contain("zero store writes"),
+            "an absent/already-drained SessionClose must emit the zero-store-writes obs trace (#819)"
+        );
+    }
+
+    /// #819 (b) — positive control: a SessionClose for a REAL registered session
+    /// drains and persists, returns Ack, and must NOT emit the zero-store-writes
+    /// trace (maybe_output.is_some()), so the trace does not become noise on
+    /// productive closes.
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn test_session_close_registered_session_returns_ack_without_zero_write_trace() {
+        let store = make_store().await;
+        let embed = make_embed_service();
+        let registry = make_registry();
+        let (vs, es, adapt) = make_dispatch_deps(&store);
+
+        // A real, registered session — drain returns Some(output).
+        registry.register_session("real-819", None, None);
+
+        let response = dispatch_request(
+            HookRequest::SessionClose {
+                session_id: "real-819".to_string(),
+                outcome: Some("success".to_string()),
+                duration_secs: 60,
+            },
+            &store,
+            &embed,
+            &vs,
+            &es,
+            &adapt,
+            "0.1.0",
+            &registry,
+            &make_pending(),
+            &make_services(&store, &embed, &vs, &es, &adapt),
+            crate::uds::UDS_CAPABILITIES,
+        )
+        .await;
+
+        assert!(matches!(response, HookResponse::Ack));
+        assert!(registry.get_state("real-819").is_none());
+        // A productive close must stay silent on the zero-write obs trace.
+        assert!(
+            !logs_contain("zero store writes"),
+            "a real drained SessionClose must NOT emit the zero-store-writes obs trace (#819)"
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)


dug-21 (commented):
## Security Review — 819-security-reviewer (fresh context)

**Risk: LOW. No blocking findings.** Observability-only change; the idempotent 204 contract is genuinely unchanged.

### Control-flow / contract (verified against source, not narrative)
- The new block at `listener.rs:2352` (`if maybe_output.is_none()`) is **purely additive**: it sits after the `if let Some(ref output) = maybe_output {…}` persisting block and before the unconditional `HookResponse::Ack` at 2360. No early return, no altered dispatch, no signature change. Response is `Ack` on every path exactly as before.
- Predicate is the exact logical negation of the persisting block's `Some` guard, so the trace fires iff Steps 3/4 (SESSIONS update, SIGNAL_QUEUE write, consumers) were skipped. F2 framing ("for the closing session") is accurate — it does not couple to stale-sweep state. Correct.

### OWASP
- **Injection**: none. Python smoke uses a fixed literal `SELECT COUNT(*) FROM observations` — no string interpolation, no user-controlled SQL. No shell/subprocess, no network beyond the test's own hook socket.
- **Access control / deserialization / input validation**: unchanged. No new inputs cross a trust boundary; the change consumes an already-parsed `session_id: &str`.

### Logging / PII (re: vnc-038 #5089)
- `session_id` is **already logged at `info!`** in this same function (`listener.rs:2223`, `2268`) and throughout the file. This change logs it at `debug!` — strictly **less** exposed than existing points, and gated behind the opt-in `unimatrix_server::obs` target. Not secret/PII-classified. No new leak.
- **Log-injection**: `session_id = %session_id` is a structured field; a hostile session_id with embedded newlines is a **pre-existing** consideration shared by the info! points above — not introduced or widened here. Non-blocking.

### Dependencies / secrets
- **Zero** dependency changes (no Cargo.toml/Cargo.lock diff). `tracing-test = "0.2"` already a dev-dep; `sqlite3` is Python stdlib. No hardcoded credentials/tokens/keys.

### Blast radius (worst case)
- If the predicate were subtly wrong, worst case is a **false or missing debug line under an opt-in target** — no data path, no response change, no lock/IO on the hot path (the `debug!` is level-gated and holds no registry lock). Cannot corrupt data, DoS, or disclose. Bounded to noise.

### Regression risk
- Near-zero. ADR #3467 obs-target silencing contract is **preserved** — the target matches the 5 existing obs points (1370/1383/1444/1694/1714). Rust controls assert both the 204-Ack idempotency and the emit/no-emit split.

### Non-blocking notes
1. Incidental rustfmt churn on `test_looks_like_feature_id_distinguishes_fragments` (`listener.rs:~7356`) is unrelated to #819. Formatting-only, test code, harmless — already flagged as the PR's non-blocking WARN. Fine to keep or drop.

**Verdict: approve. No changes required before merge.**

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. The last-mile I flagged at fix-approach (the *automated* blind-spot closure, not just the human-facing trace) is in the diff. Optional: revert the incidental fmt churn to keep the diff surgical; budget a CI confirmation of the integration-bin link that OOMd locally.

### Vision / roadmap fit
Bugfix under `goal:self-learning` (secondary) / `goal:personal-cloud`, advancing the operability capability the issue reframed to: **N4 — a healthy deployment emits no misleading signals** (#5162, partial). This does not change N4s status — it *advances* it by removing one silent-success blind spot (a 204 close indistinguishable in logs from a persisting one) **without adding a false alarm**. The `debug!`-not-`warn!` call is the load-bearing product decision and it holds: a `warn!` here would fire on every absent/duplicate close and directly invert the very goal the issue is filed under. C11 (remote observe over HTTPS, proven) is untouched — the 204 idempotency contract is preserved.

### Approach commentary (delivery as a whole, incl. security review)
- **The outcome is actually delivered, not just the mechanism.** At fix-approach I flagged the risk that only the human-readable trace would ship, leaving the *automated* gap (the `status == 204`-only smoke) that let #818 hide "for the gates whole lifetime" still open. The diff closes it: `test_observe_row_count_delta_persist_vs_noop_close` asserts the `observations` store row-count delta (0 on the no-op close, +1 on a persisting observe) over the live UDS+hook path. That is the discriminating assertion Gate 7 lacked, and it is non-vacuous (the same COUNT(*) measured 0 vs 1). This is the deliverable that matters; the obs-target `debug!` is the correct human-facing adjunct.
- **Contract preservation verified in source**, not just narrative: the new `if maybe_output.is_none()` block (listener.rs:2352) is purely additive — after the persisting block, before the unconditional `HookResponse::Ack`. No early return, no signature change, no `observe.rs`/`Ack`→204 edit. Matches F2 (single clean predicate) and F1 (`target: "unimatrix_server::obs"` per binding ADR #3467, matching the 5 existing obs points).
- **Security review (819-security-reviewer): LOW, no blocking findings — concur.** Its claims check out against source: additive control flow, `session_id` already logged at `info!` in the same function so this `debug!` is strictly *less* exposure, zero dependency/Cargo diff, fixed-literal SQL in the smoke (no injection surface), worst case is a stray debug line under an opt-in target. Nothing to re-run.

### Merge-readiness caveats (human to weigh — none blocking)
1. **Full `cargo test --workspace` never ran green locally** — linker OOM (`ld` signal 9) at the *link step* on integration bins (`import/export/project_routing/graph_subgraph`), a standing environmental memory ceiling, not caused by this ~15-line change (those bins dont touch the observe path). Coverage was re-established via `--lib` (6130/0 green) + the live Python smoke. Reasonable, but note: per this repos setup, GH CI is JS-client-only, so the Rust integration-bin link/run gets no automated cross-check either. Confidence is high given the changes tiny, well-isolated surface, but the full-workspace integration link is unconfirmed anywhere.
2. **Incidental rustfmt churn** on `test_looks_like_feature_id_distinguishes_fragments` (unrelated to #819) rode along — already flagged as the gates non-blocking WARN and by the security review. Cosmetic. This is exactly the "revert pure-fmt churn before wave commits" hygiene item; trivially revertable to keep the diff to the fix. Your call — harmless to keep.
3. **Smoke timing**: the no-op assertion uses a fixed 3s settle then asserts zero delta; safe here because an absent-session close genuinely has no write to race. The persist path settle-polls to 10s. Low flake risk; noted only for awareness.

No open questions remain. The two real decisions were resolved correctly upstream and hold in the shipped code: **trace-not-counter** (no metrics infra exists; an `AtomicU64` on a hot handler would be gold-plating) and **`debug!` on the obs target** (protects N4). The one right-sizing question I left open at fix-approach — is the automated row-count assertion folded in or deferred? — is answered by the diff: **folded in**. That was the correct call.

### Recommended actions
1. **Merge.** The outcome (a #818-class silent drop is now *automatically* detectable) is delivered, the contract is preserved, security is LOW/clean, and the gate is PASS.
2. Optionally **revert the fmt churn** on `test_looks_like_feature_id_distinguishes_fragments` first — surgical diff, no downside.
3. **Budget a post-merge CI/full-workspace confirmation** in a higher-memory environment, since the integration-bin link never completed locally (low risk, but currently unverified anywhere).

Issue #879, opened 2026-07-02
fix: compaction tick deletes past-ceiling inbound edges instead of repointing (silent link loss on hot-node correction)


dug-21:
## Summary

`run_orphaned_edge_compaction` (`crates/unimatrix-server/.../background.rs:832`) DELETEs any `graph_edges` row whose target is not `Active`. A corrected entry becomes Deprecated (not Active), so any inbound edge the synchronous redirect did **not** reach is physically deleted at the next tick rather than repointed to the terminal — **silent inbound link loss.**

**This is the backstop half of a two-part fix.** The primary fix (#882) raises the inbound redirect ceiling so realistic hot-node corrections complete synchronously. But edges still reach the compaction tick from three sources a raised ceiling cannot prevent:
- **per-edge redirect failures** (F2 — a single edge's SQL write errors under the ceiling, warn-and-continue, `tools.rs:5312`),
- **the existing corpus** (nodes corrected under the old 50 ceiling before #882 lands),
- **races** (an edge written concurrent with a correction).

Today the tick **deletes** all of these instead of repointing them. That is a data-loss bug independent of the ceiling value — so this fix stands whatever #882 chooses.

## Root cause

The compaction tick is a garbage collector: correctly, it removes edges pointing at genuinely dead targets (quarantined entries, dead-end chains with no successor). The bug is that it does **not distinguish** two cases that both read as "target not Active":
- target Deprecated **with** a `superseded_by` successor → a valid terminal exists → **recoverable, must not delete**
- target dead **with no** successor → nothing to repoint to → **correctly deleted**

It treats both as garbage.

## Fix — repoint-then-compact (same pass)

1. **Repoint** (before delete): for each non-`Supersedes` edge whose target is Deprecated *with* `superseded_by` set, repoint to the CTE-resolved terminal (#4468 — SQL recursive CTE, never in-memory `find_terminal_active`). Reuse `redirect_graph_edge` (idempotent via `UNIQUE(source_id, target_id, relation_type)`).
2. **Compact** (existing behavior): delete only the true residual — target/source non-Active *and no valid terminal* (quarantined, dead-end chains).

Ordering already sits before `typed_graph_rebuild` in the job registry, so PPR/in-memory picks up repointed edges same-tick. Edges are a standalone table (no hash-chain participation, principle #1 not engaged); edge mutations are not audited today, so no audit change required.

## Batch budget `B` — bounded work (in scope, lightweight here)

[...]

Issue #870, opened 2026-07-01
research(ass-088): edge-consistency under correction — reconcile #744, #745, vnc-042/NG-1


dug-21:
## Question
When `context_correct` deprecates entry X in favor of terminal X′, the edges touching X must eventually point at X′ for the typed graph to stay integrity-consistent (architectural principle #4). Should that consistency be enforced at **write-time** (synchronous redirect, ceiling-bounded), at **read-time** (per-read resolution tax), or via **background-tick convergence** — and which surfaces need which?

## Why it matters
Three separately-tracked issues are faces of one problem:
- **#744 (inbound orphans)** — `run_redirect_loop` caps at `REDIRECT_CEILING = 50` and truncates-with-`warn!` past it; correcting a hot node silently leaves referrers 51+ pointing at the deprecated entry.
- **#745 (outbound orphans, historical)** — pre-vnc-035 corrections dropped outgoing edges; a one-shot repair sweep is proposed.
- **vnc-042 / NG-1 (stale neighbor labels)** — a resolved `context_get` returns the terminal's edge list but the edge *targets* still render the deprecated id+title.

vnc-042 (#843) proved the read-side **get entry** leg — `context_get` resolves to the active terminal by default. #744/#745 are the write-side orphan legs; NG-1 is the seam (resolved entry, stale edge labels). This spike picks the enforcement model that closes the remaining legs — in particular the read-side edge-target resolution (NG-1) that keeps the "get returns current" capability at `partial`.

## Scope doc
`product/research/ass-088/SCOPE.md` (already on disk)

## Depends on
- vnc-042 / #843 (read-side get-entry resolution — the proven leg this builds on)
- Relates to #744, #745

dug-21:
## Findings: ass-088 — edge-consistency under correction

Spike complete. Full findings: `product/research/ass-088/FINDINGS.md` (directional; investigation + design). No code, no PR, no Unimatrix writes.

**Hypothesis (write cheap · read resolve · tick converge) holds — with two load-bearing corrections:**

1. **The residual is deletion, not a stale pointer.** `run_orphaned_edge_compaction` (`background.rs:832`) DELETEs any edge whose target isn't `Active`. So #744's past-ceiling referrers don't dangle — they **lose their link entirely** at the next tick. Convergence must **repoint before compaction deletes**.
[...]

PR #884, opened 2026-07-02
fix: raise REDIRECT_CEILING 50->500 to stop inbound-redirect truncation (#882)

### LINKED ISSUE #882: fix: raise REDIRECT_CEILING (inbound redirect cap) from 50 to a personal-cloud level (~250)
## Summary

`context_correct`'s synchronous inbound redirect (`run_redirect_loop`, `tools.rs:5207`) caps at `REDIRECT_CEILING = 50` (`tools.rs:45`, applied `tools.rs:5234-5249`). Correcting a node with more than 50 inbound referrers redirects only the first 50 and `warn!`-truncates the rest — those referrers are then deleted at the next compaction tick (see #879). Raising the ceiling to a level that covers the **personal-cloud** working set makes hot-node corrections complete synchronously and closes the truncation entirely for realistic fan-in.

This is the **primary** fix for the lost-edge problem, chosen over building an enterprise-scale convergence tick. Rationale below.

## Why raise, not build a convergence tick

At personal-cloud scale a synchronous full redirect is cheap. `redirect_graph_edge` is DELETE + INSERT OR IGNORE per edge inside the correction transaction. Even the hottest observed node (in-degree 110) is ~220 row ops — low-single-digit ms, on an already-heavyweight, rare operation. The 50 ceiling defends against a cost that isn't real at this scale.

The amortized background convergence tick only earns its complexity (drain windows, read-cover dependency, bounded-batch logic) at enterprise scale, where a hot node reaches thousands of referrers and a synchronous redirect would hold the per-slug single-writer pool for seconds. That regime is **not** the current reality and is explicitly deferred (tracked in the ass-088 ADR / #870).

## Number

- Current max in-degree in the snapshot corpus (~4k entries): **110**. Only 5 nodes exceed 50.
- Proposed: **~250** — >2x headroom over today's hottest node, covers personal-cloud growth, and a 250-edge synchronous redirect is still ~ms.
- Exact value TBD; **250** is the working default. Consider making it **config-backed** (not hardcoded) so an enterprise deployment can raise it without a code change — optional, not required for this fix.

## Scope

- Raise `REDIRECT_CEILING` (constant, or promote to config).
- Keep the existing `warn!`-on-truncation behavior for the now-much-rarer over-ceiling case — it is the fail-loud signal that a node has exceeded even the raised bound (and that #879's backstop will catch the remainder).
- Does **not** remove the ceiling entirely — an unbounded synchronous redirect reintroduces the enterprise write-latency cliff. The ceiling stays as a guard; it just moves to a level real usage doesn't hit.

## Relationship to #879 and #870

- **#879 (compaction repoint-before-delete)** — the **backstop**. Even with a raised ceiling, edges still reach the tick from per-edge redirect failures, the pre-existing corpus, and races. The tick must repoint-not-delete those. The two ship together: raised ceiling = primary, tick repoint = safety net.
- **#870 (ass-088 research)** — the enterprise convergence-tick design is deferred there; this issue is the personal-cloud-appropriate answer that closes the live defect now.

dug-21:
Closes #882.

## Root cause
`REDIRECT_CEILING = 50` (`crates/unimatrix-server/src/mcp/tools.rs`) truncated the synchronous inbound redirect in `run_redirect_loop`. When a corrected node had >50 inbound referrers, edges 51+ stayed pointed at the now-Deprecated original and were later **hard-deleted** by orphaned-edge compaction — permanent silent link loss. A stale capacity constant (ADR-004, sized when max in-degree was single-digit); observed corpus max in-degree is now 110.

## Fix
Raise the constant `50 -> 500` (human-approved **raise-only, non-config const**). This gives order-of-magnitude headroom over the observed corpus max (110) and covers personal-cloud fan-in. Config-backing was rejected by investigator, architect, and product review: a knob invites cranking toward the single-writer write-latency cliff the ceiling guards, and the enterprise answer is a deferred async convergence tick (#870), not a bigger synchronous knob. The fail-loud `warn!`/truncation guard is preserved for the now-far-rarer over-500 case.

Single file changed: `crates/unimatrix-server/src/mcp/tools.rs` (constant + doc comment + test rewrites).

## Tests
- Ceiling tests rewritten to assert against `REDIRECT_CEILING` symbolically (no hardcoded literals).
- New regression `test_redirect_loop_corpus_max_in_degree_no_truncation`: corrects a node with in-degree 110 and asserts `truncated == false` with zero edges left on the Deprecated original — would have caught the original bug.
- Full workspace: 6735 passed / 0 failed. Clippy clean. Integration smoke: 28 passed.

## Related
- **#879 (repoint-before-delete) is the load-bearing integrity backstop and must still ship.** This raise is primary-but-partial: it shrinks but does not close the silent-loss window (per-edge redirect failures, pre-existing corpus edges, and any >500 tail still reach the compaction DELETE). Do not close silent-link-loss on #882 alone.
- ADR-004 amended to reflect the 500 ceiling.
- Enterprise convergence-tick design deferred: #870 (ass-088).

### DIFF
--- a/crates/unimatrix-server/src/mcp/tools.rs
+++ b/crates/unimatrix-server/src/mcp/tools.rs
@@ -40,9 +40,20 @@ use crate::uds::hook::MAX_GOAL_BYTES;
 
 /// Maximum incoming edges to auto-redirect per context_correct call (SR-01 ceiling).
 /// Entries with more than this many incoming edges emit tracing::warn! and redirect
-/// only the first REDIRECT_CEILING rows. See ADR-004 vnc-017.
+/// only the first REDIRECT_CEILING rows. See ADR-004 vnc-017 (amended for #882).
+///
+/// Sized for personal-cloud fan-in. The original N=50 (ADR-004) was set when the max
+/// observed in-degree was single-digit; the corpus has since grown to an observed max
+/// of 110 (ass-087), so 50 truncated real corrections — and truncated inbound edges are
+/// later hard-deleted by orphaned-edge compaction, causing silent link loss (#882).
+/// 500 restores order-of-magnitude headroom over the observed max while keeping the
+/// ceiling as a bounded write-latency guard on the synchronous correction path
+/// (~1 s worst case per correction, a rare heavyweight op). The fail-loud warn! +
+/// truncation response text below still catch any future over-500 fan-in. Enterprise
+/// scale is deferred to an amortized async convergence tick (#870), not a larger
+/// synchronous knob or a config-backed value — the ceiling stays a plain `const`.
 // pub(super) so the sibling test module can reference it via use super::REDIRECT_CEILING.
-pub(super) const REDIRECT_CEILING: usize = 50;
+pub(super) const REDIRECT_CEILING: usize = 500;
 
 /// Tool description for context_get (vnc-042).
 ///
@@ -11533,16 +11544,21 @@ mod redirect_loop_tests {
         );
     }
 
-    // ---- R-05: Fan-in ceiling (55 edges → truncate at 50) ----
+    // ---- R-05: Fan-in ceiling (REDIRECT_CEILING + 5 edges → truncate at ceiling) ----
 
-    /// R-05: When 55 incoming edges exist, exactly 50 are redirected; 5 remain at original.
+    /// R-05: When `REDIRECT_CEILING + 5` incoming edges exist, exactly `REDIRECT_CEILING`
+    /// are redirected and 5 remain at original. Asserted symbolically against the constant
+    /// so it survives any resize of `REDIRECT_CEILING` without re-pinning to a literal.
     #[tokio::test]
-    async fn test_redirect_loop_ceiling_truncates_at_50_and_warns() {
-        let dir = tempfile::TempDir::new().expect("tempdir");
-        let (store, original_id) = open_store_and_insert_active(&dir, "r05-55").await;
+    async fn test_redirect_loop_ceiling_truncates_and_warns() {
+        const OVERFLOW: u64 = 5;
+        let total: u64 = REDIRECT_CEILING as u64 + OVERFLOW;
 
-        // Insert 55 source entries (all Active) and edges to original.
-        for i in 0u64..55 {
+        let dir = tempfile::TempDir::new().expect("tempdir");
+        let (store, original_id) = open_store_and_insert_active(&dir, "r05-over").await;
+
+        // Insert `REDIRECT_CEILING + OVERFLOW` source entries (all Active) and edges to original.
+        for i in 0u64..total {
             let src_id = store
                 .insert(
                     unimatrix_store::test_helpers::TestEntry::new(
@@ -11563,15 +11579,21 @@ mod redirect_loop_tests {
             .await
             .expect("expected Some(RedirectSummary)");
         assert_eq!(
-            rs.found, 50,
-            "found must equal REDIRECT_CEILING (50 after cap)"
+            rs.found, REDIRECT_CEILING,
+            "found must equal REDIRECT_CEILING after cap"
         );
-        assert_eq!(rs.redirected, 50, "exactly 50 edges must be redirected");
+        assert_eq!(
+            rs.redirected, REDIRECT_CEILING,
+            "exactly REDIRECT_CEILING edges must be redirected"
+        );
         assert_eq!(rs.failed, 0);
         assert!(rs.truncated, "truncated must be true");
-        assert_eq!(rs.total_raw, 55, "total_raw must be 55 (pre-cap count)");
+        assert_eq!(
+            rs.total_raw, total as usize,
+            "total_raw must be REDIRECT_CEILING + OVERFLOW (pre-cap count)"
+        );
 
-        // Exactly 50 edges at new_id; 5 edges still at original_id.
+        // Exactly REDIRECT_CEILING edges at new_id; OVERFLOW edges still at original_id.
         let redirected_count = sqlx::query_scalar::<_, i64>(
             "SELECT COUNT(*) FROM graph_edges WHERE target_id = ?1 AND relation_type != 'Supersedes'",
         )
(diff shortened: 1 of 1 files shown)


dug-21:
## Security Review (fresh context) — 882-security-reviewer

Reviewed `git diff main...HEAD` cold: single file (`crates/unimatrix-server/src/mcp/tools.rs`), `REDIRECT_CEILING` 50→500 + doc comment + test rewrites. Verified against the live `run_redirect_loop` and `query_incoming_edges` implementations.

### OWASP / boundary
- **Injection:** none. `REDIRECT_CEILING` is an internal `pub(super) const`, not config-backed and not reachable from operator/attacker input. `query_incoming_edges` (store/read.rs:1694) binds `target_id` as a parameter and the SQL is untouched. No new inputs, no new deserialization.
- **Access control:** unchanged. Redirect runs within the correcting slug's own per-slug pool; no cross-slug boundary is touched.
- **Input validation:** the fail-loud guard is fully intact and unchanged — `truncated = total_raw > REDIRECT_CEILING`, the `warn!`, and the `&incoming[..REDIRECT_CEILING]` slice (tools.rs:5245-5263). The slice is only taken when `total_raw > REDIRECT_CEILING`, so it cannot panic. Over-500 fan-in still warns + truncates + reports.

### Blast radius (worst case)
Raising the cap widens the synchronous per-edge write loop from ≤50 to ≤500 iterations, each a single-writer RAII transaction on the per-slug pool. Worst case ≈1s of writer occupancy on a rare heavyweight `context_correct` against a >100-fan-in hub; other writes to that same slug queue behind it. This is bounded, self-inflicted, and slug-local — no cross-tenant amplification. `query_incoming_edges` already fetched all rows unbounded before capping (true at 50 as well), so this PR introduces no new memory behavior. Enterprise scale is correctly deferred to the async convergence tick (#870) rather than a bigger synchronous knob. Acceptable.

Note (pre-existing, not introduced here): the redirect loop is post-commit and per-edge non-atomic; a crash mid-loop leaves a partial redirect. Raising N enlarges that window, but each `redirect_graph_edge` is idempotent (UNIQUE-conflict ignore) so re-running converges. Not a new risk class.

### Regression
Low. Guard logic byte-identical. Tests strengthened: ceiling assertions rewritten symbolically against the constant (no literals to drift), plus a new `test_redirect_loop_corpus_max_in_degree_no_truncation` (in-degree 110 → `truncated == false`, zero edges left on the Deprecated original) that would have caught the original bug. The compile-time `const { assert!(REDIRECT_CEILING > OBSERVED_CORPUS_MAX_IN_DEGREE) }` invariant is a good regression pin.

### Dependencies / secrets
No new dependencies. No hardcoded credentials/tokens/keys.

**Verdict: approve from a security standpoint.** Minimal, correctly scoped, guard preserved.

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. Clean, minimal, honestly-scoped raise-only fix. One condition: keep #879 OPEN and load-bearing — this PR shrinks but does not close silent link loss, and the PR/gate language already says so.

### Vision / roadmap fit
Carries `goal:self-learning` (#5219) and defends Architectural Principle #4 (typed relationship graph — edges explicit, typed, traversable). Silent inbound-edge loss on correction degrades exactly the asset a self-learning engine is built to accumulate, so closing the truncation path is in-lane and timely: the corpus max in-degree (110, ass-087) has already outgrown the original 50 sizing — a live defect, not hypothetical. The rejection of a config knob and of building the enterprise async convergence tick now is the right vision call: Principle #6 (single binary, zero required infra) and the personal-cloud scale target make a synchronous full redirect cheap; the amortized tick stays deferred to #870/ass-088. Consistent with the standing caution against over-elevating defensive structure — a config knob would add an input-validation surface and invite cranking the synchronous path toward the write-latency cliff the ceiling guards.

### Approach commentary
- Raise-only const matches ADR-004's own escape hatch and is the minimal shape: one constant + doc comment + test rewrites, single file, no signature/schema/response changes. Blast radius contained — `REDIRECT_CEILING` is referenced only by `run_redirect_loop` and its test module; the vnc-035 carry-forward loop is unceilinged and untouched.
- Fail-loud guard preserved: `warn!` + `(truncated from N, see logs)` still fire for any future >500 fan-in.
- Tests parameterized symbolically on the constant (no re-pin to a literal) and the structural test asserts `REDIRECT_CEILING > 110` as an invariant rather than pinning 500 — so there is no ADR/test merge-atomicity trap, and a future resize won't silently break or re-pin. The new `test_redirect_loop_corpus_max_in_degree_no_truncation` (110-edge, asserts `truncated == false`, zero edges left on the Deprecated original) reproduces the exact case the old ceiling lost. This is the regression that matters.
- Value = 500, not the issue-title's ~250. This is the design-reviewer's explicit fallback ("500 if #879 may decouple/slip") and the human-approved number. Since #879 is a *separate open issue*, not a committed co-requisite pair, 500 (order-of-magnitude headroom over 110) is the defensible choice — 250 gave only ~2.3x runway in a corpus that demonstrably grows. Endorsed.

No standalone uni-security-reviewer was spawned; for a raise-only internal-constant bugfix that is appropriate, and the design review (section 5) assessed the security surface directly: `REDIRECT_CEILING` is an internal const, not operator/attacker input — no new trust boundary, no input-validation surface (a config knob *would* add one; const sidesteps it). No amplification vector — building a >500 in-degree hub needs >500 prior authenticated writes within one's own slug (1-client:1-project), and one correction then costs a bounded, slug-local ~1s. Diff contains no `unsafe`/`todo!()`/`unimplemented!()`/new production `.unwrap()` (gate check 6 confirms). I concur: no security concern introduced.

### Delivery-as-a-whole
Gate PASS across all seven checks. Verification foreground: workspace 6735 passed/0 failed, clippy `-D warnings` clean, smoke 28 passed, correction-chain/edge suites 100 passed (6 pre-existing xfail / 1 pre-existing xpass, all ONNX/tick-interval/GH#406, none touching the ceiling). Redirect suite runtime acceptable (~0.67s despite the 505-row insert). ADR-004 amended via `context_correct` → #5408. Knowledge stewardship blocks present in every phase comment (correctly "nothing novel" — bugs are GH issues, not lessons). No merge-readiness gaps.

### The one product-lens item to hold: DoD altitude (primary-but-partial)
This PR delivers a *mechanism* (fewer truncations), not the full *outcome* (no silent link loss). Per-edge redirect failures (warn-and-continue, ADR-003), pre-existing stale corpus edges, and races still reach the compaction DELETE — #879's repoint-before-delete is what actually converts loss into recoverable degraded state. The delivery handled this correctly: the PR body, fix-applied comment, and gate notes all state #879 remains load-bearing and #882 must not be treated as the complete fix. #879 is confirmed OPEN with `goal:self-learning`. The failure mode to avoid post-merge is #879 being deprioritized once the visible symptom (truncation on hot nodes) disappears — recommend the merge/close language on #882 explicitly points at #879 as the open integrity backstop so no false confidence is created.

- **Merge this PR?** Yes, as-is. Minimal, correct, green, honestly framed. No rework.
- **Is 500 (vs title's ~250) acceptable?** Yes. It is the reviewer's decoupled-#879 fallback and the human-approved value; order-of-magnitude headroom is the right call while #879 ships separately rather than as a committed pair.
- **Does closing #882 close silent link loss?** No. Keep #879 open and load-bearing. #882 is primary-but-partial by design.

### Recommended actions
1. Merge #884.
2. On #882 closure, reference #879 as the still-open, load-bearing integrity backstop (do not let the closure imply silent link loss is resolved).
3. Keep #879 prioritized — the visible symptom disappearing must not drop its priority.
4. No follow-up issue needed from this review; the deferrals (#879 backstop, #870 enterprise tick) are already tracked.

PR #885, opened 2026-07-02
fix: default-resolve context_graph supersessions across read surfaces (#881)

### LINKED ISSUE #881: fix: context_graph resolve_supersessions defaults OFF while context_get defaults ON — briefing/gating consumers see deprecated endpoints
## Summary

`context_graph`'s `resolve_supersessions` parameter defaults to **`false`** (`graph_read_subgraph.rs:158`, `graph_read_neighbors.rs:164`), while `context_get` resolves the anchor to its terminal by **default-ON** (vnc-042). The two read surfaces are inconsistent.

Consequence: any `context_graph` caller that omits the flag — traversal, briefing/injection edge-walk, or a gating decision — sees **Deprecated endpoints/neighbors** with stale titles, even though the corresponding `context_get` would resolve them. The graph read path silently under-resolves relative to the single-entry read path.

## Why this is a standalone product-behavior defect

Independent of the broader edge-consistency ADR (ass-088 / #870) and of stored-graph convergence. Even with the stored graph fully converged, a default-OFF resolve means consumers observe deprecated references during the interim window (and, for the label-rendering path, at any time). It is a read-surface consistency gap, not a corpus-hygiene problem.

Scope here is **going-forward product behavior** — align the read surfaces.

## Fix

Flip `context_graph` `resolve_supersessions` to **default-on**, matching `context_get`'s vnc-042 default — at minimum for the briefing/injection and any gating consumers. Keep an explicit opt-out for callers that genuinely want the raw stored graph (audit/debug).

Related read-label gap (NG-1): a resolved `context_get` still renders edge-*target* labels raw in `build_edges_view` (`get_edges.rs:44`) with no supersession resolution. These two form the "read cover completion" workstream in the ass-088 findings and can ship together.

## Evidence
- `resolve_supersessions` default-false: `graph_read_subgraph.rs:158`, `graph_read_neighbors.rs:164`
- `context_get` anchor resolve default-on: vnc-042
- Edge-target label gap (NG-1): `build_edges_view` (`get_edges.rs:44-101`)

## Relationship to #870 (ass-088)
Surfaced as an Out-of-Scope Discovery in `product/research/ass-088/FINDINGS.md` ("worth aligning independently"). The convergence ADR bundles the read-cover completion, but this default alignment is a small fix that stands on its own.

dug-21:
Fixes #881.

## Root cause
`context_graph`'s `resolve_supersessions` defaulted **OFF** while `context_get` resolves anchors **ON** (vnc-042). The only real consumers are **external agent callers of `context_graph`** (e.g. the `uni-capability` skill) — `context_briefing` and context injection are immune (they use `SearchService` `RetrievalMode::Strict`, not the graph path), and no server-side gating consumer exists.

Investigation surfaced two defects beyond the reported default mismatch:
- **G1 (load-bearing):** the depth-1 neighbors arm (the default depth) dispatched to `neighbors_sql`, which never consulted the flag — so even an explicit `resolve_supersessions: true` was ignored there. Flipping the default alone would have been a **no-op** for the flagship consumer.
- **N3:** subgraph resolved the neighbor *node* but recorded the edge with the *raw* target, so the dangling-edge filter dropped it, leaving the terminal as a disconnected node.

## Fix
1. Flip `resolve_supersessions` default OFF→ON in `graph_read_subgraph.rs`, `graph_read_neighbors.rs`, `graph_read_path.rs`. `false` stays as an explicit audit/raw opt-out.
2. **G1:** depth-1 `neighbors_sql` now resolves endpoints via a per-request memo (`resolve_memoized`) — bounds reads on hot nodes (in-degree ~110), one resolve per distinct endpoint.
3. **NG-1:** `build_edges_view` resolves the ≤3 displayed edge targets, substituting only when deprecated (byte-identity canary preserved).
4. **N3:** both subgraph edge-recording paths use resolved endpoints.
5. Docs: both `CONTEXT_GRAPH_DESCRIPTION` copies + `graph_read.rs` field doc now document the flag for all modes.
6. **Atomic ADR correction:** `context_correct` vnc-042 ADR #5388 → **#5409** — retracts only the graph flip-direction prohibition, preserves the per-intent principle, keeps the reverse prohibition (context_get stays default-resolve), records N2 (silent dead-end resolution + `uni-capability` gating risk) and N3.

## Tests
New cross-surface consistency test asserts `context_get`, neighbors (at **default depth 1** — the G1 catch), subgraph, and path all surface the terminal by default, and that explicit `resolve_supersessions: false` still returns raw. Plus NG-1 deprecated-target and escape-hatch tests.

- 4 new unit tests pass; `cargo test --workspace` rc=0; `cargo clippy --workspace -- -D warnings` rc=0
- Integration smoke 28 passed; graph/context-read integration 114 passed (6 pre-existing xfails, none touched)

## Post-merge follow-up
Vision session must correct SL7 (#5390) to drop the now-false "graph stays default-off" non-goal.

### DIFF
--- a/crates/unimatrix-server/src/mcp/get_edges.rs
+++ b/crates/unimatrix-server/src/mcp/get_edges.rs
@@ -41,7 +41,23 @@ use super::response::edges::{
 ///
 /// Returns `Result<EdgesView, StoreError>` so the handler maps the whole assembly once
 /// with the identical primary-read mapping (FR-19). No `.unwrap()`/`.expect()`.
-pub(crate) async fn build_edges_view(store: &Store, id: u64) -> Result<EdgesView, StoreError> {
+///
+/// NG-1 (bugfix-881 / ass-088 F5): when `resolve_targets` is true (the default — mirrors
+/// `follow_supersessions`, so the `follow_supersessions=false` escape hatch keeps raw
+/// targets), each of the ≤`GET_EDGE_DISPLAY_LIMIT` displayed edge TARGETS is resolved to its
+/// active terminal via `follow_to_current`. Substitution happens ONLY when a target is
+/// actually deprecated (`terminal != raw`); an all-active response is therefore byte-identical
+/// to the pre-resolution shape, preserving the `test_none_json_byte_identical_to_base_object`
+/// canary. Dead-end / cycle / 50-hop-cap targets keep their raw id (same `None -> raw`
+/// fallback as context_get and the BFS handlers). N5: `totals`/`authored_total` stay computed
+/// from the uncapped STORED edge set (unresolved) — they may exceed the distinct displayed
+/// targets after resolve+dedup; this is the intended honest-stored-totals design (vnc-037),
+/// NOT a counting bug.
+pub(crate) async fn build_edges_view(
+    store: &Store,
+    id: u64,
+    resolve_targets: bool,
+) -> Result<EdgesView, StoreError> {
     let pool = store.read_pool_server();
 
     // 1. ranked ≤cap displayed rows (canonicalized, LEFT JOIN confidence, LIMIT cap).
@@ -50,11 +66,30 @@ pub(crate) async fn build_edges_view(store: &Store, id: u64) -> Result<EdgesView
     // 2. honest uncapped split totals — three buckets + digest-only `authored`.
     let split: EdgeCountSplit = count_neighbors_split(pool, id).await?;
 
-    // 3. batched title join over the ≤cap displayed targets ONLY (never the uncapped set).
+    // 3. NG-1 resolve the ≤cap displayed edge-target labels to their active terminal. Bounded
+    //    to the ≤GET_EDGE_DISPLAY_LIMIT displayed targets (≤3 chain walks — N1 fine as-is).
+    //    resolution maps raw_target -> effective_target; entries only differ when a target is
+    //    actually deprecated (terminal != raw), so all-active responses stay byte-identical.
+    let mut resolution: HashMap<u64, u64> = HashMap::new();
+    if resolve_targets {
+        for edge in &rows {
+            let raw = edge.row.target_id;
+            if let std::collections::hash_map::Entry::Vacant(slot) = resolution.entry(raw) {
+                let terminal = crate::mcp::graph_read::follow_to_current(store, raw)
+                    .await
+                    .unwrap_or(raw);
+                slot.insert(terminal);
+            }
+        }
+    }
+    let effective = |raw: u64| resolution.get(&raw).copied().unwrap_or(raw);
+
+    // 4. batched title join over the ≤cap displayed (effective) targets ONLY.
     let mut target_ids: Vec<u64> = Vec::with_capacity(rows.len());
     for edge in &rows {
-        if !target_ids.contains(&edge.row.target_id) {
-            target_ids.push(edge.row.target_id);
+        let t = effective(edge.row.target_id);
+        if !target_ids.contains(&t) {
+            target_ids.push(t);
         }
     }
     let title_map = if target_ids.is_empty() {
@@ -63,13 +98,13 @@ pub(crate) async fn build_edges_view(store: &Store, id: u64) -> Result<EdgesView
         fetch_titles_batch(pool, &target_ids).await?
     };
 
-    // 4. project ranked rows → GetEdge.
+    // 5. project ranked rows → GetEdge (target_id substituted to the effective terminal).
     let edges: Vec<GetEdge> = rows
         .iter()
-        .map(|edge| project_edge(edge, &title_map))
+        .map(|edge| project_edge(edge, effective(edge.row.target_id), &title_map))
         .collect();
 
-    // 5. assemble (edges already ≤cap from SQL; totals uncapped, three buckets;
+    // 6. assemble (edges already ≤cap from SQL; totals uncapped, three buckets;
     //    authored_total threaded for the digest from the FULL uncapped set).
     Ok(EdgesView {
         edges,
@@ -84,13 +119,13 @@ pub(crate) async fn build_edges_view(store: &Store, id: u64) -> Result<EdgesView
 
 /// Project one ranked row into the discovery-list [`GetEdge`].
 ///
-/// `target_id` is the OTHER endpoint (`RankedEdge.row.target_id`); `target_title` is the
-/// title-map lookup (`None` ⇒ dangling, retained — DNB-1); `authored` is derived in
-/// [`GetEdge::new`] from the raw `source` string (`== "agent"`). The SQL-computed
+/// `target_id` is the effective OTHER endpoint (the resolved terminal when the stored target
+/// is deprecated — NG-1; otherwise the raw `RankedEdge.row.target_id`); `target_title` is the
+/// title-map lookup for that effective id (`None` ⇒ dangling, retained — DNB-1); `authored` is
+/// derived in [`GetEdge::new`] from the raw `source` string (`== "agent"`). The SQL-computed
 /// direction string is mapped to the canonical `&'static str` constant — a `↔` row is
 /// never re-derived as `→`/`←` in Rust.
-fn project_edge(edge: &RankedEdge, title_map: &HashMap<u64, String>) -> GetEdge {
-    let target_id = edge.row.target_id;
+fn project_edge(edge: &RankedEdge, target_id: u64, title_map: &HashMap<u64, String>) -> GetEdge {
     GetEdge::new(
         edge.row.relation_type.clone(),
         map_direction(&edge.direction),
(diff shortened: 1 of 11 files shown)


dug-21 (commented):
## Security Review — bugfix-881-security-reviewer (fresh context)

**Risk level: LOW. No blocking findings.** Approving on security/blast-radius grounds; the default flip is an intentional, opt-out-preserving behavior change.

### OWASP / injection — clear
- **SQLi (the flagged surface):** verified every IN-list is positional-bind only. `fetch_titles_batch` (`get_edges.rs:158`) and `run_outgoing_query`/`run_incoming_query` (`graph_queries_neighbors.rs:30,71`) build `?N` placeholder tokens from `len()` and `.bind()` each value — no id or edge_type is ever string-interpolated. `edge_types` are additionally validated through `RelationType::from_str` before reaching SQL (`handle_neighbors` rejects unknown types and `Supersedes`). Anchor id binds positionally throughout.
- **Access control:** unchanged. `Requires Read capability` still gates all modes; `validate_no_unsupported_params` still rejects `resolve_supersessions` on chain/current/inverse/filter, so the new default is mode-scoped to neighbors/subgraph/path exactly as before. No trust-boundary movement.
- **Input validation:** the flag is a plain `Option<bool>` — no int/string coercion, no new external input. depth/direction/edge_type/max_nodes validation paths untouched.

### Blast radius — bounded, fail-safe
- **Cycles / self-supersession:** `follow_to_current` (`graph_read_neighbors.rs:36`) is hard-capped at `0..50` and returns `None` on cycle, dead-end, or store error → callers fall back to the raw id. A self-loop wastes ≤50 identical reads (cosmetic), never hangs.
- **Memo correctness (the "wrong terminal for a different id" worry):** `resolve_memoized` keys the HashMap on `raw_id` and stores `raw_id -> terminal`; each distinct endpoint resolves at most once and cross-id contamination is structurally impossible. In `neighbors_sql` the anchor is left as-stored and only the non-anchor endpoint is resolved, matching the BFS handler. For direction=Both the single memo correctly spans the concatenated outgoing+incoming rows.
- **subgraph edge-recording fix:** node set and recorded edge endpoints now both use `effective_neighbor`, so the post-BFS dangling filter no longer drops the terminal — verified against the filter at `graph_read_subgraph.rs:273-297`. Dedup still keys on the raw stored triple.
- **Worst case if a subtle bug slipped through:** a displayed id/label points at a stale-but-real entry (information *staleness*, never disclosure of unauthorized data — all reads are already Read-gated and same-project scoped). No write path, no data mutation, no privilege change. Failure mode is safe.

### Regression risk — medium-low, intentional, opt-out intact
- Behavior change is by design (issue #881). The explicit `resolve_supersessions: false` / `follow_supersessions: false` audit opt-out is preserved on every surface, and `build_edges_view` correctly gates on `params.follow_supersessions != Some(false)` so the as-stored escape hatch keeps raw targets consistent with the as-stored anchor.
- Node-collapse (multiple deprecated variants → one terminal) shrinks node/edge/displayed-row counts — expected, documented in-code.
- **N2 (silent dead-end resolution):** context_graph EdgeRecord carries no resolution-notice field, so an unresolvable-deprecated endpoint shows raw with no signal, unlike context_get's loud resolution. Accepted product bet, recorded in corrected ADR #5409. Not a security issue.

### DoS — mitigated
depth-1 `neighbors_sql` has no fan-out cap (pre-existing), but the per-request memo bounds added work to one `follow_to_current` per *distinct* endpoint (≤50 read-pool gets each, chains typically 1–2). No new unbounded amplification; read-pool only.

### Dependencies / secrets — clean
No Cargo.toml / Cargo.lock changes. No new deps (HashMap is std). No `unsafe`, no `Command`/process spawning, no env/secret handling, no hardcoded credentials, no `.unwrap()`/`.expect()` on non-test paths (only safe `.unwrap_or(raw)` fallbacks).

**Verdict: LOW risk, no blocking findings. Cleared to merge on security grounds.**

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. The delivery honored every fix-approach rider (atomic ADR correction, N1 memo, N2/N3 recorded, all three modes) and the security review clears it at LOW risk. One cosmetic WARN, non-blocking. The only load-bearing item is the **post-merge SL7 corpus correction** — without it the capability map contradicts shipped behavior.

### Vision / roadmap fit
Labeled `goal:self-learning`. Two legs, sitting differently against the capability map:

- **NG-1 (get-edges label resolution)** closes **SL7** (#5390) `done_when` clause 2 — "edge-target labels also resolve to current (NG-1 closed)", the open clause that kept SL7 `partial`. With clause 1 already proven (vnc-042) and clause 2 now delivered + behaviorally tested (`test_ng1_deprecated_target_resolves_to_terminal`, escape-hatch test), **SL7's defined scope is now fully met**. Capability-status implication (recommendation only, vision session owns the mutation): SL7 clause 2 → proven; SL7 as a whole → proven.

- **The context_graph default flip** is *outside* SL7 and directly contradicts SL7's recorded non-goal ("graph resolve_supersessions stays opt-in (default-off)… each is the intended default for its interface"). The PR correctly did NOT silently leave that contradiction: it corrected ADR #5388 → #5409 atomically in-commit (retracts only the flip-direction prohibition, keeps the per-intent analysis and the reverse prohibition). But **SL7's non-goal text still says the opposite** and lives in the capability corpus, not the ADR. This is the one thing that outlives the PR.

Orthogonal to SLN3 (#5230, stored-graph integrity) — nothing here touches stored edges or orphan handling. This is purely read-surface currency.

### Approach commentary
The delivery matches the approved 5-step plan plus the one required extension, and each fix-approach rider landed:
- **G1** (depth-1 `neighbors_sql` honoring the flag via `resolve_memoized`) is present as real code, not a flag flip — the flagship `uni-capability` depth-1 consumer actually resolves now. This was the load-bearing "last mile"; without it the flip was a no-op for the most common call.
- **subgraph edge-endpoint fix** (extension) was necessary and correct: resolving the node but recording the raw edge target left the terminal as a disconnected node that the dangling filter dropped — a latent defect invisible under prior tests. The cross-surface consistency test now guards it at default depth 1.
- **N1 memo**, **N2/N3 recorded in #5409**, **atomic ADR correction**, **byte-identity canary preserved** — all confirmed in the diff.

**Security review assessment (I did not re-run it):** the LOW-risk verdict is well-grounded. SQLi surface is positional-bind throughout; the flag is a plain `Option<bool>` (no coercion); memo keys on `raw_id` so cross-id contamination is structurally impossible; `follow_to_current` is 50-hop-capped with `None → raw` fallback (no hang); read-pool only, no write path, worst case is staleness not disclosure. No new deps, no secrets, no `unsafe`. I concur — nothing to address before merge on security grounds.

**The one product divergence worth naming (N2):** context_graph resolves *silently* — EdgeRecord carries no resolution-notice field, unlike context_get's loud resolution (#5387/#5385). A `uni-capability` gating decision made off a silently-resolved (or silently-unresolvable dead-end) graph read cannot distinguish an active terminal from an as-stored deprecated node. Accepted as a product bet and recorded in #5409 — correct for now. But it is recorded only in an ADR; there is no tracked issue to revisit it if gating consumers start needing loud resolution. Recommend a lightweight follow-up issue so the bet is revisitable, not lost.

- **Merge as-is, or address the fmt WARN first?** Merge as-is. The stray reformat (`context_get_tool_attr().description.as_deref()` split across 3 lines in `get_resolution_tests`) is cosmetic and almost certainly rustfmt-mandated — if so, reverting it would fail `cargo fmt --check`. Verify with `cargo fmt --check` before touching it; do not block merge on it.
- **Is the SL7 non-goal correction part of this PR?** No — correctly deferred. Capability-map mutation is the vision session's assignment, not the delivery PR's. But it must happen promptly after merge (see actions), or the corpus is self-contradictory: SL7 says graph stays default-off while shipped behavior is default-on.
- **Should the context_graph read-currency flip be captured as its own capability?** Worth the vision session's consideration. The flip delivers "context_graph read surfaces resolve to current by default" — a read-currency property parallel to SL7 but for a different surface. Right now it lives only as a corrected ADR with no capability node. Not a blocker; a corpus-completeness call for the vision session.

### Recommended actions
1. **Merge.** All root-cause legs present and independently gate-verified; security LOW; tests green (4 new + workspace + 28 smoke + 114 graph/context-read integration).
2. **Post-merge (load-bearing):** hand the vision session the action to `context_correct` SL7 (#5390) — mark clause 2 proven, and rewrite/remove the now-false "graph resolve_supersessions stays default-off is intended" non-goal so the map matches #5409 and shipped behavior. This is the single item that, if skipped, leaves the corpus contradicting itself.
3. **Post-merge (optional):** file a tracked follow-up for N2 (add loud resolution / a notice field to EdgeRecord) so the accepted silent-resolution bet is revisitable if `uni-capability` gating consumers need it — today it lives only in an ADR.
4. **Optional:** vision session decides whether the context_graph read-currency flip warrants its own capability node.
5. Before touching the fmt WARN, run `cargo fmt --check`; leave the line if it is fmt-mandated.

*Independence contract satisfied — spawn prompt carried only IDs and artifact references; I formed this from the artifacts and the corpus directly.*

Issue #363, opened 2026-03-23
feat: content efficacy model — utilization-based freshness, curation surface, quarantine TTL lifecycle


dug-21:
## Background

The current freshness model uses time-decay (`max(updated_at, last_accessed_at)` with a 7-day half-life) as a proxy for knowledge relevance. This predates session/injection data. It now produces misleading health signals: the entire corpus decays equally during inactive periods, old-but-valid reference knowledge is penalized, and lambda (~0.60) mostly reflects \"it's been a while\" rather than genuine knowledge health.

This issue captures the design direction agreed in a design session on 2026-03-23.

---

## Core Shift: Time as TTL Gate, Not Health Signal

Time moves from the **confidence score** into **lifecycle transition thresholds**. A lesson learned 6 months ago with a successful injection history should have the same confidence as a week-old one. An entry that was never injected across all sessions is a pruning candidate regardless of age.

The \"2 weeks off\" property becomes correct: the corpus pauses during project inactivity rather than decaying.

---

## Primary Evidence Signal: \"Never Injected\"

The key distinction:

- **Never injected (ever)** — has never found an audience across all sessions. Primary lifecycle progression candidate.
- **Not recently injected** — normal for reference knowledge (ADRs, edge-case procedures, old lessons). Leave it alone.

An entry with even one successful injection has demonstrated topical relevance. Lessons from early project phases are still valid when work returns to that area — pure utilization-based pruning would destroy them before they're needed.

---

## Lifecycle Model

```
Active
  ↓  manual  OR  auto (never-injected + no votes + age > threshold)
Deprecated
  ↓  MANUAL ONLY — admin decision, no auto-promotion
Quarantine                ← purposeful block, admin-gated (unchanged)
  ↓  TTL (e.g. 90 days)
Candidate for Deletion    ← surfaced in context_status for human confirmation
  ↓  confirmed OR second TTL expires
Removed
```

**Deprecated → Quarantine stays manual.** Auto-promoting deprecated entries breaks the implicit contract of deprecated status and can silently destroy correction chains without human acknowledgement.

---

## Correction Chain Protection

Entries with `superseded_by` pointing to an active entry are load-bearing in the knowledge history. Chain-connected deprecated entries require explicit admin action to quarantine — the human must acknowledge the history can be discarded. Standalone deprecated entries (no chain connections, never injected) are the natural long-term cleanup targets, but the manual gate still applies.

---

### 1. Freshness Dimension in Lambda (confidence/coherence)
[...]

PR #888, opened 2026-07-03
fix: repoint recoverable inbound edges before compaction to stop silent link loss (#879)

### LINKED ISSUE #879: fix: compaction tick deletes past-ceiling inbound edges instead of repointing (silent link loss on hot-node correction)

dug-21:
## Summary

Fixes the silent inbound-link-loss bug in the background compaction tick (#879). The tick DELETEd any `graph_edges` row whose target was not `Active`, conflating two states that both read as "not Active":
- Deprecated **with** a `superseded_by` successor → a valid terminal exists → **recoverable** (must repoint)
- dead / quarantined **with no** successor → nothing to repoint to → correctly deleted

Inbound edges to a corrected (Deprecated-with-successor) node that the synchronous redirect didn't reach — past the `REDIRECT_CEILING=50` cap, per-edge redirect failures, and commit/redirect races — were physically deleted at the next tick instead of repointed to the terminal.

## Fix — repoint-then-compact (one pass, Option A / unbounded)

1. **Repoint** (before the DELETE): for each edge whose target is Deprecated with `superseded_by` set, repoint to the CTE-resolved terminal (`query_current_terminal`, never in-memory `find_terminal_active` — #4468) via the idempotent `redirect_graph_edge`.
2. **Compact** (existing behavior): delete only the true residual — non-Active with no valid terminal.

Design decisions from the architect review baked in:
- **Unbounded repoint** (no `LIMIT B`) so the repoint set matches the unbounded DELETE set exactly — bounded-repoint + blanket-DELETE would delete overflow before a later tick repoints it.
- **F2: filter on `source = EDGE_SOURCE_AGENT` column**, not a relation_type blocklist — lossless and prevents laundering machine-generated edges into trusted agent attribution (`redirect_graph_edge` stamps `weight=1.0/source=agent`).
- Guards: self-loop skip, non-Active-source skip, per-edge warn-and-continue (never aborts the tick — vnc-017 #4462).
- **All three DELETE sites consolidated to one shared helper** (`run_orphaned_edge_compaction`, the `run_single_tick` stdio inline block, and the test helper now call the production fn) — the divergent test copy is why the gap escaped the suite.

Ordering respects the crt-034 SR-06 invariant: repoint runs inside the compaction block ahead of the DELETE, before typed-graph rebuild.

## Scope

Going-forward correctness only — corpus already truncated under the old ceiling is unrecoverable (compaction is EveryTick). Flagged-but-out-of-scope (separate issues): edges to `Proposed` entries, quarantine-restore edge loss, `DependencyOnDeprecatedRule` starvation.

## Tests

8 new tests, incl. the catching test (seed S→A, deprecate A→B, run the **production** compaction fn, assert S→B survives), dead-end deletion, duplicate-terminal dedup, machine-source non-repoint, self-loop / non-Active-source guards, and >ceiling end-to-end.

- `cargo test --workspace`: PASS
- `cargo clippy --workspace -- -D warnings`: clean
- Integration smoke: 28/28; `get_edges`: 18/18

## Gate

Bug Fix Validation — **PASS** (13/13)

Closes #879

### DIFF
--- a/crates/unimatrix-server/src/background.rs
+++ b/crates/unimatrix-server/src/background.rs
@@ -26,7 +26,7 @@ use unimatrix_observe::extraction::{
     quality_gate, run_extraction_rules,
 };
 use unimatrix_observe::types::ObservationRecord;
-use unimatrix_store::{EntryRecord, ShadowEvalRow, Status, counters};
+use unimatrix_store::{EntryRecord, ShadowEvalRow, Status, counters, query_current_terminal};
 
 use unimatrix_adapt::AdaptationService;
 
@@ -38,6 +38,7 @@ use crate::infra::embed_handle::EmbedServiceHandle;
 use crate::infra::nli_handle::NliServiceHandle;
 use crate::infra::rayon_pool::RayonPool;
 use crate::infra::session::SessionRegistry;
+use crate::mcp::edge_write::{EDGE_SOURCE_AGENT, redirect_graph_edge};
 use crate::server::PendingEntriesAnalysis;
 use crate::services::ServiceError;
 use crate::services::co_access_promotion_tick::run_co_access_promotion_tick;
@@ -507,55 +508,15 @@ async fn run_single_tick(
         }
     }
 
-    // crt-021 Step 2: GRAPH_EDGES orphaned-edge compaction.
+    // crt-021 Step 2: GRAPH_EDGES repoint-then-compaction (repoint added bugfix-879).
     //
-    // Deletes edges where either endpoint no longer exists in the entries table.
-    // Runs AFTER maintenance_tick (which includes VECTOR_MAP compaction) and
-    // BEFORE TypedGraphState rebuild, so the rebuild sees the post-compaction state.
-    //
-    // Uses direct write_pool — this is a bounded maintenance write, not an
-    // analytics event (per ADR-001 write-path contract: direct write_pool for
-    // maintenance, analytics queue only for bootstrap-origin writes).
-    //
-    // Non-fatal: if the DELETE fails, we log the error and proceed with rebuild
-    // against the pre-compaction state. Orphaned edges at build time are silently
-    // skipped by build_typed_relation_graph (missing endpoints → missing node_index
-    // entries → edge silently ignored). One more tick cycle will retry compaction.
-    //
-    // Intentionally unbounded in crt-021 (NF-09). A per-tick LIMIT batch is
-    // deferred post-ship; indexes on source_id/target_id make this efficient.
-    {
-        let compaction_result = sqlx::query(
-            "DELETE FROM graph_edges
-             WHERE source_id NOT IN (SELECT id FROM entries WHERE status = ?1)
-                OR target_id NOT IN (SELECT id FROM entries WHERE status = ?1)",
-        )
-        .bind(Status::Active as u8 as i64)
-        .execute(store.write_pool_server())
-        .await;
-
-        match compaction_result {
-            Ok(result) => {
-                let rows_deleted = result.rows_affected();
-                if rows_deleted > 0 {
-                    tracing::info!(
-                        rows_deleted = rows_deleted,
-                        "background tick: GRAPH_EDGES orphaned-edge compaction complete"
-                    );
-                }
-            }
-            Err(e) => {
-                // Non-fatal: log error, proceed with rebuild on pre-compaction state.
-                // Orphaned edges persist for one more tick cycle — not a correctness
-                // issue; build_typed_relation_graph silently skips edges with missing
-                // endpoints (node_index lookup returns None → edge is skipped).
-                tracing::error!(
-                    error = %e,
-                    "background tick: GRAPH_EDGES compaction failed; proceeding with rebuild on pre-compaction state"
-                );
-            }
-        }
-    }
+    // Repoints recoverable inbound edges (target Deprecated-with-successor) onto
+    // the Active terminal, THEN deletes the true residual (endpoints missing,
+    // quarantined, dead-end chains). Runs AFTER maintenance_tick (which includes
+    // VECTOR_MAP compaction) and BEFORE TypedGraphState rebuild, so the rebuild
+    // sees the post-compaction state. Shared helper — do NOT inline a divergent
+    // second copy of the DELETE here (F3).
+    run_orphaned_edge_compaction(store).await;
 
     // ── ORDERING INVARIANT (crt-034, ADR-005) ─────────────────────────────────────
     // co_access promotion MUST run:
@@ -825,11 +786,27 @@ async fn run_single_tick(
 // and still used by the legacy `background_tick_loop`.
 // ---------------------------------------------------------------------------
 
-/// GRAPH_EDGES orphaned-edge compaction (crt-021 Step 2).
+/// GRAPH_EDGES repoint-then-compaction (crt-021 Step 2; repoint added bugfix-879).
 ///
-/// Extracted verbatim from `run_single_tick` (the `513-544` block). Non-fatal:
+/// Two ordered phases in a single pass:
+///   1. Repoint recoverable inbound edges (target Deprecated-with-successor) onto
+///      the target's Active terminal, so a hot-node correction does not silently
+///      drop inbound referrers the synchronous redirect loop did not reach
+///      (ceiling truncation, per-edge failure, races). See
+///      [`repoint_deprecated_target_edges`].
+///   2. Delete the true residual: edges whose source/target is non-Active AND has
+///      no valid terminal (endpoints missing, quarantined, dead-end chains).
+///
+/// The repoint MUST run before the DELETE and inside this block — never as a new
+/// step between compaction and typed-graph rebuild (crt-034 SR-06 ordering, #3827).
+///
+/// Non-fatal throughout: repoint is per-edge warn-and-continue (vnc-017 #4462);
 /// on DELETE error, logs and returns — rebuild proceeds on pre-compaction state.
-pub(crate) async fn run_orphaned_edge_compaction(store: &Arc<Store>) {
+pub(crate) async fn run_orphaned_edge_compaction(store: &Store) {
+    // Phase 1: repoint recoverable inbound edges before the DELETE (bugfix-879).
+    repoint_deprecated_target_edges(store).await;
+
+    // Phase 2: delete the true residual.
     let compaction_result = sqlx::query(
         "DELETE FROM graph_edges
          WHERE source_id NOT IN (SELECT id FROM entries WHERE status = ?1)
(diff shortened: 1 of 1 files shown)


dug-21 (commented):
## Security Review — bugfix-879 (repoint-then-compact)

**Risk level: LOW. No blocking findings.** Fresh-context review of the full diff on `crates/unimatrix-server/src/background.rs` plus verification against `edge_write.rs`, `graph_queries.rs`, and the `Status` schema.

### OWASP / injection
- All SQL is parameterized (`?1..?n` binds). The candidate scan, the residual DELETE, and the test helpers use binds. The only `format!`-built query (`query_current_terminal`) interpolates a **compile-time const** column list (`ENTRY_COLUMNS_E`); the `id` is bound. No untrusted string interpolation. **No injection surface.**
- No external/attacker input reaches this path — it is a background maintenance tick over local DB state. No access-control or secrets concern; no hardcoded credentials in the diff.

### Attribution integrity (the security-relevant control) — SOUND
- The candidate scan filters on the `source` **column** (`ge.source = EDGE_SOURCE_AGENT`), not a relation_type blocklist. Machine classes use distinct source constants (`nli`, `cosine_supports`, `EDGE_SOURCE_CO_ACCESS`), so they never enter the repoint set.
- `redirect_graph_edge` re-materializes as `weight=1.0, source='agent'` — but since only already-`agent` edges are selected, there is **no cross-class laundering** and no weight inflation (agent edges are canonically weight 1.0; confirmed at `edge_write.rs:205`). I could not find a path where an `nli`/`cosine`/`co_access` edge is repointed. `test_compaction_does_not_repoint_machine_source_edge` asserts the machine edge is deleted-as-residual and no `source='agent'` row is minted.

### Terminal resolution — no loop / wrong-terminal / self-loop gap
- `query_current_terminal` is a recursive CTE bounded by `c.depth < 50`, so a corrupted or malicious supersession **cycle** terminates and yields `None` → edge left for DELETE (no infinite loop).
- Final SELECT requires `superseded_by IS NULL AND status = 0`, so a **quarantined** terminal (status=3) resolves to `None`. This satisfies `redirect_graph_edge`'s caller-contract quarantine check even though the repoint caller does no explicit quarantine validation — the resolver enforces it. No repoint onto a quarantined node.
- **Self-loop guard** (`terminal.id == source_id`) covers B→A where A's terminal is B; `test_compaction_repoint_skips_self_loop` confirms no B→B is created.

### Blast radius (worst case of a subtle bug)
- Bounded. The repoint set is a **subset** of what the residual DELETE would remove anyway (agent-source + Active source + Deprecated-with-successor target). Worst case of a mis-resolution is an edge repointed to the wrong Active terminal instead of being deleted — a graph-integrity blemish, not data loss beyond the pre-fix baseline, and requiring pre-existing corrupted supersession data. No external trigger, no privilege boundary, no disclosure.
- Failure posture is safe: per-edge warn-and-continue, non-fatal candidate-scan error, and the residual DELETE always runs afterward.

### Regression risk
- The consolidated `run_orphaned_edge_compaction` DELETE is byte-identical to the removed inline block; the test helper now calls the **production** function instead of a duplicated SQL string, reducing drift risk.
- Repoint now runs on the stdio single-store path (`run_single_tick`) as well — this is the intended fix, and no boundary makes that path unsafe. Signature change `&Arc<Store>` → `&Store` is compile-enforced.

### Non-blocking notes (follow-up, not merge-gating)
1. **N+1 on the repoint loop**: `query_current_terminal` (a CTE) plus a `redirect_graph_edge` transaction run per candidate edge, every tick, unbounded. Unbounded is deliberate (Option A — must match the DELETE set), and it is self-healing, but under pathological correction volume this is O(N) transactions/tick. The only perf guard test covers the DELETE (1000 rows), not the repoint path. Consider the deferred bounded-batch (NF-09) if correction volume grows.
2. **Redundant terminal re-resolution**: multiple inbound edges to the same deprecated target re-resolve the identical terminal; memoizing by `old_target_id` would cut CTE calls. Efficiency only.

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. The delivery closes a named success criterion of the self-learning goal, the security review is LOW with no blocking findings, and every fix-approach decision (Option A, F2 source=agent, F3 consolidation) landed intact. One housekeeping ask before/at merge: file the three flagged out-of-scope findings as tracked issues so they don't die in this thread.

### Vision / roadmap fit
Lands directly on goal #5219 (self-learning) success criterion: *"The typed knowledge graph stays integrity-consistent under correction — no orphaned edges, no dropped referrers."* The bug was a literal violation of that clause — a correction silently dropped inbound referrers at the next tick. This is closing a committed gap, not incidental cleanup. It also reinforces the poisoning-resistance criterion via the F2 `source=agent` filter (machine-inferred edges cannot be laundered into agent-attributed PPR mass).

Sequencing worked out: #882 (raise REDIRECT_CEILING) is now CLOSED — it reduces how many edges reach the tick; #879 is the backstop that makes the tick non-destructive at any ceiling. Independent and complementary, as the investigator predicted.

### Delivery assessment (whole)
- **Root cause, right layer.** Repoint runs *before* the residual DELETE, inside the compaction block ahead of typed-graph rebuild (crt-034 SR-06). Fixes the state-conflation, not a symptom.
- **All three fix-approach decisions honored.** Option A / unbounded (repoint set == DELETE set, no overflow gap — the one forbidden combination, bounded-repoint + blanket-DELETE, is avoided); F2 filters on the `source` column not a relation_type blocklist (lossless by construction); F3 consolidates all three DELETE sites into one production helper.
- **Regression-durability is the real win.** The test helper now calls the production `run_orphaned_edge_compaction` instead of a pinned SQL copy — the copied-SQL test is exactly why the original gap survived the suite. This closes the meta-defect, not just the defect.
- **Guards all present and tested:** self-loop skip, non-Active-source exclusion (in SQL), CTE terminal resolver (#4468, not the tick-stale in-memory one), per-edge warn-and-continue (#4462). Catching test seeds S→A, deprecates A→B, runs the production fn, asserts S→B survives.
- **Minimal diff:** single file (`background.rs`), no schema/audit change (edges are outside the hash chain).

### Security review — concur, no re-run
LOW risk, no blocking findings, and I see nothing that changes that call. The two non-blocking notes are efficiency-only:
1. **N+1 unbounded repoint** — O(N) CTE + redirect txns per tick. Deliberate under Option A, convergent/idempotent, bounded by TICK_TIMEOUT. The fix-approach correction materially de-risks this: the pre-existing corpus backlog is *already lost* (compaction is EveryTick), so steady-state repoint volume is small (per-edge failures + races + the commit→redirect window). Do not size for a burst the corpus won't produce.
2. **Redundant terminal re-resolution** — memoize by `old_target_id`. Pure efficiency.
Neither is merge-gating.

### Open items / recommended actions
1. **Merge as-is.** Correctness, guards, catching test, security posture, and suites (workspace green, clippy clean, smoke 28/28, bug-area integration green) all satisfied.
2. **File the three flagged out-of-scope findings as tracked issues before this thread is archived** — I could not find existing issues for any of them:
   - edges to `Proposed` (status 2) entries deleted every tick (same over-breadth, no supersession) — legitimately a *different* outcome, deferral is right, but track it;
   - quarantine-restore edge loss (edges to quarantined targets deleted; restore loses them) — reversible-state defect, distinct semantics;
   - `DependencyOnDeprecatedRule` one-tick-window starvation (this fix partially restores its signal for the repointed class) — latent design question.
   These are correctly out of scope for #879 (each a different outcome, not the last mile of this one), but right now they live only in the #879 comments.
3. **Retrospective (not now):** extend lesson #4156 (allowlist compaction / retention-leak) with the new corollary the reviewers surfaced — *"bounded-repoint + blanket-DELETE reintroduces the loss"* (F1). Genuinely new failure mode; worth capturing when the fix ships.
4. **Optional, low priority:** the perf-guard test covers the 1000-row DELETE path only; the repoint path has correctness tests (incl. >ceiling 60-edge) but no latency guard. Fine to leave given small steady-state volume — revisit only if #363/#625-class volume concerns materialize.

### Scope / right-sizing note
Going-forward-only is the correct boundary and is stated plainly in the PR — pre-existing lost edges are unrecoverable (backup restore only), and the PR sets that expectation. No gold-plating, no deferred last mile: the value-delivering outcome (a correction no longer drops referrers) is fully delivered, not a mechanism awaiting a follow-up.

Issue #895, opened 2026-07-04
Detect agent-Prerequisite edges to an entry being deprecated (detect-at-deprecate governance signal)


dug-21:
**Feature crt-058** — reclassified from bug to enhancement. Scope approved (design Session 1). This body mirrors the approved `product/features/crt-058/SCOPE.md`.

## Summary

`context_deprecate` is a rare, terminal operation (entry no longer needed or causing harm, no successor — `superseded_by` stays NULL). Today it deletes **no edges**; both inbound and outbound edges touching the entry are only removed later by the `EveryTick` orphaned-edge compaction (~900s later). Between the flip and the next tick, live entries keep dangling agent-declared references to the retired entry.

`context_correct` already heals the analogous case by *repointing* inbound edges onto its successor. Deprecation has no successor, so the parallel action is to **delete**. This feature makes a product decision: clean those edges up **eagerly, at the deprecation event**, for the one entry being deprecated — pulling the tick's blanket delete forward — rather than waiting for the periodic sweep.

## Behavior

At `context_deprecate`, after the step-5 idempotency early-return and the status flip, run one synchronous, non-fatal statement:

```sql
DELETE FROM graph_edges WHERE (source_id = ?entry OR target_id = ?entry) AND source = 'agent'
```

Then: **(a)** warn the calling agent inline with the count removed; **(b)** record the removal in the audit log. On any error: log at `warn`, report zero, return normal success — never propagate. The EveryTick compaction remains the backstop.

## Value / framing

Integrity / data-accuracy: resolve dangling references by deleting them at the source event, keeping the graph's dependency data accurate the instant an entry is retired. A *resolution*, not a flag. Integrity is a documented rationale, not a product goal — feature kept minimal. Explicitly **not self-learning / not drift-adaptation**.

## In scope

- Both directions: inbound (`target_id = entry`) + outbound (`source_id = entry`).
- Agent/human-authored edges only: `source = 'agent'` (the only agent-authored provenance value; no distinct `human` value exists in `graph_edges.source`).
- Inline count to caller (a) + audit record (b).
- Non-fatal; single indexed statement on `write_pool_server()`; no new table/migration/helper indirection.

## Non-goals

- Not self-learning / drift-adaptation.
- Not detection / governance nudge — the condition is resolved (deleted), not flagged.
- No relation-type filtering (agent edges removed regardless of relation); provenance-filtered only.
[...]

PR #897, opened 2026-07-04
fix: retire tick-starved DependencyOnDeprecatedRule and stale_dependency_edges metric (#891)

### LINKED ISSUE #891: DependencyOnDeprecatedRule input destroyed by compaction within one tick
## Summary

Split off from #879 (out of scope there). `DependencyOnDeprecatedRule` reasons over edges whose target is a Deprecated entry. But the orphaned-edge compaction tick (`EveryTick`) deletes those same edges within ~one tick, so the rule's input is destroyed before it can reliably observe it — the rule is effectively starved.

Note: #879 now **repoints** agent-sourced edges to Deprecated-with-successor targets before the DELETE, which further removes exactly the edges this rule would have inspected. This interaction should be reconciled: either the rule runs against a snapshot before repoint/compaction, or its trigger semantics are redefined.

## Evidence
- Compaction is registered `EveryTick` and deletes non-Active-target edges. `crates/unimatrix-server/src/background.rs`.
- `DependencyOnDeprecatedRule` depends on the presence of edges-to-Deprecated that compaction removes.

## Proposed direction
Decide the intended ordering/contract between `DependencyOnDeprecatedRule` and the compaction tick — run the rule ahead of compaction on the same tick, or against pre-compaction state, or redefine the rule to key off `superseded_by` metadata rather than the live edge.

Context thread: #879

dug-21:
Closes #891.

## Root cause
`DependencyOnDeprecatedRule` and the `stale_dependency_edges` cohesion metric were structurally starved: their input (agent `Prerequisite` edges to a Deprecated target) is deleted by the `EveryTick` orphaned-edge compaction within ~one tick, while they only read on-demand at `context_cycle_review`. Both also carried a direction defect — they keyed on Deprecated *source*, but the finding claim asserts the *target* is Deprecated ("update to the successor"), so any finding would have named the wrong entry.

## Fix (approved scope)
Right-sized to a retirement + direction-correctness fix rather than building capture-before-destroy machinery for a rare, post-#879 residual signal:

1. **Retire `DependencyOnDeprecatedRule`** from the detection registry and handler injection.
2. **Retire the `stale_dependency_edges` cohesion metric** and its status/report field.
3. **Delete both direction-defective queries** (`query_stale_prerequisite_edges_for_cycle`, cohesion Query 3) as dead code — their only callers were the retired rule/metric — so nothing ships asserting the wrong direction.
4. Remove the 4 obsolete `infra-001` integration tests that asserted the retired rule/metric; redirect-behavior coverage is retained by `test_correct_auto_redirects_prerequisite_edges`.

**Not touched:** compaction DELETE predicate, #879 repoint, tick ordering, schema/migrations. System/machine edges (CoAccess, NLI, cosine, S1/S2/S8) confirmed a non-issue — disposable derived data, blanket-deleted by compaction and never regenerated to a non-Active endpoint.

## Follow-up
The real governance signal — detect agent `Prerequisite` edges to an entry at `context_deprecate` time (before compaction destroys the evidence) — is deferred to #895, with the emit-surface (inline vs audit vs batched) as the open product decision.

## Validation
- `cargo test --workspace` PASS, `cargo clippy --workspace --all-targets -D warnings` PASS
- Integration smoke: 28 passed; blast-radius slice green; pytest collection clean
- Gate 3 Bug Fix Validation: PASS (7/7)
- No `unsafe`/`todo!()`/`unimplemented!()`/TODO/FIXME/xfail introduced


dug-21 (commented):
## Security Review — bugfix-891 (retire tick-starved dependency rule/metric)

**Risk: LOW. No blocking findings.** Fresh-context review of \`git diff e69139dd..HEAD\` (commits 9a804709 + 64b0cb8b).

### OWASP / trust boundary
- **Deletion-only, no new attack surface.** The removed SQL (\`compute_graph_cohesion_metrics\` Query 3, \`query_stale_prerequisite_edges_for_cycle\`) used hardcoded relation-type literals + bound params — a safe query was removed, none added. No input-validation sink removed at any boundary; the deleted cycle query *consumed* a bound \`feature_cycle\` param, so removal drops an input sink rather than exposing one.
- **No access control weakened.** The retired \`DependencyOnDeprecatedRule\` only ever emitted an informational \`Warning\` finding — it never gated anything. \`validate_target\` in \`edge_write.rs\` still allows \`Deprecated\` targets exactly as before (only the doc comment changed). Removing the rule stops *enforcing* nothing.
- **No secrets** in the diff (scanned).

### Dangling-caller / regression check (verified by grep)
- All \`default_rules()\` call sites updated to the single-arg form; no live reference to \`DependencyOnDeprecatedRule\`, \`query_stale_prerequisite_edges_for_cycle\`, or the \`GraphCohesionMetrics.stale_dependency_edges\` field remains outside retirement-breadcrumb comments and the new negative test. Clean.
- No JS/TS edge-client consumer reads \`stale_dependency_edges\`.
- Compaction DELETE, #879 repoint, and tick ordering (\`background.rs\`) are untouched — confirmed absent from the diff.
- Regression risk low: the rule was structurally starved (never fired on real data) and the metric was near-permanently 0, so live \`context_cycle_review\`/\`context_status\` output loses no real information — corroborated by the byte-identical baseline tests needing only the 23→22 rule-count assertion changed.

### Non-blocking findings
**S1 (low) — README doc drift.** \`README.md:241\` still documents \`stale_dependency_edges\` as one of \"six graph cohesion metrics\" reported by \`context_status\`. The field no longer exists in the output; update the sentence and the count. (Historical planning docs under \`product/\` are archival — leave them.)

**S2 (low) — Response-shape change not yet release-noted.** \`stale_dependency_edges\` was an always-present (non-\`skip_serializing_if\`) field in the \`context_status\` JSON; removing it is a breaking change to a stable MCP response schema. No in-repo consumer requires it and the failure mode for any strict external parser is safe-and-loud (missing-key error, not silent corruption), so this is acceptable — but the CHANGELOG has no #891 entry. The release step should note the field removal (and that the retired metric's meaning is gone), as the investigator and design review both flagged.

### Blast radius (worst case)
A strict external client that hard-required \`stale_dependency_edges\` in \`context_status\` JSON would hit a missing-key parse error — safe, loud, no data corruption or privilege impact. No such consumer exists in-repo. Worst case is low and self-announcing.

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. Deletion-only retirement, correctly right-sized against my fix-approach recommendation, security LOW with no blockers. One residual doc-drift the security review's S1 missed (README:211) — cheap to fold in before merge; CHANGELOG note is a release-time action.

### Vision / roadmap fit
Squarely on goal #5219 (Self-learning intelligence), whose integrity-consistency criterion the `dependency_on_deprecated` rule and `stale_dependency_edges` metric were meant to serve on the `context_cycle_review` / `context_status` drift surface. As shipped, both were structurally dead — the rule's Prerequisite-to-Deprecated input is deleted by the EveryTick compaction before the on-demand read runs — and the rule emitted the wrong fact (source-Deprecated) when it ever fired. A dead signal that lies when it fires is a hole in the trustworthiness criterion, not an asset. Retiring it makes the drift surface *honestly* smaller rather than falsely populated. The real governance signal (detect-at-deprecate, before compaction destroys the evidence) is deferred to #895. Net-positive for the goal.

### Approach commentary
- **The retirement scope is the call I recommended at fix-approach (OQ5), taken one coherent step further.** I flagged "fix-the-lie-and-defer" as a legitimate cheaper interim vs. standing up the full capture-table machinery for a narrow post-#879 residual. The human chose full retirement — cleaner than a fixed-direction-but-still-starved rule, since without the capture table the rule can *never* fire correctly regardless of direction. Retiring the whole thing removes the false signal outright.
- **Deleting both direction-defective queries as dead code sidesteps the F2 direction decision entirely.** Once the rule/metric are retired, `query_stale_prerequisite_edges_for_cycle` and cohesion Query 3 have no callers, so nothing ships asserting either the wrong or a newly-corrected direction. That is the cleanest resolution of the source-vs-target defect — no direction call needed, no wrong-direction finding can escape.
- **This is not a stranded partial.** The deferred value (a working dependency-on-deprecated signal) is a *different, larger* outcome requiring the capture machinery — a legit defer to #895, not a last-mile cut. Retiring a non-functional mechanism now leaves no mechanism-without-outcome behind. Right-sized.
- **Security review concurrence.** Its LOW / no-blocker verdict is sound: deletion-only, a safe bound-param query removed (no sink added), no access control weakened (the rule only ever emitted an informational Warning; `validate_target` still allows Deprecated targets). I concur.
- **Verification is solid.** cargo test/clippy green, integration smoke 28 passed, gate 7/7 PASS. The one rework round (4 obsolete Python integration tests in infra-001 the Rust-only fix missed) was caught by the tester and cleaned; redirect-behavior coverage is retained by `test_correct_auto_redirects_prerequisite_edges` (asserts `graph_edges` directly, not the retired metric). No coverage gap.

### Assessment of security-review findings (recommend actions)
- **S1 (README doc drift) — partially fixed, one location still stale.** Commit `cd350e44` corrected README:241 (the cohesion-metrics paragraph: "six→five", dropped the `stale_dependency_edges` sentence). But **README:211 was not touched and still documents the retired rule**: it says "23 detection rules across 4 categories" (now 22) and describes "a `DependencyOnDeprecated` rule (severity: Warning) that fires when any `Prerequisite` edge … points to a deprecated source." That rule no longer exists and the count is wrong — the README would ship contradicting the PR. Docs-only, not a correctness blocker, but it is the natural sibling of the S1 fix. **Recommend: fold a one-line README:211 fix into the same docs commit before merge** (cheap), or a fast follow-up if you prefer to merge now.
- **S2 (response-shape change not release-noted) — accept, handle at release.** Removing `stale_dependency_edges` from the `context_status` JSON is a breaking change to a stable MCP response field. No in-repo consumer requires it; the failure mode for any strict external parser is safe-and-loud (missing-key error, not silent corruption). Acceptable to merge. The CHANGELOG entry is generated at release (`/uni-release`), so this is a release-step action, not a merge gate — but it must land in the release notes (as the investigator, design review, and I flagged at fix-approach: the retired metric's meaning is simply gone).

No open product questions remain in the PR. The three fix-approach decisions were all resolved and I concur:
1. **Direction (F2)** — mooted by deleting both queries as dead code. Nothing ships asserting a direction. Correct.
2. **Trust filter (F3)** — mooted; no capture path was built.
3. **Full-fix-vs-defer (OQ5)** — resolved as retire-now, build-right-later (#895). This is the right-sized call.

### Recommended actions
- **Before merge (cheap):** fix README:211 — "23"→"22" detection rules and remove the `DependencyOnDeprecated` rule description. Closes the S1 gap the security review missed.
- **At release:** add the #891 CHANGELOG entry noting `stale_dependency_edges` removed from `context_status` output and the rule retired (S2).
- **Confirm #895 carries the real signal** (detect agent Prerequisite→Deprecated edges at `context_deprecate` time, before compaction) so this retirement is a clean hand-off, not a silent capability drop.
- Otherwise: **merge.**

PR #906, opened 2026-07-05
fix: type the context_cycle_review transcript MCP axis so it is client-callable (#901)

### LINKED ISSUE #901: context_cycle_review: transcript scoped-retrieval param is untyped in the MCP schema — not callable by schema-driven clients
## Problem
The crt-057 `transcript` scoped-retrieval axis on `context_cycle_review` is exposed as a **free-form, untyped object** in the MCP tool schema (no `type`/`properties`). As a result, a schema-driven MCP client has no shape to construct against, sends the value as a string, and the server's strict deserializer rejects it — so the opt-in scoped retrieval is **effectively not callable by a real client as shipped**.

## Evidence (live, on the installed crt-057 binary)
```
context_cycle_review(feature_cycle="crt-057", transcript={"match": "context_cycle_review"}, format="json")
→ MCP error -32602: failed to deserialize parameters:
  invalid type: string "{\"match\": \"context_cycle_review\"}", expected struct TranscriptScope
```
The non-destructive default (`context_cycle_review` with no `transcript`) works correctly — no `transcript_candidates` block — and `format:"summary"` correctly returns `ERROR_INVALID_PARAMS`. Only the `transcript` axis is affected.

## Root cause
`TranscriptScope` / `Window` (unimatrix-observe) were made **serde-only in Wave 1 with no `JsonSchema` derive**, so `RetrospectiveParams.transcript` is exposed via `#[schemars(with = ...)]` as a free-form object. The tool schema therefore never advertises `{ phase?, anchor?, match?, window? }`. This was flagged during crt-057 delivery (Wave-1 return: "JsonSchema NOT derived … belongs at the params layer in a later wave") but never added. Every gate/test exercised the Rust types directly (unit/contract-proven) — no live schema-driven MCP call — so it passed all three gates. (See lesson #5444: an interface internally consistent but not consumer-callable.)

## Impact
- The core #871 fix (non-destructive default) is live and correct — unaffected.
- The opt-in scoped transcript retrieval (phase/anchor/match/window) is unusable via the typed MCP schema. Agents cannot invoke it.

## Fix
- Give `TranscriptScope` / `Window` proper `JsonSchema` derives (or a params-layer schema mirror) so the tool schema advertises the object shape and clients construct it.
- Add a **live/e2e MCP-call test** (not just serde-level) that invokes `context_cycle_review` with a real `transcript` object through the MCP boundary, so this class is caught in future.

## Refs
crt-057 · PR #900 · relates lesson #5444.

dug-21:
## Summary
Fixes #901. The crt-057 `transcript` scoped-retrieval axis on `context_cycle_review` was exposed as a free-form untyped object in the MCP tool schema, so schema-driven clients sent it as a string and the strict deserializer rejected it with `-32602` — the opt-in scoped retrieval was uncallable by a real client.

## Root cause
`TranscriptScope` and `Window` (`unimatrix-observe`) derived only `Deserialize`, no `JsonSchema`; `RetrospectiveParams.transcript` papered over it with `#[schemars(with = "Option<serde_json::Value>")]`. `tools/list` therefore never advertised `{ phase?, anchor?, match?, window? }`. Every crt-057 gate exercised the Rust types directly — no live schema-driven MCP call — so it passed (lesson #5444: internally consistent but not consumer-callable).

## Fix
- Promote `schemars = "1"` to `[workspace.dependencies]`; `unimatrix-observe` and `unimatrix-server` both reference `schemars.workspace = true` (no generator-version drift).
- Derive `JsonSchema` on `TranscriptScope` and `Window` (serde `rename = "match"` preserved and verified as the literal wire key).
- Remove the `#[schemars(with = ...)]` escape hatch and the stale Wave-1 doc note on `RetrospectiveParams.transcript`; the schema now infers from the derived impl.
- Serde deserialization path untouched — zero runtime behavior change.

`transcript` now advertises `anyOf: [ { $ref: #/$defs/TranscriptScope }, { type: null } ]`, with `TranscriptScope` → `phase`/`anchor`/`match`/`window` and `window` → `$ref Window`. Self-contained `$defs`; no field required.

## Tests
- `test_schema_transcript_advertises_object_shape` — schema-snapshot; asserts the emitted `anyOf`/`$ref`/`$defs`, four sub-fields, literal `match` key, nested `window`→`Window`, non-required. Teeth-verified (fails with the escape hatch restored).
- `test_cycle_review_transcript_object_passes_mcp_boundary` — first live duplex `tools/call` client test in the workspace; drives `context_cycle_review` with a real `transcript` including a nested `window` and asserts extraction does not return `-32602`.

Workspace `cargo test` rc=0, clippy clean, 28 smoke + 13 protocol + 40 tools-subset green. Gate: Bug Fix Validation — PASS (11/11).

## Follow-ups (out of scope, deferred by direction)
- Reusable MCP-boundary schema-contract harness across all tools — this PR seeds the first `tools/call` test only.
- Extending pattern #3813 ("no `serde_json::Value` escape hatch on MCP params") — a convention decision for the crt-057/901 retro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3839,6 +3839,7 @@ version = "0.9.0"
 dependencies = [
  "bincode 2.0.1",
  "regex",
+ "schemars",
  "serde",
  "serde_json",
  "sqlx",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ all = "warn"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 bincode = { version = "2", features = ["serde"] }
+schemars = "1"
 
 # Patch anndists 0.1.4: fix cfg_if + #[cfg] block compilation with edition 2024.
 # Upstream declares edition = "2024" but has code incompatible with it.
--- a/crates/unimatrix-observe/Cargo.toml
+++ b/crates/unimatrix-observe/Cargo.toml
@@ -16,6 +16,7 @@ tokio = { version = "1", features = ["rt"] }
 regex = "1"
 unimatrix-learn = { path = "../unimatrix-learn" }
 sqlx = { version = "0.8", features = ["sqlite", "runtime-tokio"] }
+schemars = { workspace = true }
 
 [dev-dependencies]
 tempfile = "3"
--- a/crates/unimatrix-observe/src/types.rs
+++ b/crates/unimatrix-observe/src/types.rs
@@ -3,6 +3,7 @@
 //! Core observation types (HookType, ObservationRecord, ParsedSession, ObservationStats)
 //! are defined in unimatrix-core and re-exported here for backward compatibility (col-013 ADR-002).
 
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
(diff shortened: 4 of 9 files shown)


dug-21 (commented):
## Security Review — bugfix-901 (fresh context)

Read cold: full `main...HEAD` diff, issue #901, investigator + fix reports, and the modified source in full (`unimatrix-observe/src/types.rs`, `mcp/tools.rs`, `mcp/distill_scope.rs`, `server.rs`).

### OWASP / input-validation surface
- **No widened deserialization surface.** The change adds `JsonSchema` derives and removes the `#[schemars(with = "Option<serde_json::Value>")]` escape hatch. The serde path is untouched — a `TranscriptScope` object was always deserializable at the boundary; the bug was only that schema-driven clients couldn't *discover* the shape to construct it. Making the shape discoverable documents an input path that already existed; it does not create a new one.
- **Only untrusted axis (`match` regex) is already guarded and untouched.** `validate_scope_regex` → `compile_bounded_regex` (distill_scope.rs) size-bounds the pattern and rejects malformed input with `ERROR_INVALID_PARAMS`. The Rust `regex` crate is linear-time (no catastrophic-backtracking ReDoS). This is pre-existing crt-057 code — not modified by this PR.
- No injection (no shell/SQL/format-string sink), no path traversal, no access-control change, no new deserialization of untrusted bytes.

### Dependency safety
- `schemars` is **not a new crate in the tree** — 1.2.1 was already locked and used by `unimatrix-server`. The Cargo.lock delta is a single line: `schemars` added to `unimatrix-observe`'s dep list. No new transitive dependencies.
- `unimatrix-server`'s `schemars = "1"` → `{ workspace = true }` resolves to the same locked 1.2.1 (no version bump; lock confirms server's entry is unchanged).
- `cargo audit` should run as the standard pre-merge gate, but this PR introduces no new audit surface.

### Blast radius (worst case)
Fail-safe. If the derive emitted a wrong schema, the failure mode is a client being unable to call the tool (`-32602` rejection at the boundary) — a usability regression, never data corruption, disclosure, or privilege escalation. A schemars derive cannot alter runtime serde deserialization behavior. `test_schema_transcript_advertises_object_shape` (teeth-verified against escape-hatch restoration) guards the emitted shape.

### Regression risk to other tools
None identified. `JsonSchema` is derived on exactly two types (`TranscriptScope`, `Window`), and the escape hatch is removed on exactly one field (`RetrospectiveParams.transcript`). No other tool references these types, and derive is opt-in per-type — no other advertised schema changes.

### Secrets
None in the diff.

**Verdict: approve from a security standpoint. No changes required.**

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. The fix lands the outcome (a schema-driven client can now construct and call the `transcript` axis), the approach is convention-aligned, and both security reviews are sound. One non-blocking item for your awareness: the full `test_tools.py` suite was substituted, not run to completion.

### Vision / roadmap fit
This closes a *delivered-but-unreachable* gap in **SL6 (#5225, `partial`)** — "the system learns from observed agent activity" — whose `done_when` requires a distilled activity dimension that "informs retrieval/**review**." The scoped `transcript{}` retrieval on `context_cycle_review` (crt-057, ADR-002 #5434) is exactly that review-side path. crt-057 shipped it serde-consistent but not consumer-callable, so at the MCP boundary the capability was mechanism-without-outcome. This fix is what actually delivers it. SL6 stays `partial` by design — transcript is one channel of many; no capability-map change is implied. No conflict with in-flight work (ass-078 fold-at-ingest and the other untapped activity dimensions are downstream and untouched).

### Delivery assessment (whole)
- **Approach** — derive `JsonSchema` on the source types (`TranscriptScope`/`Window`) rather than a params-layer mirror. Realigns to pattern #3813 / ADR #4734 (a *real* type in `schemars(with)`, never the empty-`{}` fallback) instead of recreating the schema-vs-deserializer drift class the bug embodies. Serde path untouched → zero runtime change; blast radius is one property of one tool. Correct call, and it matches the fix-approach direction.
- **Tests** — both required guards present and green: the teeth-verified schema snapshot (asserts the emitted `anyOf`/`$ref`/`$defs`, four sub-fields, literal `match` wire key, nested `window`→`Window`, non-required) and the first live duplex `tools/call` client test in the workspace (nested `window`, asserts no `-32602`). The fix-approach condition — ship the live boundary test, do not defer it — was met.
- **Honest caveat, correctly disclosed** — the boundary test alone does not reproduce the client-side string coercion (a valid object deserializes regardless of the advertised schema); the **teeth-verified snapshot test is the actual regression catch**. The two together cover the outcome. This is stated plainly in the fix report — no hidden gap.
- **Scope discipline** — the two follow-ups (reusable MCP-boundary schema-contract harness across all tools; extending #3813 to ban the `serde_json::Value` escape hatch) were correctly held out. Do not scope either into this PR.

### Security review — assessed, concur
Two reviews (issue summary + fresh-context PR review), both LOW / READY / no blocking findings. I concur:
- No widened deserialization surface — schema discoverability documents an input path that already existed; serde extraction is byte-identical.
- The only untrusted axis (`match` regex) is pre-existing crt-057 code — size-bounded via `compile_bounded_regex`, linear-time `regex` crate (no ReDoS). Not touched by this PR.
- No new dependency (schemars 1.2.1 already locked; Cargo.lock delta is one line adding it to `unimatrix-observe`).
- Blast radius is fail-safe: a wrong-schema worst case is a client unable to call the tool, never corruption/disclosure/escalation.
Nothing to add; no security-driven follow-up.

### For your awareness (non-blocking)
The complete 198-test `test_tools.py` was killed at ~test 15 by an environment time ceiling (~25 min function-scoped server cost), not a test failure. The gate substituted the 40-test bug-relevant subset (every `cycle_review`/`transcript` case) plus 13-test protocol and 28 smoke, and judged the substitution acceptable. For a single-property schema-shape defect fully covered by the snapshot + protocol + boundary tests, that judgment is reasonable. Flagging only so the decision is yours: the post-merge CI run is the backstop for the un-run remainder.

### Recommended actions
- **Merge.** Root cause fixed, guards present and teeth-verified, security clean, blast radius bounded.
- Confirm the **MCP-boundary schema-contract harness** follow-up issue is actually filed — the "no `call_tool` client test anywhere in the workspace" gap is real and workspace-wide (every tool's advertised schema is currently unguarded against this class). This PR seeds the first such test; the generalization should not be lost.
- Leave the **#3813 extension** for the crt-057/901 retro, as planned.

Issue #916, opened 2026-07-05
test(personal-cloud): operator client-works verification (C15 floor gap — currently docs-only)


dug-21:
Closes a **claim-floor gap** for Personal Cloud (capability C15 #5157). "Operator can verify a remote client works" is **documentation-only** — the runbook's verify step is `curl /health` (proves the *server* is up, not that a *client* works); no behavioral test backs it.

**Acceptance:** a test (or scripted runbook check) that confirms a remote client actually performs a context_* operation over the pinned-TLS bundle path — not just a health probe. Found: capability-graph verification pass 2026-07-05.

dug-21:
# Bug Investigation Report: 915-916-investigator (unified diagnosis for #915 + #916)

## Bug Summary
Two personal-cloud claim-floor gaps from the 2026-07-05 capability-graph verification pass, sharing one subsystem (remote client → HTTPS container → real `context_*` op over the pinned-TLS bundle path):
- **#915 (C1)**: no green smoke for the container-cloud HTTPS leg (`docker compose up` → HTTPS → register → attach → `context_*` round-trip).
- **#916 (C15)**: operator client-works verification is docs-only — the runbook's verify step is an unauthenticated `curl /health`.

## Root Cause Analysis

**The container-cloud HTTPS path is NOT broken. It works end-to-end today — behaviorally, on real release runs. The floors fail because the proof is mis-located (the only client-carried `context_*` proof lives in a non-blocking lane that is red *by design*) and incomplete at two edges (the literal shipped `docker-compose.yml` is never exercised; the runbook's verify step has no behavioral backing).** This is a test/attestation gap, not a defect.

### What is already behaviorally green (traced, not asserted)
1. **Blocking + green** (`smoke-amd64`/`smoke-arm64`, v0.10.0 run 28760760039 and v0.11.0 run 28864724285, both `success`): `docker-http-posture-smoke.sh` Gates 1–7 — docker run of the pushed GHCR image → HTTPS listener active → register slug → pinned-cert bearer POST `/v1/<slug>/observe` → write lands in the per-slug store (hash store unchanged) → `client-bundle` emit → host `init --bundle` attach (hermetic HOME, credstore 0600) → hook fire → store growth.
[...]

PR #932, opened 2026-07-07
fix(infra-001): green blocking smoke for container-cloud HTTPS + operator client-works (#915, #916)

### LINKED ISSUE #915: test(personal-cloud): container-cloud HTTPS deploy smoke — docker compose up → HTTPS green (C1 floor gap)
Closes a **claim-floor gap** for Personal Cloud (capability C1 #5142). "Deploys two ways (local + your own cloud)" is proven for **local/STDIO** (infra-001 smoke 23/23) but the **container-cloud HTTPS** leg has **no green smoke** — the cited client_bundle_e2e is CLI cert-emission only (no HTTPS listener, no attach, no context_* call). Deferred to the release gate (same gate as N5). Until this is green, the "personal cloud" claim is honestly only 'deploy locally + slugs + secure client setup', not container-cloud.

**Acceptance:** a green, hermetic `docker compose up` (no operator config) → HTTPS listener up → register slug → attach → context_* round-trip, wired into CI / the release gate; Docker-absent is a HARD skip, never silent-green. Found: capability-graph verification pass 2026-07-05.

dug-21:
Closes #915 (C1 personal-cloud claim floor) and #916 (C15 personal-cloud claim floor).

## Root cause
Both floors were **attestation gaps, not defects** — the container-cloud HTTPS path works, but:
- the only green client-carried `context_*` proof lived in a permanently-red, non-blocking lane (`test_https_uds_parity_matrix`, deadlocked by the D5 measurability limit → #837 C0-flip);
- nothing anywhere ran `docker compose up` (the operator's first command — the #5248 drift class);
- the runbook verify step was `curl /health` only (proves the server is up, not that a client works).

## Fix — one smoke, two appended gates
Rides the already-blocking `smoke-amd64/arm64` release lanes (nan-020 ADR-001 append-only exit-1 contract); no CI wiring changes.
- **Gate 9 `client_works_gate` (#916/C15):** drives a real `context_status` through the **shipped** `mcp-bridge.js` over the pinned-TLS bundle, using the Gate-6 credstore. First green *blocking* client-carried `context_*` proof.
- **Gate 10 `compose_boot_gate` (#915/C1):** `docker compose up` on the literal shipped defaults → pinned TLS `/health` → `down -v`.

### Four blocking constraints (all implemented + validated in the diff)
1. Gate-9 bridge drive wrapped in `timeout` (rc=124 → `fail()`).
2. Gate-10 compose-plugin absent → `fail()` (exit 1), never self-skip.
3. Gate-10 curl pinned `--resolve cloud.example:8443:127.0.0.1 --cacert` (tests the shipped SAN).
4. Gate-10 binds `UNIMATRIX_IMAGE=$IMAGE --pull never` and asserts booted-image digest == `$IMAGE` — **tests the version under test, never ghcr `:latest`**.

## Files
- new `product/test/infra-001/scripts/claim-floor-lib.sh` (Gates 9+10 + teardown)
- new `product/test/infra-001/scripts/bridge-single-call-driver.js` (one `context_status` via shipped bridge)
- new `product/test/infra-001/scripts/release-gate-claim-floor-logic-test.sh` (22 off-Docker stub-seam cases)
- mod `docker-http-posture-smoke.sh` (source lib, extend cleanup, call Gates 9+10)
- mod `docker-compose.yml` (`image:` → back-compat `${UNIMATRIX_IMAGE:-…:latest}`)
- mod `docs/client-setup.md` (client-works verify step)

Out of scope / untouched: `mcp-bridge.js`, `packages/unimatrix/**`, `harness/parity_outcome.py`, `harness/parity_matrix_support.py` (#837).

## Verification
- Gate: Bug Fix Validation **PASS** (7/7, from artifacts).
- Tester (foreground): new logic test **22/22**; sibling logic tests green; **Gates 1-10 end-to-end PASS on a real Docker build from HEAD** (Gate 9 real bridge path + Gate 10 compose boot with digest match); `cargo test`/`clippy` clean; `pytest -m smoke` **32 passed**; zero new shellcheck codes.

## Follow-up (deferred, non-blocking)
The C1 (#5527) / C15 (#5570) `delivery:` tag flips are evidence-gated on the next **green release tag** and will be done post-tag/at retro — noted on both issues so the claim isn't stranded un-flipped.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,11 @@
 
 services:
   unimatrix:
-    image: ghcr.io/dug-21/unimatrix:latest
+    # Operator-facing default stays ghcr :latest; the ${UNIMATRIX_IMAGE:-...} interpolation
+    # is back-compat (unset => :latest). The release smoke's compose-boot gate (#915)
+    # OVERRIDES it with UNIMATRIX_IMAGE=<version-under-test> + `--pull never` so the gate
+    # proves the candidate build, never the already-released :latest.
+    image: ${UNIMATRIX_IMAGE:-ghcr.io/dug-21/unimatrix:latest}
     # To build locally instead of pulling:
     #   build: .
     restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,12 @@ services:
     ports:
       # TLS port ONLY — pinned-TLS HTTPS (ADR-007, FR-A8, AC-W1-S2, R-12).
       # There is NO plaintext port: OSS posture has no plaintext-to-client mode.
-      - "8443:8443"
+      # The ${UNIMATRIX_HOST_BIND:-} prefix is back-compat and GATE-SCOPED: unset (the
+      # operator-facing default) => "8443:8443" (0.0.0.0, unchanged). The release smoke's
+      # compose-boot gate (#915) sets UNIMATRIX_HOST_BIND=127.0.0.1: so the GATE publishes
+      # on loopback only (no external exposure / parallel-run host-port collision) — it does
+      # NOT change this operator default; an operator may legitimately want 0.0.0.0.
+      - "${UNIMATRIX_HOST_BIND:-}8443:8443"
     volumes:
       - unimatrix-data:/data
       # /shared is READ-WRITE by default (ADR-003 nan-015 #4652). The image bakes
--- a/docs/client-setup.md
+++ b/docs/client-setup.md
@@ -90,6 +90,8 @@ When TLS is terminated by a reverse proxy (nginx, Caddy, cloud load balancer), s
 
 ## Verifying the connection
 
+### 1. Liveness (server is up)
+
 The `/health` endpoint requires no authentication:
 
 ```bash
@@ -102,7 +104,27 @@ Returns:
 {"version": "x.y.z", "schema_version": 27}
 ```
 
-Use this for Docker HEALTHCHECK or external monitoring.
+Use this for Docker HEALTHCHECK or external monitoring. Note: `/health` proves only
+that the server is *up* — not that *your client works*. For that, do step 2.
+
+### 2. Client works (a real authenticated `context_*` op over the pinned-TLS bundle)
+
+After `unimatrix init --bundle …` writes `.mcp.json`, confirm the attached client can
+perform a real operation over the pinned-TLS bundle path — not just reach `/health`. This
+reads the bridge command your `init --bundle` wrote into `.mcp.json` and drives one
+stateless `context_status` call through it (requires `jq`):
+
+```bash
+printf '%s\n%s\n' \
+  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"verify","version":"1.0.0"}}}' \
+  '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"context_status","arguments":{}}}' \
+  | node "$(jq -r '.mcpServers.unimatrix.args[0]' .mcp.json)" \
+         "$(jq -r '.mcpServers.unimatrix.args[1]' .mcp.json)"
+```
+
+A non-error JSON-RPC result (no `"error"` field, no `"isError": true`) confirms the client
+performs real `context_*` operations over the pinned-TLS bundle. This is the same check the
+release smoke's client-works gate runs (`docker-http-posture-smoke.sh`, Gate 9).
 
 ---
 
--- a/product/test/infra-001/scripts/bridge-single-call-driver.js
+++ b/product/test/infra-001/scripts/bridge-single-call-driver.js
@@ -1,0 +1,175 @@
+"use strict";
+
+// bridge-single-call-driver.js (#915/#916) — drive ONE stateless context_* tools/call
+// THROUGH the SHIPPED `node mcp-bridge.js <projectHash>` over stdio JSON-RPC. This is
+// the client-works proof (C15/#916): it proves the ATTACHED client can perform a REAL
+// context_* op over the pinned-TLS bundle path — not merely that /health is up.
+//
+// It is a thin test helper (test tree only): it adds NO transport/cert/credstore code —
+// it spawns the UNMODIFIED shipped bridge and speaks newline-delimited JSON-RPC to it
+// (the framing the bridge's StdioFramer reads/writes). The bridge owns the pinned-HTTPS
+// session, credstore read (keyed by the projectHash arg + HOME), Mcp-Session-Id
+// capture/replay, and SSE parse; this driver re-implements NONE of that (reuse-as-is,
+// pattern #5129 — never a native http MCP entry).
+//
+// Distinct from bridge-cycle-driver.js: that drives the full context_cycle parity
+// workload (start -> tools -> stop -> review) and needs a manifest. This driver makes
+// exactly ONE tools/call (default context_status — stateless, embed-free) so the fast
+// posture smoke never perturbs the shared-lane gate-8 credstore/cycle nor eats the
+// embed-retry window.
+//
+// Usage:
+//   node bridge-single-call-driver.js <projectHash> <toolName> --bridge <path-to-mcp-bridge.js>
+// Stdout: ONE json line: {ok, tool, error}. Diagnostics -> stderr.
+// Exit: 0 on a non-error tool result; 1 otherwise. The SHELL gate wraps this whole
+// process in `timeout` (bounded) so a bridge hang cannot eat the blocking lane's job
+// timeout; the internal rpc timeouts below are defense-in-depth.
+
+const { spawn } = require("child_process");
+const fs = require("fs");
+const path = require("path");
+
+function emit(o) {
+  process.stdout.write(JSON.stringify(o) + "\n");
+}
+function die(msg) {
+  emit({ ok: false, tool: null, error: String(msg) });
+  process.exit(1);
+}
+
+function parseArgs(argv) {
+  const pos = [];
+  const opt = {};
+  for (let i = 2; i < argv.length; i++) {
+    const a = argv[i];
+    if (a === "--bridge") {
+      opt.bridge = argv[++i];
+    } else {
+      pos.push(a);
+    }
+  }
+  return { projectHash: pos[0], tool: pos[1], opt };
+}
+
+async function main() {
+  const { projectHash, tool, opt } = parseArgs(process.argv);
+  if (!projectHash || !tool) {
+    die("usage: bridge-single-call-driver.js <projectHash> <toolName> --bridge <p>");
+  }
+  const bridgePath = opt.bridge;
+  if (!bridgePath || !fs.existsSync(bridgePath)) die("bridge path missing: " + bridgePath);
+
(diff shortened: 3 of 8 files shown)


dug-21 (commented):
## Security Review — PR #932 (#915 C1 + #916 C15 claim-floor gates)

**Reviewer:** uni-security-reviewer (fresh context). **Risk: LOW. Blocking: none.**

Scope reviewed cold: full `main...HEAD` diff — `claim-floor-lib.sh` (Gates 9/10), `bridge-single-call-driver.js`, `release-gate-claim-floor-logic-test.sh`, the `docker-http-posture-smoke.sh` wiring, `docker-compose.yml` interpolation, `docs/client-setup.md`. All changes are CI/test-harness (bash + a node test driver) + a compose env interpolation + docs. No shipped Rust crate or edge-client (`packages/unimatrix/**`) code changes.

### What holds up
- **TLS genuinely enforced.** No `-k`/`--insecure` anywhere. `compose_health_code` pins `--cacert <served-leaf> --resolve cloud.example:8443:127.0.0.1` against the SAN — real hostname + chain verification, no insecure fallback.
- **No shell injection.** All expansions quoted; no `eval`/command-substitution in live code. `bridge-single-call-driver.js` uses `spawn(execPath, [bridge, projectHash])` (argv, no shell). `compose_extract_cert`'s `sh -c` body is a fixed single-quoted string. `projectHash` is read back from the credstore dir (init-written 16-hex SHA-256), not external input.
- **Secrets contained.** Bridge is a token-free child; failure dumps are `tail -n 40`, failure-only. `HOME=$SANDBOX/home` isolates the credstore to the hermetic sandbox (never the runner's `~/.unimatrix`); Gate-6 already asserts the token file is mode 0600. No token written to image or artifacts; `down -v` drops volumes.
- **Wrong-image / supply-chain guard is real.** `--pull never` blocks a silent `:latest` pull, and the digest assert (`docker image inspect $IMAGE {{.Id}}` == container `{{.Image}}`) compares config-digest to config-digest — same namespace, not bypassable by a tag swap; empty digests fail loud.
- **Bounded.** Gate-9 bridge drive wrapped in `timeout` (rc 124 handled); Gate-10 listener wait is deadline-bounded.
- **Append-only rc contract preserved.** New failures fold into the existing `fail()`/exit-1; no new exit codes; compose-plugin-absence is a HARD fail (not exit-3 skip) on a Docker-capable-but-mis-provisioned lane. Sourced-guard still returns before the Docker path.

### Non-blocking findings

**S1 (low) — teardown-arm ordering leaves orphans on a partial `up`.** `compose_boot_gate` sets `COMPOSE_UP=1` *after* `compose_do_up || fail`. If `docker compose up -d` partially starts containers then returns nonzero, `fail` exits before the trap is armed, so `cleanup()` skips `compose_teardown` → orphaned containers/volumes on the runner. Recommend arming `COMPOSE_UP=1` immediately *before* `compose_do_up`. Ephemeral-runner impact only.

**S2 (low) — Gate 10 publishes 8443 on 0.0.0.0.** The shipped compose maps `8443:8443`, so the gate briefly exposes the personal-cloud listener on all runner interfaces. Real ops are bearer+cert gated; `/health` leaks only version/schema. Also two concurrent `compose_boot` runs on one host collide on 8443 (`-p uni-$$` scopes the project, not the host port). Fine for a single-lane ephemeral release gate — noting the exposure/parallelism constraint.

**S3 (low) — standalone smoke contract widened.** Gates 9 and 10 run unconditionally, so the standalone `#783` smoke now hard-requires a working bridge round-trip *and* the docker-compose plugin. Intended per #915/#916, but any lane running the standalone smoke without the compose plugin will now hard-fail where it previously passed. The in-file "standalone contract is unchanged" comment refers to Gate 8 only — confirm lane provisioning.

**S4 (low) — Gate-9 `timeout` kill may orphan the bridge child.** On rc 124 the outer `timeout` SIGTERMs the driver, not the spawned bridge (separate process, no process-group kill). Bridge should exit on stdin-pipe close; minor.

### Blast radius
Worst case of a subtle bug here is a **false-green release gate** (a broken client/compose boot passing), not a production data/availability impact — this is test-only code. The digest guard + `--pull never` specifically defend against the highest-value failure (greening the wrong/`:latest` image). None of the findings enable that.

No changes requested.

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready. Two pre-tag operational items to weigh (compose-plugin runner provisioning + S1 teardown ordering); neither blocks the code.

### Vision / roadmap fit
Directly closes the two named claim-floor gaps for `goal:personal-cloud` (#5519): container-cloud HTTPS deploy smoke (C1/#915) and operator client-works verification (C15/#916). The goal entry itself names exactly these two as the remaining floor gaps that hold back an honest "deploys two ways + easy/secure client setup" claim — this is floor work, not north-star (C0 parity / multi-LLM stay future, correctly untouched). It also stays cleanly off the #837 D5 parity deadlock rather than trying to promote the permanently-red nan-021 lane, which is the right sequencing call — that lane is a separate, owned problem.

### Approach commentary
- **Rides the existing blocking lanes** (smoke-amd64/arm64) via two appended gates in the append-only exit-1 contract — no new CI wiring, no new lane to babysit. This is the lowest-blast-radius way to make the proof blocking-green.
- **The proof is relocated to where it counts.** The client-carried `context_*` round-trip existed only in a non-blocking, by-design-red lane; Gate 9 makes it blocking-green by driving the *shipped* `mcp-bridge.js` (no reimplemented transport/cert code — reuse-as-is). Gate 10 exercises the operator's literal first command (`docker compose up`) which nothing tested before — the #5248 compose-drift class that bit twice.
- **Constraints held.** `--pull never` + config-digest==config-digest assertion genuinely prevents greening `:latest` instead of the version under test (both sides read image IDs — same namespace, not tag-swappable). TLS pinned with `--cacert` + `--resolve` against the shipped SAN, no `-k`. Client-works gate uses `context_status` (stateless, embed-free) so it neither perturbs the shared credstore/cycle nor eats the embed-retry window — sensible.
- The runbook edit makes the doc's verify step the *same command the gate proves*, which is what actually retires the "docs-only, unbacked" nature of the C15 gap.

### Capability coverage
Target: C1 #5527 — Zero-friction deploy yields a working, serving instance; C15 #5570 — Operator can verify a remote client works
Archetype: both **threshold** (functional floor capabilities)  (corpus tags: `functional`, `capability`)
Coverage: **partial → whole (on green release tag)** for each — the code closes the floor; the attestation completes when the gates go green on a tag
done_when clause map:
 - C1 "clean docker run boots HTTP-on … slug write lands per-slug" → already covered (Gates 1–7, delivery:proven)
 - C1 raised floor "`docker compose up` (goal's entry promise) boots a serving HTTPS instance" → covered by **Gate 10** (compose boot + pinned /health 200 + digest match). This is the leg #915 adds ABOVE C1's literal done_when (which said `docker run`) — correctly identified by the diagnosis, not silently.
 - C15 "operator follows doc → confirms MCP round-trip" → covered by **Gate 9** + doc step 2 (identical command)
 - C15 "… + observe landing" → covered IN CI by existing Gate 7 (bundle-path observe → per-slug store), but the **doc's new verify step reproduces only the MCP round-trip, not an observe-landing confirmation** → ⚠️ minor: a literal reading of C15 ("following the doc … confirms observe landing") leaves the observe-landing confirmation without an operator step. Not a code gap; a doc-completeness nuance.
Verdict: meets the defined floor for both; the one nuance is whether C15's doc must also let the operator eyeball observe landing.
Capability-status implication (recommendation only): after a green release tag — C1 #5527 refresh `done_when`/`proven_by` (compose leg + client-carried op) and C15 #5570 flip `delivery:missing → delivery:proven`, both via `context_correct`/`context_tag` by the vision session. Do NOT flip pre-tag; evidence is the green tag run, not this merge. The PR handles this correctly (deferred + noted on both issues so the claim isn't stranded).

### Right-sizing
Outcome / DoD altitude: the property must be ENFORCED in a blocking green gate — a real client-carried `context_*` op over the pinned-TLS bundle, and a real `docker compose up` that boots+serves — not merely "a test file exists." The PR is at outcome altitude (blocking gates that fail loud), not mechanism altitude. Good.
Size verdict: **right-sized.** Two appended gates + one back-compat compose line + doc. Explicitly rejected the two over-builds (promoting the red parity lane; converting Gates 1–7 to compose) — both correctly out of scope.
Deferrals — `delivery:` tag flips (C1/C15) → **DIFFERENT gating condition (legit defer)**: the flip is evidence-gated on a green release-tag run, which this merge does not itself produce. Not the point of the work (the point is the blocking gates); pulling the flip in now would assert proven without the release-tag evidence — wrong.
Follow-up smell: none. The deferred tag flip is genuinely evidence-gated, not disguised scope.

### Assessment of the security review (not re-run)
Risk LOW / blocking none is consistent with the diff: test-harness + compose interpolation + docs, no shipped Rust/edge-client change; worst case is a false-green release gate, which the digest guard specifically defends. Of the four low findings, one is worth acting on before the first tag:
- **S1 (teardown ordering) — recommend fixing now (one line).** `COMPOSE_UP=1` is armed AFTER `compose_do_up || fail`; a partial `up` that returns nonzero exits via `fail` before the trap is armed → orphaned containers/volumes on the runner. Moving `COMPOSE_UP=1` to immediately BEFORE `compose_do_up` is strictly safer (`down -v` on a no-op is harmless, `|| true`). Cheap, and this is CI-hygiene on shared runners.
- S2/S3/S4 (port on 0.0.0.0, widened standalone contract, timeout may orphan the bridge child) — acceptable for a single-lane ephemeral release gate; no action needed beyond S3's provisioning check below.

- **Q: flip C1/C15 delivery tags with this PR?** No — defer to the next green release tag exactly as the PR does. The evidence for `proven` is a green tag run; this merge isn't one. Confirm the flips happen at retro/post-tag so the claim isn't stranded.
- **Q: is `context_status` (not an embed round-trip) enough for the client-works proof?** Yes for C15's floor — a non-error authenticated `context_*` round-trip over the pinned-TLS bundle is the "client works" signal; the embed-dependent `store→search` is already blocking-green on the same route via the embed smoke. Keeping Gate 9 embed-free is the right call for a fast posture lane.
- **Q: does the doc need an observe-landing step (C15 nuance above)?** Human's call: either accept (observe landing is background and CI-proven by Gate 7; the authenticated round-trip is the meaningful operator signal) or add a one-line "confirm your session's observe landed" to `client-setup.md` step 2 to match C15's done_when literally. Non-blocking; I lean accept.

### Recommended actions
1. **Before the next release tag — confirm the docker-compose plugin is provisioned on BOTH amd64 and arm64 release runners.** Gate 10 hard-fails (constraint 2, never self-skip) if the plugin is absent, so a mis-provisioned arm64 runner would red every release. The diagnosis flagged this as "inferred, not traced." This is the single highest-value pre-tag check. (GitHub-hosted ubuntu-22.04 ships compose v2; verify the arm64 lane's runner does too.)
2. **Consider pulling in the S1 one-line teardown-ordering fix** before merge (or as an immediate follow-up) — arm `COMPOSE_UP=1` before `compose_do_up`.
3. Merge as-is otherwise. Tester evidence (Gates 1–10 end-to-end PASS on a real Docker build from HEAD, 22/22 logic cases, pytest 32 passed) plus the LOW security verdict support merge-readiness.
4. Post-tag: execute the C1/C15 capability-map updates (vision session), and confirm the deferred flips are not forgotten.

Issue #919, opened 2026-07-05
test(proactive-delivery): end-to-end PreCompact run() restore test (PD4 floor gap)


dug-21:
Closes a **claim-floor gap** for Proactive-delivery (capability PD4 #5365). "Continuity survives context compaction" — crt-054's `compaction_events` is **producer-only boundary substrate** (records `compacted_at`; `high_water` reserved/unread), and crt-028's own ACCEPTANCE-MAP flags AC-01 **WARN: no direct end-to-end integration test of run() with a real transcript fixture** (only the assembly helpers are unit-tested).

**Acceptance:** an end-to-end PreCompact `run()` test with a real transcript fixture showing the post-compaction agent is handed restored continuity (transcript block + briefing prepended) so it **continues without re-searching**. Found: capability-graph verification pass 2026-07-05.

PR #933, opened 2026-07-07
fix(hook): replace tautological AC-SR02 test with SubagentStart injection e2e (#918)

### LINKED ISSUE #918: test(proactive-delivery): SubagentStart injection writer+routing test (PD2 floor gap)
Closes a **claim-floor gap** for Proactive-delivery (capability PD2 #5368). "Relevant knowledge is pushed without the agent asking" works in practice (dogfood-observed) but the cited AC-SR02 test is an **inline-envelope tautology** — it re-builds the JSON with `serde_json::json!` and asserts on its own literal; it never invokes the production writer (`write_stdout_subagent_inject_response`) nor drives the SubagentStart→ContextSearch→inject routing (hook.rs run() step-5b).

**Acceptance:** a test that (a) invokes the real writer and (b) drives the auto-routing path, capturing that non-empty `additionalContext` is injected with **no `context_*` tool call** issued by the agent (the 'without asking' clause). Found: capability-graph verification pass 2026-07-05.

dug-21:
Closes #918.

## Root cause
The AC-SR02 test `write_stdout_subagent_inject_valid_json_envelope` rebuilt the injection envelope with its own `serde_json::json!` and asserted on that literal — it invoked **zero production code** (delete the writer body and it still passed). As PD2/AC-SR02 evidence it was a false claim: it never exercised the real writer nor the SubagentStart→ContextSearch→inject routing.

## Fix
- **Delete** the tautological test.
- **Add** `crates/unimatrix-server/tests/hook_subagent_inject_e2e.rs` — an end-to-end test driving the real `unimatrix` binary against a fake `UnixListener`. Dual-asserts:
  - fake server received exactly one `ContextSearch{source:"SubagentStart"}` (routing happened *without the agent asking* — no `context_*` tool call in the flow), and
  - captured child stdout parses to a **non-empty** `additionalContext` envelope (the real writer emitted).
  - Neither assertion relies on exit code (the hook exits 0 on internal failure by design, #3324).
- **Add** `UNIMATRIX_HOOK_TIMEOUT_MS` env override (default 40ms preserved), resolved once and applied at **both** transport consumption sites so the e2e socket round-trip cannot be muted by a 40ms CI race. ADR #5622.

## Validation
- New e2e test: **3/3 passes, non-flaky**
- `cargo test --workspace`: green (RC=0)
- `cargo clippy --workspace -- -D warnings`: clean
- Integration smoke: 32 passed / 0 failed
- Gate: Bug Fix Validation **PASS** (7/7)

Diff is minimal — 2 files (`hook.rs` seam + new test).

### DIFF
--- a/crates/unimatrix-server/src/uds/hook.rs
+++ b/crates/unimatrix-server/src/uds/hook.rs
@@ -26,6 +26,27 @@ use crate::uds::transcript_block::{extract_transcript_block, prepend_transcript,
 /// Leaves 10ms margin in the 50ms total budget for process startup + hash computation.
 const HOOK_TIMEOUT: Duration = Duration::from_millis(40);
 
+/// Environment override for [`HOOK_TIMEOUT`] (milliseconds).
+///
+/// Operability knob for slow/loaded hosts where a legitimate injection can miss the
+/// 40ms per-read socket timeout and silently degrade (exit 0). Also lets hook e2e
+/// tests decouple correctness from the latency budget so a routing/envelope test does
+/// not double as a timing test. Precedent: `UNIMATRIX_PARITY_DIR`
+/// (parity_corpus_gen.rs). See ADR bugfix-918.
+const HOOK_TIMEOUT_ENV: &str = "UNIMATRIX_HOOK_TIMEOUT_MS";
+
+/// Resolve the hook transport timeout, honoring the `UNIMATRIX_HOOK_TIMEOUT_MS`
+/// override. Read once at hook startup and applied at BOTH transport sites
+/// (`LocalTransport::new` and `request`). Defaults to [`HOOK_TIMEOUT`] (40ms) when
+/// the var is unset, empty, or unparseable — behavior identical to the pre-seam const.
+fn resolve_hook_timeout() -> Duration {
+    std::env::var(HOOK_TIMEOUT_ENV)
+        .ok()
+        .and_then(|v| v.trim().parse::<u64>().ok())
+        .map(Duration::from_millis)
+        .unwrap_or(HOOK_TIMEOUT)
+}
+
 /// Maximum byte budget for injection output (~350 tokens at 4 bytes/token).
 pub(crate) const MAX_INJECTION_BYTES: usize = 1400;
 
@@ -251,8 +272,12 @@ pub fn run(
             | HookRequest::RecordEvents { .. }
     );
 
-    // Step 7: Connect and send
-    let mut transport = LocalTransport::new(socket_path, HOOK_TIMEOUT);
+    // Step 7: Connect and send.
+    // Resolve the transport timeout once (UNIMATRIX_HOOK_TIMEOUT_MS override, default
+    // 40ms) and apply it at BOTH sites below — LocalTransport::new sets the connect-time
+    // read/write timeout, and request() re-applies it on the racy read path.
+    let hook_timeout = resolve_hook_timeout();
+    let mut transport = LocalTransport::new(socket_path, hook_timeout);
 
     match transport.connect() {
         Ok(()) => {
@@ -268,7 +293,7 @@ pub fn run(
                     eprintln!("unimatrix: fire-and-forget failed: {e}");
                 }
             } else {
-                match transport.request(&request, HOOK_TIMEOUT) {
+                match transport.request(&request, hook_timeout) {
                     Ok(response) => {
                         // Route stdout writing based on source (ADR-006 crt-027):
                         // SubagentStart requires hookSpecificOutput JSON envelope;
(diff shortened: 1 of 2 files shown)


dug-21 (commented):
## Security Review — #933 (bugfix-918) — Risk: LOW, non-blocking

Fresh-context review of `git diff 81546ac4...HEAD`: hook.rs timeout seam + new e2e test, one deleted tautological unit test.

### OWASP scan — clean
- **Injection**: none. No shell/SQL/format-string with untrusted input. `UNIMATRIX_HOOK_TIMEOUT_MS` is parsed strictly as `u64`, so it cannot inject anything — a non-numeric value simply falls back.
- **Input validation (the one new external input)**: `resolve_hook_timeout()` is fail-safe. Verified bad inputs all fall back to the 40ms default with no panic:
  - unset / empty / whitespace / `"abc"` → `parse::<u64>()` fails → default
  - negative `"-5"` → `u64` parse fails → default
  - overflow (`> u64::MAX`) → parse out-of-range → default
  - `Duration::from_millis(u64::MAX)` cannot overflow/panic (secs = millis/1000 fits in u64)
- **`"0"` case — no hang**: resolves to `Duration::ZERO`; `set_read_timeout(Some(ZERO))` in `connect()` returns `InvalidInput` (std rejects a zero duration), so `connect()` returns `Err` → graceful-degrade arm → exit 0. It never sets `None`, so it cannot produce an infinite block.
- **Access control / trust boundary**: unchanged. Socket path derivation (`detect_project_root` + `compute_project_hash`, `$HOME/.unimatrix/{hash}/unimatrix.sock`) is untouched. The env var is same-trust-level as the invoking user; it is not a privilege boundary.
- **Deserialization**: no new production deserialization. Reads stay bounded by `MAX_PAYLOAD_SIZE` via `read_frame`.
- **Secrets**: none. Fixtures (`sess-918`, entry title/content) are benign; no tokens/keys/credentials.

### Regression risk — negligible
When the env var is unset/empty/unparseable, `resolve_hook_timeout()` returns exactly `HOOK_TIMEOUT` (40ms) — behavior byte-identical to the pre-seam const. Default hook latency path is unchanged. All `HOOK_TIMEOUT` transport sites (`LocalTransport::new` + `request`) are threaded; the const survives only as the fallback. No missed site.

### Blast radius — safe direction
Change is on the hook client socket read path. Worst case of a subtle bug is a failed/tight timeout → `connect()`/`request()` error → graceful degradation (exit 0, pattern #3324): a *missed context injection*, not data corruption, info disclosure, or privesc. The hook only reads context to inject; the timeout never gates a write of user data. Fail-safe.

### Test surface — acceptable
E2e binds a `UnixListener` under a `TempDir` `$HOME` (auto-cleaned). Accept loop is bounded (8 conns) so a routing regression cannot hang. `read_frame` bounded by `MAX_PAYLOAD_SIZE`. `.expect()`/`panic!` are test-only. No path-traversal or socket-handling risk beyond the standard `sun_path` length ceiling (a portability nit, not security).

### Informational (non-blocking) — no upper clamp on the timeout
`resolve_hook_timeout()` accepts any `u64` millis with no ceiling. A large value (e.g. `3600000`) lets the synchronous hook block far past the ADR-002 ~50ms budget and stall the agent turn. This is same-trust, self-inflicted, and is the knob's intended purpose (slow hosts), so not blocking — but a sanity cap (e.g. clamp to a few seconds) would be cheap defense-in-depth against a fat-fingered value. Optional.

**Verdict: approve. No blocking findings.**

dug-21:

**Gate**: pr-review | **Stance**: Merge as-is. The PR faithfully implements the fix-approach shape I reviewed; security is LOW/non-blocking; the one informational finding (no timeout ceiling) is optional defense-in-depth, not a merge blocker.

### Vision / roadmap fit
Serves the `proactive-delivery` claim-floor. PD2 (#5541) — "relevant knowledge is pushed without the agent asking" — was `delivery:proven` on live dogfood but carried an explicit self-declared GAP: *"no automated regression test guards this; proven-by-dogfood can regress silently."* This PR closes exactly that gap for the SubagentStart path with a real end-to-end guard. #918 is one of the two items (with #919) standing between "works in practice" and "honestly claimable + guarded" on this floor. Nothing here pulls against another goal.

### Approach commentary
The delivery matches what was approved at fix-approach, with no drift:
- Tautological `write_stdout_subagent_inject_valid_json_envelope` deleted (it invoked zero production code — delete the writer body and it still passed), replaced with a de-citation note re-pointing AC-SR02 at the new e2e.
- New `hook_subagent_inject_e2e.rs` drives the real binary and carries the **dual assertion** that was made load-bearing: (a) fake server received exactly one `ContextSearch{source:"SubagentStart", role:"developer", query non-empty}` via mpsc, AND (b) captured child stdout parses to a non-empty `additionalContext` envelope. Neither leans on exit code (hook exits 0 on internal failure, #3324) — correct, since `status.success()` proves nothing here.
- The `UNIMATRIX_HOOK_TIMEOUT_MS` seam is threaded at **both** consumption sites (`LocalTransport::new` + `request`), default 40ms preserved, ADR #5622. This was the one live decision at fix-approach and it was pulled in as recommended.
- The in-process `&mut impl Write` seam was correctly **not** added (would cover only the writer clause, not routing — redundant over-build).

The genuinely new risk locus flagged in design review — a fake `UnixListener` answering inside the hook's own read timeout, with an empty first connection from the no-op replay — is handled: the accept loop tolerates the EOF first connection and is bounded to 8 connections so a routing regression cannot hang. 3/3 non-flaky confirms the modeling holds.

### Capability coverage
Target: PD2 #5541 — "While working, relevant knowledge is pushed without the agent asking"   Archetype: **threshold** (binary; `proven` is terminal)
Coverage: **whole** for the done_when as worded, for the SubagentStart event — first automated proof of the pushed-without-asking behavior end-to-end through the compiled binary. (Original dogfood was UserPromptSubmit; this guards the crt-027 SubagentStart path. One durable representative guard on the injection pipeline is appropriate scope, not a partial.)
done_when clause map:
 - "relevant entries appear in the agent context on a hook/workflow event" → covered by assert (b) non-empty `additionalContext` envelope on captured stdout
 - "without the agent issuing any context_* command" → covered by assert (a) the ONLY request the fake server sees is hook-synthesized `ContextSearch{source:SubagentStart}` — no `context_*` call in the flow
Verdict: meets both clauses; the previously tautological evidence is replaced with a real guard.
Capability-status implication (recommendation only): PD2 **stays `delivery:proven`** — this hardens, it does not re-prove. Recommend the vision session update #5541 `proven_by` to retire the "no automated regression test" GAP note now that the SubagentStart regression guard exists. Status-tag ownership stays with the vision session / uni-capability skill — I only recommend.

### Right-sizing
Outcome / DoD altitude: **durable, trusted PD2 evidence** — a guard that stays green for the right reason and never gets muted. Not "a test file exists."
Size verdict: **right-sized.** The timeout seam is the last mile of *this* outcome (a flaky evidence test gets muted, recreating the exact "evidence that proves nothing" hole #918 exists to close) — correctly pulled in, small incremental effort, and it carries genuine production operability value (a slow host can miss a 40ms per-read timeout and silently degrade). Clear of the over-build line.
Deferrals — the optional empty-`Entries` → empty-stdout companion case was not added: LEGITIMATE (it proves a DIFFERENT branch — the silent-skip contract — and was explicitly optional; the shipped test covers the value-bearing branch).
Follow-up smell: none. No tracked follow-up is secretly the point of the work.

- **Security review informational finding — no upper clamp on `UNIMATRIX_HOOK_TIMEOUT_MS`.** My read matches the reviewer: same-trust, self-inflicted, and unbounded upward is the knob's intended purpose (slow hosts). **Do not block merge.** A sanity clamp (e.g. cap at a few seconds) is cheap defense-in-depth against a fat-fingered value that could stall an agent turn past the ~50ms budget. If the human wants it, it is a trivial follow-up — but I would not auto-file it; relay and let the human decide.

### Recommended actions
1. **Merge as-is.** Gate PASS 7/7, security LOW/non-blocking, e2e 3/3 non-flaky, `cargo test --workspace` RC=0, clippy clean, smoke 32/0.
2. Post-merge (vision session): update PD2 #5541 `proven_by` to note the regression guard now exists and retire the silent-regression GAP for the SubagentStart path. Keep the tag at `delivery:proven`.
3. Track #918 + #919 together as the pair that cleans the `proactive-delivery` claim-floor (#919 is the PD4 restore-test counterpart).
4. Optional, human's call: a small upper clamp on the timeout seam as defense-in-depth — do not file without agreement.

Issue #935, opened 2026-07-08
chore(clippy): manual_repeat_n warnings in mcp/response/verbosity.rs (rust-1.95 lint)


dug-21:
Two `std::iter::repeat(x).take(n)` sites in `crates/unimatrix-server/src/mcp/response/verbosity.rs` (~lines 208, and one more) trip the new rust-1.95 `clippy::manual_repeat_n` lint. Pre-existing / toolchain-induced — the file is untouched by feature work, and the sites are in test helpers. Surfaced during vnc-046 (#934) Stage 3c under `clippy --workspace -- -D warnings`. Trivial: `repeat_n(x, n)`. Filed to keep it off the vnc-046 PR scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #934, opened 2026-07-07
[vnc-046] Per-slug state isolation for the cloud (HTTPS) observe path


dug-21:
## Summary

Per-slug state isolation for the cloud (HTTPS) observe path. On a multi-project cloud instance each project slug is served by its own per-slug `UnimatrixServer`, but the observe/write path and several read paths silently fall back to **daemon-global** state — a per-slug **split-brain** where the object that writes and the object that reads are different instances.

This feature completes the vnc-038 per-request resolution funnel once at the seam: the resolver resolves **all** per-slug observe state (registry, pending, services), `build_project_server` reaches full construction parity (incl. the 5 config-snapshot fields), and a real boot assertion converts the whole "constructor-default field never overwritten on the per-slug path" bug class from silent-read-zero into loud-at-boot. It restores goal #5519's OSS per-project isolation invariant on the transcript, knowledge-read, and config paths where it is currently not met on cloud (HTTPS behavior == local UDS, C0 fidelity).

Governing pattern: #5629 (construction parity + funnel completeness + `Arc::ptr_eq` boot guard).

## Lineage

- **Supersedes bug #930** — the cloud/HTTPS `transcript_delta` fold defect is entirely within this feature's **P1** (the entry point; INV-T1 / AC-01). #930 will be **resolved by this feature's P1** — do not close it without a human decision.
- **#925 kept OPEN — NOT subsumed (ADR-005).** #925 (cycle-review foreign-session metrics sweep) is on a different plane (metrics-plane SQL, cross-feature within one slug) than this feature's INV-T2 (transcript-candidate plane, cross-slug). Disjoint defense-in-depth. Human owns any #925 close/keep call.
- **Depends on #800 (infra-001 multi-slug HTTP fixture) — OPEN.** The behavioral suite (AC-06) and INV-C config-parity proof **extend** the #800 fixture rather than fork one. Fixture owner unconfirmed — confirm before Session 2 delivery.

## Fix Patterns (from the #930 architect audit — 9 items, 3 patterns)

- **P1** — construct per-slug + converge write/read: `session_registry`, `transcript_hold` (paired), `pending_entries_analysis`. Closes #930 core + F2 cross-slug transcript fold. `transcript_hold` MUST pair with `session_registry` and land before the tick loop (F1/SR-03) or the purge gate splits → unbounded memory.
[...]

Issue #925, opened 2026-07-06
fix(observation): cycle-review metrics sweep whole foreign sessions on one text match — require per-record topic attribution


dug-21:
## Problem
`context_cycle_review`'s **metrics plane** (hotspots, file-zones, Phase Timeline, `mutation_spread`/`file_breadth`/`design_artifact_count`/`adr_count`, baselines / "What Went Well") can attribute **another feature's records** to the reviewed cycle.

**Observed:** a `vnc-044` review (session `50f31ee2-…`) swept in ~150 records from a *concurrent* `nxs-014` session (`33b250ce-…`) — nxs-014 ADRs, `ACCEPTANCE-MAP.md`, `issue-912-body.md` — inflating `mutation_spread` (102 files), `design_artifact_count`, `adr_count`, the file-zones, and the Phase Timeline. The transcript-candidate plane was **not** affected (it filters by `SessionState.feature`, `infra/session.rs:482`).

## Root cause — the topic predicate is dropped between the two query stages
`SqlObservationSource::load_cycle_observations` (`crates/unimatrix-server/src/services/observation.rs`) is two-stage:
- **Step 2 — session discovery** (`:397`): `SELECT DISTINCT session_id FROM observations WHERE topic_signal = ?cycle_id AND ts_millis BETWEEN start AND stop` — correctly scoped by the per-record feature tag (`topic_signal`) **and** time.
- **Step 3 — record load** (`:446`): `... WHERE session_id IN (discovered) AND ts_millis BETWEEN min AND max` — **the `topic_signal = ?cycle_id` predicate is gone.** The Rust post-filter (`:473-481`) re-checks the **time window only**.

So once a session is admitted by **one** matching record in Step 2, **all** of its in-window records are loaded regardless of each record's own `topic_signal` — an all-or-nothing per-session sweep.

`topic_signal` is content-derived per observation (`crates/unimatrix-observe/src/attribution.rs::extract_topic_signal:84`, written at ingest in `uds/hook.rs:377`).

## Requirement (what the fix must enforce)
A record must count toward a cycle's metrics **only if that record itself is in-topic** — i.e. attribution must be **per-record by `topic_signal` (feature-switch partitioning)**, NOT whole-session admission because the session emitted one text match for the feature id. Session-level admission-by-text is the defect; membership must be evaluated per record, by topic.

## Trigger (both must hold)
[...]

Issue #878, opened 2026-07-02
investigate: full `cargo test --workspace` link-step OOM (ld SIGKILL) — regression, not 'environmental'


dug-21:
## Symptom
Full `cargo test --workspace` fails at the **link step** — `ld` killed with signal 9 (OOM) — on several integration-test binaries. Named in two recent PRs: `import_integration`, `export_integration`, `project_routing_integration`, `graph_subgraph_integration` (all `crates/unimatrix-server/tests`). Observed on **#873** (rc=101, "swap exhausted") and **#877** (ld signal 9). In both, coverage was salvaged via `--lib` + Python smoke, so the full-workspace integration link is currently **unverified anywhere** (GH CI is JS-client-only).

## Why this is NOT just "environmental"
It worked for a long time and regressed recently. The dev-container hasn't changed. What *has* grown is the link-memory footprint:
- **No `[profile.*]` overrides** in root `Cargo.toml` and **no `.cargo/config` linker config** — every integration binary links with default **`debug = 2`** (fat debug info, the dominant driver of linker peak RSS) via stock **BFD `ld`** (memory-hungry), with **no mold/lld**.
- **~41 integration-test binaries** link independently (store: 23, server: 10, engine: 4, observe: 3, learn: 1). `cargo test` links multiple concurrently → peak RSS = (parallel link jobs) × (per-binary debug-link RSS).
- As test surface / dependencies grew, concurrent peak link RSS crossed a **fixed** memory ceiling. That's a footprint regression, not a shrinking environment.

## Investigate (find what changed)
1. **Bisect to last-green.** When did full `cargo test --workspace` last link clean? Correlate the first-red commit with: new integration binaries, a dependency bump, or a codegen/profile change. This is the load-bearing question — name the change.
2. **Measure.** Peak `ld` RSS per failing binary (`/usr/bin/time -v` on a single `cargo test -p unimatrix-server --test import_integration --no-run`). How far over the ceiling, and is it one fat binary or N-parallel-links summing?
3. **Rule out the box** (human says unchanged): container/cgroup memory + swap limits (`cat /sys/fs/cgroup/memory.max`, `free -h`), and whether link `--jobs`/`NUM_JOBS` default changed.
4. **Debug-info growth.** Did something recently pull heavy generic monomorphization into these bins, or bump a dep that balloons debug info?

## Candidate mitigations (for the fix session — do not pre-decide here)
[...]

PR #936, opened 2026-07-08
[vnc-046] Per-slug state isolation for the cloud (HTTPS) observe path

### LINKED ISSUE #930: fix(transcript): cloud/HTTPS transcript_delta 204-ACKed but never lands in the cycle_review fold
## Summary
On a personal-cloud (HTTPS) instance, transcript_delta frames are accepted by the
server (HTTP 204) but never surface in the transcript buffer that cycle_review folds.
Structured observation (reads, phase timeline, tool counts, cycle attribution) works;
the raw transcript fold is empty ("Transcript bytes: unavailable (no data)", no
candidates). Local/UDS mode folds the transcript correctly — this is an HTTPS-path
fidelity gap (dents C0 #5594; SL-OBSERVE #5580 transcript leg).

## Client is exonerated (external-instance investigation)
- Delta leg is wired: the delta rides as a second concurrent POST from every
  fire-and-forget spawn carrying a non-empty transcript_path (SessionStart, Stop,
  UserPromptSubmit, PostToolUse, PostToolUseFailure, PreToolUse(context_cycle)) —
  index.js:280-286 -> delta.maybeSendDelta(), shipping [last_offset, file_len) as one
  transcript_delta RecordEvent.
- Deltas are emitted and accepted: live probe on the exact production path returned
  {send:{ok:true, status:204}}. Offset file advanced to 474163 == transcript byte size
  (writeOffset fires only on 2xx). health.json: last_success recent, last_failure null,
  consecutive_failures 0, queue_depth 0.
- session_id parity holds: delta path (sessionIdOf) and cycle-stamp path
  (decorateCycleStamp) both key on resolveSessionId -> raw CC UUID; offset files are
  named by the same UUIDs cycle attribution uses. #832 fix functioning.
- observe_url identical to the working observation events; no separate flag/toggle.

## Server-side hypotheses (where to look)
1. Fold key mismatch: the transcript_delta frame carries session_id but NO cycle_stamp
   (decoration mutates only the carrying event, not the separately-assembled delta
   frame). If the fold requires a per-frame cycle_stamp instead of joining
   delta->session->cycle, the buffer reads empty despite 204s.
2. Buffer not persisted across HTTP requests: local/UDS holds the per-session
   TranscriptBuffer on long-lived SessionState; over stateless HTTP the delta POST and
   the later cycle_review fold are separate requests. If the HTTP path lacks a
   persistent per-session buffer registry, apply_delta runs on a buffer discarded at
   204. (Same family as vnc-027 session-split.)

## Impact
- C0 (#5594) — HTTPS != local for the transcript leg; a production personal-cloud
  instance cannot stream/fold raw transcript activity.
- SL-OBSERVE (#5580) — the transcript leg of the unified observation stream is absent
  in cloud mode; cycle_review transcript candidates are always empty.

## Repro
Personal-cloud HTTPS instance; run a real session; confirm transcript_delta POSTs 204;
run cycle_review — transcript fold is empty despite advancing offsets.

dug-21:
## Summary
Makes cross-project contamination on the cloud (HTTPS) observe path **structurally impossible** by completing the vnc-038 per-request resolution funnel once at the seam. Originated as bug #930 (transcript_delta 204-ACKed but never folded); the audit revealed a 9-item per-slug state split-brain across 3 fix patterns, so it became a design→delivery feature (#934).

**Root cause:** state that lives daemon-global in `ObserveContext` got constructor-defaults on per-slug `UnimatrixServer` instances — per-slug MCP reads (transcript fold, knowledge briefings, config) saw wrong/empty state. Server-side only; the slug already rides the URL (`/v1/{slug}/observe`), no wire/client/protocol change.

## What changed (3 patterns + guardrail)
- **P1** — per-slug `session_registry` + `transcript_hold` (paired) + `pending_entries_analysis` + per-slug signature scanner → **fixes #930** and the cross-slug transcript-fold leak.
- **P2** — `ObserveContext.services` resolved per-request → closes the **cross-project knowledge-read leak** (privacy/security floor; vnc-038 had only closed the write funnel).
- **P3** — 5 config-snapshot fields overwritten per-slug (config parity).
- **Vestigial** — deleted `ObserveContext.vector_store`/`adapt_service` (~87 `dispatch_request` call sites updated).
- **Guardrail seam** — resolution methods on `StoreResolver` (no side-map, no default impl); a **release** boot assertion (`assert_per_slug_isolation`) + a **compile-time exhaustive field census** (no `..`) that fails compilation until any future field is classified — converts the whole "constructor-default never overwritten" class from silent-read-zero into loud-at-boot.

## Durable guardrail (the acceptance spine)
Solution-independent **bidirectional** behavioral isolation suite, N≥2, asserted only through `/v1/{slug}/…` — slug B can never observe slug A's transcript/knowledge/config and vice-versa, plus a **negative-control test** that goes RED on a mis-route. Any future tech/wiring change that breaks isolation trips it.

## Gates & tests
- Gate 3a PASS (1 rework: per-slug scanner) · Gate 3b PASS · Gate 3c PASS
- Behavioral isolation suite **30/0** · Full workspace `cargo test` **6982/0** · Integration smoke **35/0** · #800 multi-slug HTTPS suite **4/0** · #878 link smoke PASS

## Lineage / linked issues
- **Fixes #930** (its P1 entry point; INV-T1 proves the fold non-empty).
- **#925 stays OPEN** — different plane (metrics-plane cross-feature SQL vs this transcript-candidate cross-slug); per ADR-005 it is **NOT subsumed** — do not close as such.
- **#800** multi-slug HTTP fixture built in-scope (extended infra-001, not forked).
- **#935** — two pre-existing `manual_repeat_n` clippy warnings in `verbosity.rs` (rust-1.95 lint, untouched file); filed separately, not in this PR's scope.

## Pre-merge CI action
The live `cycle_review`-based legs (AC-07 HTTPS==UDS parity, OQ-2 count>0, R-16) were deferred to the **infra-003 Docker gate** rather than shipping a flaky fake-green local check (#4452). Recommend running that lane pre-merge. Isolation itself is fully proven (in-process 30/0 + real-HTTPS #800 4/0).

Full design + gate reports under `product/features/vnc-046/`.


dug-21 (commented):
## Security Review — vnc-046 (fresh context)

**Risk level: LOW · No blocking findings.** This change *strengthens* cross-tenant isolation; it removes no validation and adds no external surface.

### Cross-tenant mis-route (the feature's core purpose) — structurally closed
The per-slug funnel cannot resolve to the wrong slug's state:
- `resolve_store` + the three new `*_for` methods all do `self.slugs.get(s)` against the **same** map, keyed by a `ProjectKey` parsed **once** in `route_observe` (Step 0/0b, same `&key`). No second key derivation to drift.
- Within an entry, `session_registry`/`pending`/`services` are `Arc::clone`d off the one `server` **before** it moves into `McpAdapter::new` (`ProjectEntry::from_server`) — write instance == read instance by construction.
- Defense in depth is real, not decorative: boot `Arc::ptr_eq` convergence per slug (`assert_per_slug_isolation`, aborts boot — not `debug_assert`), a **non-`cfg(test)`** compile-time exhaustive field census (no `..`) that fails the *release* build on any new unclassified field, and an N=2 **bidirectional** behavioral suite through the assembled `PathRouter` edge with a negative-control that proves the isolation predicate is non-vacuous (guards the #5348 one-directional false-GREEN).
- Test doubles all implement the no-default `*_for` methods **fail-closed** (`UnknownProject`), so a lenient harness cannot re-admit the split-brain.

### OWASP
- **Access control / tenant isolation** — primary concern; addressed above. P2 knowledge-read leak closed by `services_for` resolving the per-slug `ServiceLayer` (was a daemon-global). INV-K2 asserts durable non-contamination bidirectionally.
- **Path traversal** — slug still validated by the `ProjectSlug` newtype before the single `base_dir.join(slug)` site; unchanged. No new join surface.
- **Injection / deserialization** — no new parsing; observe payload path unchanged.
- **Input validation** — `resolve_slug_config` is a pure code-move to `slug_config.rs`; the 64 KiB cap and `#[cfg(unix)] 0o022` permission check are preserved verbatim.

### Regression (NFR-4, UDS/stdio unchanged)
`dispatch_request` dropped two params that were **already** `_`-prefixed/unused (`vector_store`, `adapt_service`); `handle_connection` still receives them (renamed `_`), so accept-loop wiring is structurally identical. The ~87 call-site edits are mechanical arg removals, compile-enforced. Zero behavioral delta on the local path.

### Dependencies / Secrets
- No new supply-chain surface: the `tower` **dev**-dependency added is the same crate/version/features already present as a production dependency.
- No hardcoded secrets, tokens, or keys in the diff.

### Finding 1 (LOW, non-blocking) — boot assertion checks global, not per-slug, retention
`assert_per_slug_isolation` gates on `config.retention.transcript_hold_max_sessions == 0` using the **global** `UnimatrixConfig`, but `transcript_hold_max_sessions` is per-slug overlayable (project-wins merge, `config.rs:4332`) and the hold in `build_project_server` is built from the **slug's resolved** value. A slug that overlays `transcript_hold_max_sessions = 0` (global non-zero) would build a zero-capacity hold yet evade this specific sub-check.

- Blast radius: **not** a tenant-isolation issue — no cross-slug data surfaces. Worst case is the intended loud-abort not firing for a per-slug-disabled hold (a purge/memory-hygiene guard, degrades to the pre-feature behavior for that one mis-configured slug). The `has_hold` and `Arc::ptr_eq` convergence checks still use the correct per-slug handles.
- Recommendation: pass the slug's resolved retention (or capture the resolved `transcript_hold_max_sessions` into the `IsolationProbe`) so the zero-check matches the value the hold was actually built with. Or drop the sub-check if 0 is a legitimate per-slug choice.
- Blocking: no.

Reviewed cold against the diff, ARCHITECTURE.md, and RISK-TEST-STRATEGY.md. Approving from a security standpoint.

dug-21:

**Gate**: pr-review | **Stance**: Merge-ready on the isolation mechanism; make the infra-003 Docker gate GREEN a pre-merge condition and do NOT close #930 until that live-fold leg is green — the in-PR proof stops one layer short of #930's literal symptom.

### Vision / roadmap fit
Squarely on `goal:personal-cloud` (#5519). It restores a *core success criterion* — "each project fully isolated (own DB, vector index, hash chain, analytics; no cross-project sharing)" and "cross-project contamination structurally impossible" — on the cloud observe/knowledge-read/config paths where it was **silently not met** on multi-slug HTTPS. This is remediation of the gap vnc-038 left (it isolated only the store *write* funnel); vnc-046 completes the funnel to full observe-state parity (registry, pending, services, 5 config-snapshot fields) beside `resolve_store` on the same `StoreResolver`. No conflict with in-flight work; the #800 multi-slug HTTPS fixture was built in-scope (dependency resolved, not forked). The scope growth from bug #930 → 9-item feature is proportionate, not gold-plating: the audit found a whole "constructor-default-never-overwritten" class, and the compile-time field census + release boot assertion convert that class from silent-read-zero to loud-at-boot — defense the vision explicitly sanctions ("structurally impossible"), not incidental over-build.

### Approach commentary
- The **class guard** (exhaustive no-`..` field census + `Arc::ptr_eq` boot assertion that aborts a real `Result`, not `debug_assert`) is the right altitude: it guards the goal-level integrity invariant, not a one-off. Keep it.
- The **bidirectional N=2 suite + negative control** is the correct answer to lesson #5348 (a one-directional probe false-GREENs the reverse mis-route). Good.
- **Correct call NOT to ship a flaky local `cycle_review` check** (anti-fake-green #4452). The deferral is honest and named. My push below is only that the named vehicle must actually run before the outcome is claimed.

### Capability coverage
Target: N3 #5593 — Writes never mis-routed across projects   Archetype: threshold (nfr)   (corpus tags: delivery:proven)
Coverage: threshold → the feature EXTENDS N3's proven surface — observe transcript-fold + knowledge-read + config isolation, previously silently un-isolated on multi-slug, now proven bidirectional N=2. N3 done_when clause (2) requires this be "re-proven on every release by a standing blocking gate (RED fails the manifest)" — that gate is infra-003; see gap below.
Also-target: C11 #5554 (remote observe per-slug, threshold, delivery:proven) — its done_when "server records them **under the slug**, not a UDS fallthrough" was in fact violated on the multi-slug fold path (#930); this feature makes C11 genuinely true on multi-slug. C11 proven_by standard is **live** ("server logged transcript_delta under http- session").
Also-target: C4 #5533 (isolation seam, threshold) — maintained/extended; one funnel preserved (`*_for` beside `resolve_store`, no side-map). Stays proven.
Also-target: C0 #5594 (marquee fidelity ROLLUP, **curve**, delivery:partial) — ADVANCES dimension D2 (behavioral observe→topic_signal) on the multi-slug cloud path. A curve is NEVER "whole"; do not fault for incompleteness. Isolation itself is excluded-by-design from C0 (D6 removed, #845) and lives under N3 — correct.
done_when clause map:
 - N3(1) "write for A lands only in A's store, across observe + MCP-write" → covered in-process (30/0) + real-HTTPS observe (#800 4/0); MCP-read fold surface → DEFERRED to infra-003 wire
 - C11 "records under the slug (fold non-empty)" → covered at the per-slug store-row layer (AC-01 `test_observe_transcript_fidelity_a/_b`); ⚠️ the literal fold-into-`cycle_review` over the live wire (the #930 symptom) → DEFERRED to infra-003 (named gap, declared)
 - N3(2) "standing blocking release gate" → infra-003/infra-004 gate exists; must run GREEN, not INFRA-dark, for these newly-covered sub-states
Verdict: Meets the defined isolation mechanism. The one honest, **declared** gap is the *live* outcome-proof of the #930 fold — deferred to the Docker gate, not run in this PR.
Capability-status implication (recommendation only): keep N3 / C11 as-is; do NOT let #930's close or any re-assertion of these capabilities rest on the in-process store-row proof — their evidence standard is **live**. Flip nothing until infra-003 is green on the observe-fold + MCP-read legs. (Recommendation only — capability-map mutation is the vision session's call.)

### Right-sizing
Outcome / DoD altitude: the property that must WORK — a delta POSTed to `/v1/{A}/observe` is *folded into A's `cycle_review`* and never B's, **over the real HTTPS wire**, with `signal_class_counts > 0`. Not "the per-slug store row exists."
Size verdict: right-sized BUILD (mechanism, guardrail, and isolation all proven). The **outcome evidence** stops one layer short — in-PR proves store rows + ptr-identity + real-HTTPS observe isolation, but the live `cycle_review` fold (AC-07 parity + OQ-2 count>0) is deferred. Structurally the fold reads `services_for` → the same per-slug store, so the mechanism argument is strong; but #930 is exactly a "204-ACK masks persists-nothing" class, so the residual risk lives in the fold/serving path that is not exercised in-PR.
Deferrals:
 - AC-07 live HTTPS==UDS `cycle_review` parity + OQ-2 count>0 → **LAST MILE of the #930/C11/N3 outcome** (same outcome, deferred vehicle — not a different one). Incremental effort: **small** (run the existing infra-003 Docker gate). Recommend: make it a pre-merge gate, not a post-merge nicety.
 - Security Finding 1 (boot assertion checks *global* not per-slug-resolved `transcript_hold_max_sessions`) → last mile of the feature's OWN guardrail correctness; effort small. Recommend pull in now or file with a named owner.
 - arm64 (N3 D-3) → DIFFERENT scope, legit defer (routing is architecture-independent Rust).
 - #925 (metrics-plane, cross-feature within one slug) → DIFFERENT plane, legit keep-open (ADR-005).
Follow-up smell: the "run infra-003 pre-merge" recommendation in the PR body IS the point of the #930 fix (proving the fold), and the gate already exists — pull it into the merge condition rather than tracking it as follow-up.

1. **Close #930 now?** No. Fix it in the merge but close it only after infra-003 shows the observe delta folding non-empty into `cycle_review` over the live wire. The PR already flags "do not close without a human decision" — my recommendation is: the decision is "green Docker fold-leg first."
2. **AC-07 / OQ-2 deferral acceptable?** Yes as a *deferral of vehicle*, not of outcome. Acceptable to merge with it deferred ONLY if the infra-003 gate is run green pre-merge (or the release manifest gate is confirmed blocking-and-not-INFRA-dark for these legs). Do not treat the in-process pins as the outcome proof for a capability whose standard is live.
3. **Security Finding 1 (LOW) — fix or defer?** Recommend fixing in this PR: it is a correctness hole in the boot assertion that is this feature's headline guardrail, the fix is small (feed the slug's *resolved* retention into `IsolationProbe`), and a per-slug `transcript_hold_max_sessions=0` overlay currently evades the loud-abort → the exact unbounded-memory/purge-split hazard (F1/SR-03) this feature exists to prevent. If not fixed here, file with a named owner and reference SR-03.
4. **Two pre-existing `verbosity.rs` clippy `-D warnings` blockers (#935)?** Not vnc-046's scope; correctly filed separately. Human decision — they block `-D warnings` on `main` too, so worth a fast separate fix, but not a reason to hold this PR.

### Recommended actions
1. Run the **infra-003 Docker multi-tenant-isolation gate GREEN** in the pre-merge lane (observe-fold + MCP-read 2×2). Confirm it is not INFRA-dark (N3's OQ-3/R-15 silent-vacuous residual).
2. **Do not close #930** until that fold-leg is green; then close with the live evidence linked.
3. Decide on **Security Finding 1** — recommend fixing in-PR (small, it is the guardrail's own correctness).
4. Merge once (1)+(3) are resolved. Keep **#925 OPEN** (ADR-005, not subsumed) and leave #935 on its own track.
5. Capability map: after infra-003 is green, the vision session can re-affirm N3/C11 on the live evidence — reviewer does not mutate the map.

dug-21:
Security review Finding 1 (LOW) is fixed in `b4b6aa88`: `assert_per_slug_isolation` now captures the slug's **resolved** `transcript_hold_max_sessions` into `IsolationProbe` and the zero-check reads that per-slug value instead of global config. New test `test_assert_per_slug_isolation_per_slug_zero_hold_returns_err` closes the overlay-`0` evade path. Boot-assertion suite 6/6, workspace clippy clean.

dug-21:
## infra-003 multi-tenant isolation Docker gate — GREEN ✅

Ran on branch HEAD (`b4b6aa88`, incl. the security fix). Full release image built from the branch `Dockerfile`, 2-slug container, real HTTPS.

```
[infra003-smoke] warmup PRESENT — embedding path warm + store durable; proceed to matrix.
[infra003-smoke] observe: A has obs-a not obs-b; B has obs-b not obs-a => observe GREEN
[infra003-smoke] mcp: A has mcp-a not mcp-b; B has mcp-b not mcp-a => mcp GREEN
[infra003-smoke] ALL GATES PASSED — bidirectional 2x2 isolation holds on both surfaces.
```

This closes the live `cycle_review`/serving-path leg deferred in-PR (AC-07 parity + OQ-2 count>0 class): warm embedding path + durable store over the real wire, bidirectional on both observe and MCP-write surfaces. **#930's literal symptom is proven gone end-to-end.** exit 0 (GREEN).

Issue #945, opened 2026-07-10
Retire the Rust hook client (uds/hook.rs) now that the JS client is the only live client


dug-21:
## Context
Surfaced during #944. The install has fully migrated to the JS hook-client; `crates/unimatrix-server/src/uds/hook.rs` is no longer on any live path — its only production entry is the `unimatrix hook` CLI subcommand (`main.rs:551-555`), a no-DB process that connects OUT to the server socket (a client, not inbound processing). The live path is JS client → UDS → `listener.rs` → store; the server never re-parses `tool_input`.

## Problem
vnc-047 (#940/#943) spent implementation and test effort enhancing `hook.rs` with the tags block — code that is no longer in live use. We enhanced and tested a retired client.

## The tension to resolve
`hook.rs` is not purely dead: it currently serves as the **parity oracle** — `parity_corpus_gen.rs` runs its `build_request` as 'the oracle pipeline' that the JS client's payloads are tested against. Retiring it is therefore not a free delete. Options to weigh:
1. **Remove it** — and replace the parity oracle with a spec/golden-based source of truth (the JS client validated against fixtures, not against a second implementation).
2. **Keep it, but reclassify** — stop treating it as a shippable/enhanceable client; freeze it as oracle-only with a clear header, so future interface changes aren't duplicated into dead client code.
3. **Status quo** — accept the double-maintenance cost.

## Ask
Decide whether to retire the Rust hook client and, if so, how to preserve (or replace) the parity oracle. Design-session candidate, not a bugfix.

## References
Surfaced in #944 · vnc-047 #940 · #943 · parity oracle: `parity_corpus_gen.rs:167-180`

dug-21:
I'm fine to keep it but needs to be labeled that changes to hook don't ACTUALLY modify production Unimatrix behavior cuz its no longer wired.  Likely just need proper labeling.

PR #946, opened 2026-07-10
fix: persist context_cycle tags via the JS hook-client + delete lying MCP ack (#944)

### LINKED ISSUE #944: bug(context_cycle): bare MCP call silently accepts and DROPS tags (accept-and-drop footgun)
## Summary
context_cycle called as a bare MCP tool with `tags=[...]` **silently accepts and drops** the
tags: the param is accepted and the call returns success, but nothing persists, so the tags
never appear in context_cycle_review. Tag persistence is hook-path only (vnc-047 / #940,
shipped #943). An agent/user who sets a run tag via the MCP tool call gets no error and no
tags — a silent footgun.

## Evidence (from #943)
- vnc-047 persists tags + surfaces them in the review on the HOOK path only. The bare MCP
  context_cycle handler accepts the `tags` param (interface stability, AC-06) but persists
  nothing.
- Codified as intended in the suite: `test_bare_mcp_cycle_tags_not_persisted` and
  `test_context_cycle_accepts_tags_param` (infra-001) assert the bare MCP path returns
  success and persists no tags (AC-EXTRA-1). So the suite is GREEN on the wrong behavior.
- The review-surfacing itself is correct and assembled-path-proven
  (`test_review_surfaces_tags_json_and_markdown_assembled`) — but only over the hook path.
  The gap is the SET path, not the read.

## Why it's a bug
Silent accept-and-drop is wrong on any path. Determine the right behavior:
1. **Persist on the MCP path too** — context_cycle(start, tags) works regardless of path; OR
2. **Reject / fail-loud on the MCP path** — do not accept a param you will drop (fail loud,
   never silent — consistent with the project's presentation-honesty / fail-loud posture).

Original design intent (per human recollection) was NOT to accept tags on the MCP path at all
— i.e., a rejection, not a silent accept. Investigate, decide "the right way," and fix.

## Scope
- context_cycle bare MCP handler tag behavior (SET path). The review-read is already correct
  — do not touch it.

## References
vnc-047 · #940 (feature) · #943 (delivery) · AC-EXTRA-1 · ADR-006 (deferred-mutation home).

dug-21:
## Root cause (v3, authoritative)
vnc-047 added a \`tags\` field to \`context_cycle(start)\`, meant to ride the hook and persist server-side exactly like \`goal\`. Two defects shipped instead:

1. **Tags never reached the server.** The deployed **JS hook-client** (\`build-request-tools.js\`) is *field-selective* — it builds the outbound payload from a whitelist and was given a \`goal\` block but never a \`tags\` block. So \`insert_cycle_start_with_tags\` was never reached and \`cycle_tags\` stayed empty in **every live session**. (The Rust \`hook.rs\` tags code exists but is only the parity *oracle*, not on the live path.)
2. **The MCP handler lied.** \`cycle_tag_ack_phrase\` claimed tags were "accepted for recording" for input the handler dropped — a silent accept-and-drop.

The server write/read path and the whole-set-once freeze guard (keyed on \`cycle_tags\` rows) were already correct and are **untouched**.

## Fix
- **JS** (\`packages/unimatrix/lib/hook-client/build-request-tools.js\`): forward a \`tags\` key on cycle_start — start-only, strings-only, omit-when-empty, **infallible** (malformed tags degrade to no-key and can never drop the cycle event). Mirrors the \`hook.rs\` Step 4c oracle. +5 unit tests.
- **Rust** (\`crates/unimatrix-server/src/mcp/tools.rs\`): delete \`cycle_tag_ack_phrase\` + call site; keep the load-bearing \`CycleParams.tags\` field. Removed 4 obsolete ack tests, added a negative-ack test.
- **Parity gate** (\`parity_corpus_cases_b.rs\` + goldens): add \`cycle::tags\` cases — the guard whose absence let this ship. The start-with-tags golden now contains the \`tags\` key.
- **Test cleanup**: deleted obsolete Python \`test_context_cycle_ack_echoes_tags\`.

## Semantics (confirmed with owner)
Write-once-set per feature cycle: an empty first start does **not** lock; a later start carrying tags writes and locks; once a non-empty set exists it is frozen (no add/modify). ADR-002 unchanged — the fix just makes the already-shipped semantic *reachable*.

## Verification
Phase 3 PASS — Rust 4542, JS 779, parity 24/24, \`cargo test --workspace\` rc=0, clippy clean, smoke 35/35. Gate: Bug Fix Validation **PASS (7/7)**.

## Delivery follow-ups
- ADR-007 #5659 (ack-echo clause) to be \`context_correct\`'d — withdraw only the ack-echo part.
- Post-merge: rebuild server binary + refresh \`~/.unimatrix/dogfood-client\` (stale-binary rollout).
- Follow-up #945: retire the now-oracle-only Rust hook client.

Closes #944.

### DIFF
--- a/crates/unimatrix-server/src/mcp/tools.rs
+++ b/crates/unimatrix-server/src/mcp/tools.rs
@@ -542,40 +542,14 @@ pub struct CycleParams {
     /// starts are a whole-set no-op. Values are opaque — stored verbatim, non-empty is
     /// the only check (no vocabulary/length/prefix validation). Old callers omitting this
     /// field receive None. Persistence rides the hook path (not this handler); this field
-    /// declares the interface and feeds only the best-effort ack echo.
+    /// only declares the interface — frame writers read `tool_input.tags` from this
+    /// schema-declared param exactly like `goal`. The MCP handler does nothing extra for
+    /// tags: no recording claim, no read-back (parity with `goal`).
     pub tags: Option<Vec<String>>,
     /// Agent making the request.
     pub agent_id: Option<String>,
     /// Response format: summary, markdown, or json.
     pub format: Option<String>,
-}
-
-/// Build the best-effort tag-ack phrase appended to the `context_cycle` response (vnc-047
-/// C12, ADR-007). NON-GATING: echoes the CALLER's own input — never reads stored `cycle_tags`
-/// and never surfaces the whole-set-once freeze outcome (Non-Goal #6). Applies the same
-/// non-empty filter as the write path (value-opaque; no other validation). An empty/all-blank
-/// list yields an empty string, leaving the base ack unchanged. Wording is accept-for-recording
-/// (fire-and-forget), NOT a durability guarantee — `context_cycle_review` is authoritative.
-/// `pub(crate)` for unit testability without full handler construction.
-pub(crate) fn cycle_tag_ack_phrase(cycle_type: CycleType, tags: Option<&[String]>) -> String {
-    let tag_view: Vec<&str> = tags
-        .unwrap_or(&[])
-        .iter()
-        .map(|s| s.trim())
-        .filter(|s| !s.is_empty())
-        .collect();
-    if tag_view.is_empty() {
-        String::new()
-    } else if cycle_type == CycleType::Start {
-        format!(
-            " {} run-identity label(s) accepted for recording at cycle start: [{}]. \
-             Recording is fire-and-forget; use context_cycle_review to confirm.",
-            tag_view.len(),
-            tag_view.join(", ")
-        )
-    } else {
-        " tags ignored — labels are only recorded at cycle start.".to_string()
-    }
 }
 
 /// Extract the active workflow phase for a session — infallible, O(1) clone.
@@ -4202,7 +4176,11 @@ impl UnimatrixServer {
             }
         }
 
-        let base_ack = if let Some(ref g) = validated_goal {
+        // tags parity with goal: the schema-declared `params.tags` is read by frame
+        // writers off `tool_input.tags` (the hook path persists it). The MCP handler does
+        // NOTHING extra for tags — no recording claim in the ack, no read-back (vnc-047 fix,
+        // #944). `context_cycle_review` is authoritative for what actually persisted.
+        let response_text = if let Some(ref g) = validated_goal {
             format!(
                 "Acknowledged: {} for topic '{}' with goal: '{}'. \
                  Attribution is applied via the hook path (fire-and-forget). \
@@ -4217,12 +4195,6 @@ impl UnimatrixServer {
                 action, validated.topic
             )
         };
-
-        // 4c. vnc-047 (C12, ADR-007): best-effort, NON-GATING tag ack echo on the EXISTING
-        // ack string (no new interface, no read-back API). Echoes the CALLER's own input —
-        // never reads stored cycle_tags, never surfaces the freeze outcome (Non-Goal #6).
-        let tag_phrase = cycle_tag_ack_phrase(validated.cycle_type, params.tags.as_deref());
-        let response_text = format!("{base_ack}{tag_phrase}");
 
         // 5. Audit log (fire-and-forget)
         let metadata_json = match ctx.client_type.as_deref().filter(|s| !s.is_empty()) {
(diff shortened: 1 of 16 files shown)


dug-21 (commented):
## Security Review — PR #946 (fresh context)

Reviewed the full diff cold, the modified source files, and the untouched server write path.

### Injection (OWASP A03) — clear
- `cycle_tags` INSERT (`unimatrix-store/src/db.rs:456-458`) is fully parameterized via `.bind()`; no string interpolation. It is **not** in this PR's changed-file set — confirmed unchanged.
- The JS block builds no query/command from `tags` — it filters strings into a JSON payload key. No injection surface introduced.
- Read path (`get_cycle_tags`, `ORDER BY tag`, parameterized) also unchanged.

### Infallibility of the JS tags block — confirmed
`tiObj` is normalized to a plain object at :280-281, so `tiObj.tags` cannot throw. `Array.isArray` + the `typeof t === "string" && t.trim() !== ""` short-circuit make every op inside safe, and the whole thing is wrapped in try/catch that degrades to `tags = null`. A throw is impossible in practice and would be caught regardless — so a malformed `tags` can never drop the cycle frame (topic + goal survive). Verified by `test_cycle_tags_malformed_never_throws_and_keeps_goal`.

### Regression risk — low
- Ack deletion removes only an appended phrase; the topic/goal `format!` ack strings are byte-identical. No change to topic/goal handling.
- The tags block is appended after `put("goal")` and only sets a new key; feature_cycle/phase/outcome/next_phase/goal insertion order and values untouched. cycle-stop golden updated to prove tags-on-non-start don't persist.

### Blast radius — fail-safe
Worst realistic case of a subtle bug here is tags silently not persisting (regression to the original #944 symptom) — a correctness issue in the fail-safe direction. No data-corruption/disclosure path: the write is transactional (`BEGIN IMMEDIATE` + whole-set-once freeze), value-opaque, single-tenant.

### Note (non-blocking): deliberate no-cap on tags
`tags` has no count cap and no per-tag byte cap, unlike `goal` (`MAX_GOAL_BYTES`). Acceptable in the current 1-client:1-project self-supplied trust model — a caller can only bloat their own single-tenant DB, and the whole-set-once freeze makes it a one-time-per-cycle write. The effective ceiling is the 1MB hook stdin cap (`MANIFEST stdin_cap_bytes`). Suggest a one-line comment recording that the missing cap is intentional (oracle parity) and that the stdin cap is the backstop, so a future reader doesn't "fix" it and break parity. Not required for merge.

### Also verified
- Deserialization: no new surface (existing stdin/serde frame paths).
- Dependencies: no Cargo.lock / Cargo.toml / package.json / package-lock.json in the diff — zero new deps, consistent with the PR claim.
- Secrets: none in the diff; fixtures use synthetic values.

Approve from a security standpoint.

dug-21:

**Gate**: pr-review | **Stance**: Recommend merge. The v3 fix is the correct, minimal restoration of a shipped-but-dead capability; JS port matches the Rust oracle exactly and the missing parity gate is added. Two post-merge conditions are load-bearing (dogfood-client refresh; ADR-007 scope — the latter appears already done). One residual: true production proof is deferred to the post-merge client refresh.

*(Independence note: spawn prompt was clean — agent ID, gate, issue, PR, review inputs only, no conclusions injected. I verified the JS↔oracle parity by reading `hook.rs:860-917` against the diff, and confirmed the ADR-007 correction from the corpus (#5675) rather than taking the gate report's word.)*

### Delivery as a whole
This PR resolves a three-round diagnosis (v1 disclosed-drop and v2 both superseded). The v3 root cause is right and now visible in the diff: vnc-047 shipped every Rust server layer for run-identity `tags` but (a) never ported the field to the deployed **JS** hook-client, so `cycle_tags` stayed empty in every live session, and (b) grew an MCP ack that claimed "accepted for recording" for input the handler drops. The fix ports Step 4c to JS and deletes the lying ack — the two severed links, nothing else. Server write/read/freeze path untouched (correctly — it was already right). 16 files, all in scope; the gate correctly excluded the `sas-tail` parity drift.

### Vision / roadmap fit
Goal alignment is dual, and worth naming precisely:
- **`integrity` (#5474)** — "never … contradictory." An ack that unconditionally says "accepted for recording" while nothing persists is a presentation-integrity defect. Deleting it (returning to `goal`/topic parity, which make no recording claim) restores wording-to-truth alignment. This is the honesty half.
- **`self-learning`** — the *actual* value of run-identity tags (`arm:A`, `workflow:v1.3`) is A/B-arm / workflow-cohort analysis of retrieval quality. vnc-047 *claimed* that capability but it was **dead in production** (JS port missing). This fix is what actually delivers it. The capability vnc-047 asserted was never real until this PR; treat vnc-047's tag capability as moving from asserted → deliverable **only after** the post-merge verification below, not on merge alone.

The systemic bare-client gap (on a non-hook MCP client, topic/goal/tags all persist nothing — col-022 session-unaware handler) is unchanged and correctly out of scope. The v1/uni-zero "harmonize the disclosure across topic/goal/tags" push is now **moot**: v3 makes tags actually persist on the hook path and deletes the claim entirely, so there is no divergent disclaimer to harmonize. The right pivot happened.

No standalone `uni-security-reviewer` comment exists on this issue/PR — the security surface was assessed inside both design reviews (§5) and gate check 6, not as a dedicated pass. For a change that is (JS additive payload key + Rust fn deletion + test deletions) with no server runtime path touched and no new untrusted sink, an embedded security-surface treatment is proportionate; I would not require a fresh standalone review to merge. Findings I concur with:
- **SQLi — safe.** Tags bind parameterized (`db.rs .bind(tag)`); since value-opacity forbids validation, parameterization is the only defense and it is present. Unchanged by this PR.
- **Unbounded tag count/length — known, accepted gap; do NOT cap here.** In the single-client trust model these are the operator's own labels in the operator's own DB, and whole-set-once bounds the write. Capping now would break value-opacity/oracle parity (parity gate red). Flag for the human: if `personal-cloud`/direct-HTTPS MCP integrations ever make tags externally supplied, a symmetric *count* cap (Rust oracle + JS + defensive listener) belongs in a deliberate ADR-002 amendment — not this bugfix.

### Approach commentary
- **JS port correctness — verified against ground truth.** I read `hook.rs:860-917` and diffed it clause-by-clause against the new JS block: Start-only ✓, strings-only after-trim ✓ (`typeof t === "string" && t.trim() !== ""` ↔ `filter_map(as_str).filter(!trim().is_empty())`), no byte cap ✓, omit-when-empty ✓, infallible via `try/catch` ✓, placed **after** `put("goal")` ✓ so a throw can never drop the already-built cycle frame. Surviving members are kept verbatim (not trimmed) on both sides. The `cycle-start-tags-filtered` golden `["keep","also-keep"]` is exactly what the oracle produces.
- **The real prevention is the parity-corpus case, and it's here.** The bug shipped because vnc-047 changed the Rust oracle with no corpus case, so the JS port silently drifted. This PR adds `cycle::tags` / `_filtered` / `_omitted_when_empty` / `_ignored_non_start` arms + goldens — the exact drift gate whose absence let this ship. Wrong-key-name regressions are now caught deterministically.
- **Whole-set-once vs accumulate** — design-reviewer-v2 ruled "keep whole-set-once, don't fold a semantic flip into a bugfix." I concur. One thing for the human to register: whole-set-once has **never actually executed in production** (tags never persisted), so this PR is the first time the freeze semantic becomes observable. Watch it once tags go live; if the freeze proves wrong for the workflow, that is a deliberate ADR-002 supersession filed separately — not a reason to hold this merge.

### Right-sizing
Outcome / DoD altitude: the OUTCOME is "a run-identity-tagged `context_cycle(start)` in a hook-instrumented session lands in `cycle_tags` and surfaces via `context_cycle_review`, and no surface claims persistence that didn't happen." The PR delivers both halves via the hook path and proves the seam with the layer-2 UDS test — this is the outcome, not just the mechanism.
Size verdict: right-sized. No gold-plating (no cap added; server untouched); not too small (the layer-2 JS-frame→listener→`cycle_tags` test crosses the seam the vnc-047 tests never did).
Follow-up smell: #945 (retire the now-oracle-only Rust hook client) is genuinely separate, not the point of this work — legitimate defer. The post-merge dogfood-client refresh is operational, not a deferred feature.

- **ADR-007 correction scope** — must withdraw only part (a) the ack echo and retain part (b) the listener frozen-skip C13 trace. Corpus already reflects this: **#5675 = "PART (a) WITHDRAWN by bugfix #944; frozen-skip listener tracing retained."** Recommend the human confirm #5675 reads as intended; it appears already done.
- **Is the fix truly proven end-to-end?** In test, yes (parity goldens + layer-2 UDS). In the **live dogfood path**, not until the deployed `~/.unimatrix/dogfood-client` is refreshed post-merge — the same stale-client class as the original defect. This is the one residual risk and it is inherent, not a code gap.

### Recommended actions
1. **Merge.** No blocking findings; root cause covered, fix minimal, drift gate restored, JS↔oracle parity verified.
2. **Post-merge, load-bearing: refresh the deployed `~/.unimatrix/dogfood-client` (and rebuild the server binary).** Until then the JS fix does not reach the live hook and the capability remains dead exactly as before merge. Then run one real `context_cycle(start, tags=[...])` → `context_cycle_review` to confirm tags surface — that is the only proof that closes #944 in production.
3. **Confirm ADR-007 #5675** is scoped to part (a) only (appears complete).
4. **Verification caveat to accept knowingly:** the gate validated from artifacts (cargo/clippy/pytest not re-run per lesson #5207) and the full tools+lifecycle pytest exceeded the 10-min ceiling (covered by `--collect-only` + targeted `-k cycle or tag` runs). Reasonable for a JS + ack-delete + test-delete change with untouched server runtime; budget a post-merge CI round-trip for the cross-platform JS matrix as usual.
5. **Do not auto-file** the accumulate-semantics question or an externally-supplied-tags cap — both are deliberate future ADR-002 decisions, the human's call, only if a real need appears once tags are live.

Issue #973, opened 2026-08-02
[clippy] transcript_hold.rs:268 match-can-be-? blocks `clippy --workspace -- -D warnings`


dug-21:
## Summary

`cargo clippy --workspace --all-targets -- -D warnings` fails on a pre-existing lint in `unimatrix-server`:

```
error: this `match` expression can be replaced with `?`
  --> crates/unimatrix-server/src/infra/transcript_hold.rs:268:33
```

## Context

Discovered during verification of the #972 bugfix (branch `bugfix/972-hnsw-idmap-membership-filter`). The #972 branch touches only `unimatrix-vector` and does not modify `unimatrix-server`, so this lint is pre-existing and unrelated to that fix. It reproduces on the branch point `651c608d`.

## Reproduction

```bash
cargo clippy -p unimatrix-server --lib -- -D warnings
```

## Fix

Replace the `match` at `transcript_hold.rs:268` with the `?` operator (clippy's suggested rewrite using `guard.get(session_id)?`).

## Notes

Not fixed in the #972 PR to keep that bugfix scoped. Filing to preserve the signal so the workspace clippy gate can be made green in a targeted change.

Issue #828, opened 2026-06-22
harden: make atomic vector dump flip graph+data as one versioned set (F1 from #826 security review)


dug-21:
Advisory follow-up from the #826 (#824) security review. **Non-blocking** — the #824 atomic-dump is already strictly safer than the pre-PR window and never loses store data.

## Observation (F1, low)
`persistence.rs` publishes `hnsw.graph` and `hnsw.data` as two separate renames over **stable** filenames, then the `.meta` last. A crash between the graph and data renames can pair a new graph with an old data file. Load's existence checks pass, so this surfaces as either an `Err` (→ safe Item-1 boot fallback) or a subtly inconsistent index that self-heals from the store. No store data loss.

## Suggested hardening
Point the meta `basename=` at a **versioned** graph/data pair (e.g. `unimatrix.v{N}.hnsw.graph/.data`) so the single `.meta` rename flips the whole set atomically — one publish point instead of three. Clean up the prior version's files after the meta flip.

## Context
- Source of truth is the store; the maintenance heal pass already repairs a degraded index.
- Keep filename/`basename` conventions consumable by snapshot/eval (#2676, #2661).

Labels: goal:personal-cloud

Issue #827, opened 2026-06-22
feat: eager boot-time heal-from-store for per-slug vector index (Item 2, deferred from #824)


dug-21:
Deferred Item 2 from #824 (which landed Items 1+3: load-fallback + atomic dump). Investigator, architect, and product reviewer all concurred this is a strictly larger, separate capability.

## Scope
Eagerly rebuild a per-slug vector index from the store **at boot** when the vector dir is missing/torn, rather than relying on incremental recovery.

## Why deferred (not a bugfix)
A true boot-time rebuild means re-embedding every active entry: embed budget, model-load dependency, async batching, and boot latency proportional to corpus size. That is feature/design work, not a minimal fix.

## Why #824 is sufficient for resilience
#824 already restores graceful degradation: on a torn/missing index the slug boots **empty** and self-heals incrementally via the existing capped maintenance heal pass (`services/status.rs` Sub-case B, already capped per #3762). Item 2 changes *when* recovery happens (eager at boot), not *whether*.

## Considerations for design
- Re-embed budget / rate limiting; reuse the existing capped heal pass machinery vs. a dedicated boot rebuilder.
- Boot-latency bound (don't block slug serving on a full corpus re-embed).
- Model availability at boot (graceful skip if embed model unavailable → fall back to incremental heal).

Labels: goal:personal-cloud

PR #974, opened 2026-08-02
fix(vector): membership-filter IdMap at load to detect graph-under-counts-DB after DB copy (#972)

### LINKED ISSUE #972: fix(vector): startup count-consistency check — detect populated vectors-table vs empty hnsw index after DB copy, trigger heal (portability)
## Summary
Copying a Unimatrix deployment by moving only the DB (and config) leaves the persisted HNSW vector index out of sync with the store, and **the system does not detect it** — retrieval silently degrades and the existing heal never fires.

## Reproduction (observed)
1. Copied `unimatrix.db` and `config.toml` from machine A to machine B (the HNSW index files were **not** copied).
2. Launched Unimatrix at the new location, expecting the HNSW vectors to be rebuilt there.
3. They were **not** rebuilt. The `vectors` table in the DB was populated (copied with the DB), but the on-disk HNSW index held **zero** vectors. Every existence/error check passed, so nothing triggered a rebuild.
4. Only after **manually deleting rows from the `vectors` table** did the system self-heal and re-embed.

## Root cause
The heal pass keys off the DB `vectors` table as the source of truth for "what should be indexed." When the DB (including a fully-populated `vectors` table) is copied but the HNSW index files are not, the table says "N vectors exist" while the physical index has 0. The heal compares expected-vs-recorded (both from the DB) and sees no gap — so it does nothing. Deleting the table rows is what forced the mismatch it could see.

The healing machinery already exists (capped maintenance heal pass, `services/status.rs` Sub-case B). **The detection is what's missing** — nothing compares the *actual loaded HNSW index count* against the DB's recorded count.

## Scope
- On per-slug boot, compare the **actual persisted/loaded HNSW vector count** against the DB `vectors`-table count (the recorded truth).
- On mismatch (index under-count — including the empty-index-but-populated-table case above), mark the index degraded and enqueue the **existing** capped heal pass to repopulate. No new healing logic.
- Cheap check: two counts, no re-embed at check time. Reuse the existing heal for actual repopulation.

## Out of scope
- Eager full boot-time rebuild (#827 — the missing/torn-dir case).
- Atomic dump-flip hardening (#828).
- Any new re-embedding path — reuse the capped heal.

## Definition of done
After copying **only** `unimatrix.db` + `config.toml` to a fresh location (no HNSW index files) and starting the server, the slug detects that the recorded `vectors` count exceeds the actual index count on startup and retrieval returns correct results (index repopulated via the existing heal) — **without** manually deleting `vectors`-table rows, and not merely a logged warning.

## Vision alignment
`goal:personal-cloud` — "backup = volume snapshot; recovery = restore + restart" and moving a deployment must not silently break retrieval. Advances the reliability nfr (N7 — a degraded index never silently serves wrong) and hardens BACKUP-RESTORE portability. Relates to #827, #828.

dug-21:
## Summary
Fixes #972. After copying only `unimatrix.db` + `config.toml` (no HNSW index files) to a new location, the DB `vector_map` table recorded N vectors but the loaded HNSW graph held fewer (0 in the empty case). Because `VectorIndex::load` rebuilt the in-memory IdMap from `vector_map` with **no cross-check against the loaded graph**, `contains()` returned true for absent entries — so the existing capped heal (Sub-case B, keyed off `!contains`) never fired and retrieval silently degraded. This is the #5718 host-migration deadlock, whose only prior cure was a manual `DELETE FROM vector_map`.

## Fix
Inside `VectorIndex::load`, on the load-**success** branch only: enumerate the loaded HNSW graph's origin_ids across **all layers** (via `IterPoint` — the layer-0-only trap would drop the ~6% of points at level ≥1, GH#286/#1712), and filter the `vector_map` mappings to only graph-present data_ids **before** `from_parts`. This makes `contains()` truthful → the **existing** capped maintenance heal repopulates only the genuinely-missing entries. Retained mappings keep valid points searchable. A `warn!` reports recorded/actual/dropped counts (previously-invisible divergence).

- **No** changes to `status.rs`, **no** new re-embed path, **no** eager rebuild, **no** DELETE cure.
- Hard-`Err` degradation path (#5272) untouched.
- Chosen over the issue's literal "two-count compare": a count compare can't make the heal fire (Sub-case B keys off the still-lying `contains()`) and counts can coincidentally match via stale points. Endorsed by architect + product review.

## Changed files
- `crates/unimatrix-vector/src/persistence.rs` — membership filter in `VectorIndex::load` + 2 new tests
- `crates/unimatrix-vector/Cargo.toml` — `tracing` dep (for the `warn!`)
- `Cargo.lock`
- `product/features/bugfix-972/agents/972-investigator-report.md`

## Tests
- `test_load_graph_undercounts_vector_map_filters_absent_entries` — loads a dump whose graph (200 points) under-counts `vector_map` (5 injected graph-absent rows); asserts `contains()` false for absent, true for all 200 present (level ≥1 retention = the N1 guard).
- `test_load_graph_undercounts_retained_points_searchable` — retained points still self-search back; the absent entry never surfaces.
- 657 unit tests passing, 0 failures. `unimatrix-vector` clippy clean under `-D warnings`.

## Gate
Bug Fix Validation — **PASS** (6/7, 1 N/A). Minimal diff, root cause addressed, stewardship present for all phases.

## ⚠️ Pre-merge precondition
The mandatory **infra-001 integration smoke gate could not run in this container** (`libonnxruntime` absent system-wide + pytest uninstallable under PEP-668 — both pre-existing/environmental, not introduced here). **Run the infra-001 smoke suite via the Docker harness before merging.**

Pre-existing, unrelated (NOT in this diff): workspace clippy error at `transcript_hold.rs:268` → filed as #973.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,6 +8,9 @@ patches/anndists/target/
 
 # Documentation and product files
 product/
+# ...but the infra-001 integration test harness is COPYd by its Dockerfile
+# (test-runtime stage) and must remain in the build context.
+!product/test/infra-001/
 *.md
 !README.md
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3929,6 +3929,7 @@ dependencies = [
  "rand 0.9.2",
  "tempfile",
  "tokio",
+ "tracing",
  "unimatrix-store",
 ]
 
--- a/crates/unimatrix-server/src/test_support.rs
+++ b/crates/unimatrix-server/src/test_support.rs
@@ -111,32 +111,34 @@ impl TestHarness {
     ///
     /// Returns `None` if the ONNX model is not available.
     pub async fn new_with_expander(store_path: &Path, ppr_expander_enabled: bool) -> Option<Self> {
-        if skip_if_no_model() {
-            return None;
-        }
-
+        let embed_handle = Self::load_embed_handle().await?;
         let store =
             unimatrix_store::SqlxStore::open(store_path, unimatrix_store::PoolConfig::default())
                 .await
                 .expect("failed to open test store");
         let store = Arc::new(store);
-
-        let vector_config = VectorConfig::default();
         let vector_index = Arc::new(
-            VectorIndex::new(Arc::clone(&store), vector_config)
+            VectorIndex::new(Arc::clone(&store), VectorConfig::default())
                 .expect("failed to create vector index"),
         );
+        Some(Self::wire(
+            store,
+            vector_index,
+            embed_handle,
+            ppr_expander_enabled,
+        ))
+    }
 
-        let vector_adapter = VectorAdapter::new(Arc::clone(&vector_index));
-
-        let entry_store = Arc::clone(&store);
-        let vector_store = Arc::new(AsyncVectorStore::new(Arc::new(vector_adapter)));
-
+    /// Build (and await readiness of) the shared ONNX embed handle. `None` when the
+    /// model is unavailable (same `skip_if_no_model()` semantics as `new`). Exposed so a
+    /// round-trip test can share ONE handle across the dump → `VectorIndex::load` → heal
+    /// phases (bugfix-972 Leg 2), authoring the graph before the harness exists.
+    pub async fn load_embed_handle() -> Option<Arc<EmbedServiceHandle>> {
+        if skip_if_no_model() {
+            return None;
+        }
         let embed_handle = EmbedServiceHandle::new();
-        let config = EmbedConfig::default();
-        embed_handle.start_loading(config, None);
-
-        // Wait for model to load
+        embed_handle.start_loading(EmbedConfig::default(), None);
         let mut attempts = 0;
         loop {
             match embed_handle.get_adapter().await {
@@ -151,6 +153,30 @@ impl TestHarness {
                 }
             }
         }
+        Some(embed_handle)
+    }
+
+    /// Construct a harness around a pre-built store, a (possibly pre-loaded) vector index,
+    /// and an embed handle (PPR expander OFF). Enables the bugfix-972 Leg 2 round-trip:
+    /// author+dump a graph, inject a DB-only-copy under-count, `VectorIndex::load`, heal.
+    pub async fn from_parts(
+        store: Arc<Store>,
+        vector_index: Arc<VectorIndex>,
+        embed_handle: Arc<EmbedServiceHandle>,
+    ) -> Self {
+        Self::wire(store, vector_index, embed_handle, false)
+    }
+
+    /// Wire a full `ServiceLayer`; body is the pre-refactor `new_with_expander` verbatim.
+    fn wire(
+        store: Arc<Store>,
+        vector_index: Arc<VectorIndex>,
+        embed_handle: Arc<EmbedServiceHandle>,
+        ppr_expander_enabled: bool,
+    ) -> Self {
+        let vector_adapter = VectorAdapter::new(Arc::clone(&vector_index));
+        let entry_store = Arc::clone(&store);
+        let vector_store = Arc::new(AsyncVectorStore::new(Arc::new(vector_adapter)));
 
         let adapt_service = Arc::new(AdaptationService::new(
             unimatrix_adapt::AdaptConfig::default(),
@@ -169,12 +195,7 @@ impl TestHarness {
                 .expect("test RayonPool construction must succeed"),
         );
 
-        // crt-053: start from the production default config and toggle only the expander flag,
-        // so OFF (default) construction stays bit-identical to pre-crt-053 behavior. All PPR knobs
-        // (alpha, blend weight, inclusion threshold, depth, ceilings) remain at production defaults
-        // — the seed-filter tests observe the Phase 0 `graph_expand` injection at production parity,
-        // isolating it via a topic filter (which excludes graph-injectable neighbors from the HNSW
-        // pool) rather than by altering scoring config.
+        // crt-053: toggle only the expander flag; OFF stays bit-identical to pre-crt-053.
         let inference_config = crate::infra::config::InferenceConfig {
             ppr_expander_enabled,
             ..crate::infra::config::InferenceConfig::default()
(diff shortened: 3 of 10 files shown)


dug-21 (commented):
## Security Review — bugfix-972 (fresh context)

**Risk level: LOW. No blocking findings.** Recommend merge.

Reviewed the full `main...HEAD` diff, the changed `persistence.rs` in full, the boot-fallback caller (`main.rs:931`), and `index.rs::get_embedding`.

### OWASP
- **Injection:** none. No shell/SQL/format-string with untrusted input. Mappings come from the internal store (`iter_vector_mappings`), a trusted DB source. No new external input surface.
- **Deserialization:** unchanged. `load_hnsw` (pre-existing) is untouched; the new code only reads already-deserialized graph points and DB rows.
- **Input validation:** the fix *adds* validation — it reconciles `vector_map` against the actually-loaded graph. Net strengthening of a trust boundary that previously trusted the DB blindly.
- **Access control / secrets:** no trust-boundary change; boot-time internal op. `warn!` logs only integer counts (recorded/actual/dropped) — no IDs, content, or PII.

### Blast radius (worst case)
- **Non-destructive by design:** the filter mutates only the in-memory `IdMap`. `vector_map` is **never** DELETEd, so a copy loaded later *with* the HNSW dir rebuilds fully. This is the safest possible failure mode.
- Wrongly-drop bug → over-heal / re-embed, but heal is **capped + idempotent** → self-correcting, bounded. Wrongly-retain bug → status quo (no regression). No data-corruption, DoS, info-disclosure, or privilege path.

### Regression risk
- Filter sits strictly on the **load-SUCCESS** branch (after `load_hnsw` `?`, before `from_parts`). The hard-Err degradation path (boot → empty index, `main.rs:931`) is **untouched**. Confirmed.
- Agreement case (graph == vector_map): `dropped == 0`, `filtered == mappings` → behavior identical, plus one O(N) boot-time HashSet pass.
- `dropped = recorded - filtered.len()`: `filtered` is a strict subset → no usize underflow.

### All-layers correctness (the load-drop trap)
- Enumeration via `hnsw.get_point_indexation()` IntoIterator matches `index.rs::get_embedding` byte-for-byte (both walk all layers per GH#286 / lesson #1712). A `get_layer_iterator(0)` scan would have dropped level>=1 points; this does not. The N1 test (200 points, all retained) guards it. Correct.

### Dependency safety
- `tracing = "0.1"` added to `unimatrix-vector` — already a direct dep of 5 other workspace crates at the identical version, single locked `0.1.44`. No new transitive tree, no CVE surface, aligns with the workspace logging convention.

### Minimal-change
- Diff is tightly scoped: `load` filter + 2 tests + Cargo.toml/lock dep + 2 doc reports. No unrelated changes.

### Informational (non-blocking)
- Only the under-count direction is surfaced (`dropped > 0`). The reverse — graph holds origin_ids absent from `vector_map` (orphan graph points) — is out of scope for #972 and pre-existing; not a regression. Fine to leave.

dug-21:

**Gate**: pr-review | **Stance**: Merge-defensible. Two things gate the human's confidence, neither a code defect: (1) the mandatory infra-001 smoke is env-blocked and unrun, and (2) the DoD's second leg — retrieval actually self-repopulates via the heal — is proven by construction, not by any executed test. Run the smoke via Docker; decide add-round-trip-now vs track-as-follow-up.

### Vision / roadmap fit
Sits squarely on `goal:personal-cloud` (#5713) and its "backup = volume snapshot; recovery = restore + restart" portability promise, which is false the moment a `unimatrix.db` moves without its `vector/` dir. Extends capability **N7 (#5592)** into the exact hole its current proof does not cover: N7's `done_when` and `proven_by` (#826) exercise the *torn-or-deleted dir → empty index → heal* case; this fix closes the adjacent *loads-successfully-but-under-counts* case, where the IdMap is rebuilt from `vector_map` (recorded truth) instead of the loaded graph (actual truth), so `contains()` lies and both repair legs stay dead. N7's `residual` field names #827/#828 but not this DB-copy-undercount case — so this genuinely widens N7's coverage. Correctly sequenced: #827 (eager boot rebuild — *when* recovery happens) and #828 (atomic dump hardening) remain legitimately separate, larger work, not preconditions here.

### Approach commentary
The membership-filter is the right mechanism and I have no better alternative. It attacks the diagnosed root cause (IdMap poisoned at load) rather than the DoD's literal "two-count compare" symptom — a bare count compare can't make the existing heal fire (Sub-case B keys off `contains()`, which would still lie) and can coincidentally match under staleness. Filtering the IdMap to graph-present `data_id`s makes `contains()` truthful, so the *existing* capped heal repopulates only the genuinely-absent entries with zero `status.rs` change, and keeps valid points searchable during repopulation (strictly better than the old `DELETE FROM vector_map` cure). Fails safe in both directions. The load-bearing N1 risk (all-layers `IterPoint` vs `get_layer_iterator(0)`) is correctly handled and guarded by a 200-point test asserting level-≥1 retention. Diff is minimal and disciplined: `status.rs` untouched, eval load site untouched, hard-`Err` degradation path untouched (confirmed by the changed-files list — only `persistence.rs`, `Cargo.toml`, `Cargo.lock`, 2 docs).

No standalone security review is posted to the PR or issue thread — the only security assessment on record is the design-review's inline "Security — none." I concur with that read from the diff: the change is a read-only enumeration of an already-loaded local graph plus an in-memory filter, before the index is shared behind its `RwLock`. No new external input, no new surface, no secrets, no unsafe. Nothing here for a security review to overturn. **Recommend the leader confirm the pr-review's own fresh security pass actually ran** rather than inheriting the design-time assessment — cheap to close, and my spawn inputs named a security review I did not find.

### Capability coverage (advisory)
Target: **N7 #5592** — archetype: threshold (nfr floor: "corrupt/missing per-slug index degrades-and-heals, never silently serves wrong"). Coverage: this PR does not re-claim N7; it **extends** N7's proof to the loaded-but-undercounts case. `done_when` clause map: "detects recorded > actual index membership at load" → covered (membership filter); "self-repopulates via existing per-slug heal, not a boot abort / not a logged warning" → covered *by construction* (contains() truthful → Sub-case B fires unchanged), ⚠️ **not covered by any executed test** (see Right-sizing). Capability-status implication (recommendation only): N7 stays `delivery:proven`; proof-set extended once an end-to-end undercount→heal→search test lands. Status mutation is uni-zero (vision session)'s call, not mine.

### Right-sizing
Outcome / DoD altitude: *after a DB-only copy, the slug detects recorded > actual on startup and **retrieval returns correct results** without manual DELETE — not merely a logged warning.* The fix targets that outcome, not a warning. The honest in-time partial (only retained points searchable at boot; full correctness after the capped heal drains — ~115 ticks for a 2,300-entry corpus) is a **legitimate defer** — eager rebuild is a different outcome (#827), not a dropped last mile. Right-sized; no gold-plating, no mechanism-without-outcome at the design level.
- **The one verification gap**: the outcome has two legs. Leg 1 — `contains()` becomes truthful — is unit-proven (both new tests). Leg 2 — the existing heal then *fires and repopulates* so "retrieval returns correct results" — is asserted by code-path reasoning (Sub-case B keys off `!contains`, unchanged), **not exercised by any test at any level**. The investigator, the design reviewer (N-recommended coverage), and the fix-approach uni-zero all explicitly called for a **server-level tick-heals-then-search round-trip** (status.rs suite, #2676 four-step pattern) starting from a `meta point_count>=1` / under-populated-graph fixture, as a *delivery-acceptance* item. That test was not delivered, and the verify agent confirmed the bug "is not exercised by any existing MCP-level suite," so infra-001 smoke does not cover leg 2 either. Low risk given the reasoning is airtight and `status.rs` is provably unchanged — but the DoD's headline verb ("retrieval returns correct results") currently rests on argument, not a green test. Follow-up smell: this round-trip is arguably the point of the fix, and it fits within reasonable incremental effort (the fixtures exist). Recommend pulling it in now unless the human accepts leg-2-by-construction + a tracked follow-up.

1. **Merge before infra-001 smoke runs?** — No. The smoke gate is mandatory and was env-blocked here (`libonnxruntime` absent + PEP-668 pytest), not passed. Run `docker compose -f product/test/infra-001/docker-compose.yml up --build --abort-on-container-exit` with `PYTEST_ARGS="-m smoke"` (Docker bundles onnxruntime) before merge. This is a documented gate, not code rework.
2. **Is unit coverage authoritative for this fix?** — For leg 1 (the load filter), yes — the bug is internal to `VectorIndex::load`, not MCP-visible. For leg 2 (heal actually repopulates end-to-end), no — that is the gap above.
3. **Does #973 (workspace clippy at transcript_hold.rs:268) block?** — No. Pre-existing, reproduces on base `651c608d`, outside this diff, filed. The changed crate is clippy-clean under `-D warnings`.

### Recommended actions
- **Before merge (mandatory):** run infra-001 smoke via the Docker harness.
- **Confirm** the pr-review's own security pass ran (or accept the design-time "none" on record — I concur with it from the diff).
- **Decide on leg-2 verification:** either add the server-level undercount→tick-heals→search round-trip now (recommended; fixtures exist, closes the DoD's headline verb with a green test and extends N7's proof-set), or merge on leg-2-by-construction and file a tracked follow-up. Do not let it silently ride the env-blocked smoke, which does not cover it.
- After merge, N7's proof-set extension and lesson #5719 are already staged — hand to uni-zero (vision) for any capability-map touch.

dug-21:
## Update — round-trip test added, harness partially repaired, smoke gate deferred

**Added since gate PASS:**
- `5624a1b6` — server-level round-trip test `test_maintenance_heal_repopulates_graph_undercount_after_db_only_copy_load` (`crates/unimatrix-server/tests/heal_roundtrip_972.rs`), proving **Leg 2** of the DoD (heal fires → retrieval self-repopulates, no manual `DELETE`). Type-checks/clippy/fmt clean; gated by the crate's existing `skip_if_no_model()`, so it executes under the Docker harness (needs libonnxruntime + model).
- `957467fc` — three infra-001 smoke-harness build defects fixed (`.dockerignore` `product/` re-include, compose context `../../..`, onnx `cp -a …*.so*`).

**Smoke gate deferred (maintainer decision):** the from-source smoke build has a deeper remaining blocker — the builder stage can't link `ort`/onnxruntime (no local ORT + `ORT_LIB_LOCATION` at compile time). Full diagnosis and fix plan captured in **#975** for a focused session. This PR merges **without** the integration smoke gate; #972's code fix is unit-verified (657 tests, both new persistence tests + the round-trip test) and security-reviewed LOW.

**Merge readiness:** READY (smoke gate tracked separately in #975).

